PR #3, opened 2026-03-05
AI-365: add deterministic TOON write pipeline

shafty023:
- Add write_merged_patterns.py for deterministic JSON-to-TOON conversion.
- Refactor process-learnings to output merge-result.json instead of writing TOON directly
- Integrate as Step 8.5 in run-loop.sh.

### DIFF
--- a/plugins/code/scripts/run-loop.sh
+++ b/plugins/code/scripts/run-loop.sh
@@ -498,6 +498,25 @@ post_iteration_processing() {
     emit_skipped_step 8 "process_learnings"
   fi
 
+  # Step 8.5: Write merge-result.json → org-patterns.toon (deterministic)
+  local merge_script="$sl_tools_dir/write_merged_patterns.py"
+  local merge_result="$workdir/.learnings/merge-result.json"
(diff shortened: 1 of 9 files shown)


plugins/code/scripts/run-loop.sh
```diff
@@ -498,6 +498,25 @@ post_iteration_processing() {
     emit_skipped_step 8 "process_learnings"
   fi
 
+  # Step 8.5: Write merge-result.json → org-patterns.toon (deterministic)
+  local merge_script="$sl_tools_dir/write_merged_patterns.py"
+  local merge_result="$workdir/.learnings/merge-result.json"
+  if [[ -f "$merge_result" ]] && [[ -f "$merge_script" ]]; then
```
wongk, line 504: so this will only execute if they also have the self-learning plugin installed?

wongk (approved):

plugins/code/scripts/run-loop.sh
shafty023: That is correct. That file can't exist if the plugin doesn't exist to generate it.

PR #13, opened 2026-03-07
fix(code): resolve judges path for agent snapshot

aponamarev:
# FIX - resolve judges path for agent snapshot

This PR fixes agent snapshot path resolution so the closed loop run-loop correctly finds the judges plugin agents directory across monorepo, cache, and marketplace installation layouts. Previously, the snapshot read from the code plugin's agents path; it now resolves the judges plugin agents via a four-level fallback strategy and records the correct `plugin` field in the manifest.

## Key Changes

**Core Feature:**
- Implemented `resolve_judges_agents_dir()` in `run-loop.sh` with four fallback levels: (1) `CLOSEDLOOP_JUDGES_AGENTS_DIR` env override, (2) repo-relative path for monorepo, (3) non-versioned sibling `.../judges/agents`, (4) versioned sibling `.../judges/<version>/agents` (latest semver)
- Updated `store_agents_snapshot()` to use the resolver instead of hardcoded `$SCRIPTS_DIR/../agents`, and to read plugin version from the judges plugin manifest
- Corrected manifest `plugin` field from `"code"` to `"judges"` so the snapshot correctly identifies its source

**Supporting Changes:**
- Added `is_valid_judges_agents_dir()` and `resolve_versioned_judges_agents_dir()` helpers; updated CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,16 +32,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 #### Added
 - Compound Bash command prohibition in GitHub mode — no `&&`, `||`, `;`, or `|` pipes allowed
 
+### code v1.0.6
+
+#### Fixed
+- Fixed judges agents path resolution in `run-loop.sh` to support monorepo, cache, and marketplace installation layouts via a four-level fallback strategy (`CLOSEDLOOP_JUDGES_AGENTS_DIR` env override → repo-relative path → non-versioned sibling → latest semver-versioned sibling)
+- Fixed agent snapshot to read judge agents from the judges plugin rather than the code plugin, and corrected `plugin` field in manifest to `"judges"`
+
 ### code v1.0.5
 
 #### Changed
 - Updated `review-delta.schema.json` description to reference "code hybrid workflow" instead of "impl-plan hybrid workflow"
 - Updated `compliance-checkpoint.md` to reference `/code` instead of `/impl-plan`
 - Removed `Bash` from `visual-qa-subagent` tool list to prevent shell access during visual QA
-
-#### Fixed
-- Fixed judges agents path resolution in `run-loop.sh` to support monorepo, cache, and marketplace installation layouts via a four-level fallback strategy (`CLOSEDLOOP_JUDGES_AGENTS_DIR` env override → repo-relative path → non-versioned sibling → latest semver-versioned sibling)
-- Fixed agent snapshot to read judge agents from the judges plugin rather than the code plugin, and corrected `plugin` field in manifest to `"judges"`
 
 #### Security
 - Added credential theft blocklist to `pretooluse-hook.sh`: denies Bash commands and file access targeting macOS Keychain, browser cookie databases, SSH private keys, and cloud credentials
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Updated `compliance-checkpoint.md` to reference `/code` instead of `/impl-plan`
 - Removed `Bash` from `visual-qa-subagent` tool list to prevent shell access during visual QA
 
+#### Fixed
+- Fixed judges agents path resolution in `run-loop.sh` to support monorepo, cache, and marketplace installation layouts via a four-level fallback strategy (`CLOSEDLOOP_JUDGES_AGENTS_DIR` env override → repo-relative path → non-versioned sibling → latest semver-versioned sibling)
+- Fixed agent snapshot to read judge agents from the judges plugin rather than the code plugin, and corrected `plugin` field in manifest to `"judges"`
+
 #### Security
 - Added credential theft blocklist to `pretooluse-hook.sh`: denies Bash commands and file access targeting macOS Keychain, browser cookie databases, SSH private keys, and cloud credentials
 - Blocklist applies to all Claude sessions, not just ClosedLoop-managed sessions
--- a/plugins/code/scripts/run-loop.sh
+++ b/plugins/code/scripts/run-loop.sh
@@ -188,14 +188,10 @@ resolve_judges_agents_dir() {
   local plugins_parent
   plugins_parent="$(dirname "$code_root_dir")"
 
-  local repo_judges_agents=""
   local sibling_judges_agents="$plugins_parent/judges/agents"
   local sibling_judges_root="$plugins_parent/judges"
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -39,6 +39,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Updated `compliance-checkpoint.md` to reference `/code` instead of `/impl-plan`
 - Removed `Bash` from `visual-qa-subagent` tool list to prevent shell access during visual QA
 
+#### Fixed
```
shafty023: this is missing a plugin version bump

plugins/code/scripts/run-loop.sh
```diff
@@ -132,6 +132,117 @@ bootstrap_learnings() {
...
+    version_dir="$(basename "$(dirname "$agents_dir")")"
+    if [[ "$version_dir" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][A-Za-z0-9]+)*$ ]]; then
+      versioned_candidates+="${version_dir}|${agents_dir}"$'\n'
+    fi
+  done < <(find "$judges_root" -mindepth 2 -maxdepth 2 -type d -name "agents" 2>/dev/null)
+
+  if [[ -n "$versioned_candidates" ]]; then
+    latest_agents_dir="$(
+      printf "%s" "$versioned_candidates" \
+        | LC_ALL=C sort -t'|' -k1,1V \
```
shafty023: **[P2]** Use portable sort for versioned judges discovery

`resolve_latest_versioned_judges_agents_dir` picks the latest version via `sort -V`, which is a GNU-only extension; on environments using the default BSD sort (e.g., macOS), this branch can fail and return no path, so versioned sibling layouts are skipped entirely even when judges plugins are installed. This breaks the newly added fallback for marketplace/cache installs on those systems.

plugins/code/scripts/run-loop.sh
```diff
@@ -132,6 +132,117 @@ bootstrap_learnings() {
...
+
+  while IFS= read -r agents_dir; do
+    local version_dir
+    version_dir="$(basename "$(dirname "$agents_dir")")"
+    if [[ "$version_dir" =~ ^[0-9]+\.[0-9]+\.[0-9]+([.-][A-Za-z0-9]+)*$ ]]; then
+      versioned_candidates+="${version_dir}|${agents_dir}"$'\n'
+    fi
+  done < <(find "$judges_root" -mindepth 2 -maxdepth 2 -type d -name "agents" 2>/dev/null)
+
+  if [[ -n "$versioned_candidates" ]]; then
```
shafty023: **[P3]** Ignore newest-version corruption and try older valid judges dirs

The resolver selects the newest semver directory first and validates only that one before returning, so if the latest folder exists but contains no `.md` files (partial install/corrupt cache), it returns failure and never falls back to a valid older version. This can result in missing judge snapshots even though a previous judges version is usable.

CHANGELOG.md
aponamarev: Fixed

plugins/code/scripts/run-loop.sh
aponamarev: macOS's BSD sort does support the -V flag for version sorting — it's not a GNU-only extension. I can confirm this directly on the machine

plugins/code/scripts/run-loop.sh
aponamarev: fixed

plugins/code/scripts/run-loop.sh
```diff
@@ -132,6 +132,117 @@ bootstrap_learnings() {
...
+  local plugins_parent
+  plugins_parent="$(dirname "$code_root_dir")"
+
+  local repo_judges_agents=""
+  local sibling_judges_agents="$plugins_parent/judges/agents"
+  local sibling_judges_root="$plugins_parent/judges"
+  local tried=()
+
+  if [[ -n "$code_root_dir" ]]; then
+    repo_judges_agents="$(cd "$code_root_dir/../../judges/agents" 2>/dev/null && pwd || echo "$code_root_dir/../../judges/agents")"
```
shafty023: **[P2]** Level 2 monorepo path traverses one directory too many — effectively dead code

repo_judges_agents uses $code_root_dir/../../judges/agents which goes two levels up from the code plugin dir. In the monorepo layout (plugins/code → ../../ → repo root), this yields <repo>/judges/agents which doesn't exist. The correct monorepo-relative path is $code_root_dir/../judges/agents (one level up to plugins/, then into judges/agents). Level 3's sibling path already handles this correctly, so the function works end-to-end, but level 2 is dead code with a misleading comment. Same issue applies in cache and marketplace layouts.



> **Suggestion:** Change line 197 from `$code_root_dir/../../judges/agents` to `$code_root_dir/../judges/agents`, or remove level 2 entirely since level 3 ($plugins_parent/judges/agents) already covers all non-versioned sibling layouts.

plugins/code/scripts/run-loop.sh
```diff
@@ -132,6 +132,117 @@ bootstrap_learnings() {
   fi
 }
 
+# Returns 0 if the directory exists and includes at least one markdown file.
```
shafty023, line 135: **[P3]** No test coverage for new path resolution functions

Three new shell functions (is_valid_judges_agents_dir, resolve_versioned_judges_agents_dir, resolve_judges_agents_dir) totaling ~117 lines of fallback logic with semver sorting have no corresponding tests. These functions are deterministic and testable with temporary directory fixtures.



> **Suggestion:** Add a bats or bash-based test file that creates temporary directory structures for each layout (monorepo, cache flat, cache versioned) and asserts the correct path is returned.

plugins/code/scripts/run-loop.sh
shafty023, line 1: **[P3]** PR description references non-existent function name

PR body mentions `resolve_latest_versioned_judges_agents_dir()` but the actual function is `resolve_versioned_judges_agents_dir()`. Minor docs/code mismatch.



> **Suggestion:** Update the PR description to use the correct function name.

aponamarev: Fixed

plugins/code/scripts/run-loop.sh
aponamarev: Fixed

plugins/code/scripts/run-loop.sh
aponamarev: We don't have unit tests for bashscript yet

shafty023 (approved):

PR #14, opened 2026-03-07
feat(judges): judge context management via judge-input.json envelope

aponamarev:
# Judge context management via judge-input.json envelope

This PR standardizes judge context management by introducing a canonical `judge-input.json` envelope and moving context orchestration into the judges plugin. Judges no longer assume hardcoded artifact paths; instead, the orchestrator maps artifacts explicitly per run. Investigation-log context is reused for both plan and code evaluations with resilient fallback when missing.

## Key Changes

**Core Feature:**
- Introduced `judge-input.json` as the canonical orchestrator-to-judge contract: `task`, `primary_artifact`, `supporting_artifacts`, `fallback_mode`, and `metadata`. Judges read this envelope first, then resolve mapped artifacts.
- Moved `context-manager-for-judges` agent from `code` plugin to `judges` plugin, reducing circular dependency surface.
- Integrated `investigation-log.md` into judge context: plan mode allocates 15% (4.5K tokens), code mode 10% (3K tokens). When missing, the skill probes `@code:pre-explorer` or runs an internal fallback; judges continue non-blocking if unavailable.
- Centralized input instructions via shared preamble injection: `common_input_preamble.md` (SSOT) plus `plan_preamble.md` / `code_preamble.md`. All 13 plan and 11 code judges now use the same input contract; per-agent boilerplate removed.
- Extracted `has_code_changes` in `run-loop.sh` to return integer count of code files changed; code judges skip when `changed_count -eq 0`.

**Supporting Changes:**
- Added `judge-input.schema.json` and `judge-input-contract.md`; refactored judge agents to read envelope and mapped artifacts; simplified judges README to contract-driven architecture; updated docs, changelogs, and dependency map.

**Impact:**
- API: Judge agents now expect `judge-input.json` at `$CLOSEDLOOP_WORKDIR`; orchestrators must build this envelope before launching judges.
- Dependencies: `context-manager-for-judges` moved from `code` to `judges`; `code` no longer declares `artifact-type-tailored-context` via that agent.


plugins/judges/skills/run-judges/SKILL.md
```diff
@@ -221,19 +240,93 @@ if [ ! -f "$CLOSEDLOOP_WORKDIR/plan.json" ]; then
...
+     - `## Files Discovered`
+     - `## Key Findings`
+     - `## Requirements Mapping`
+     - `## Uncertainties`
+
+5. **Never block plan context preparation on investigation context**
+   - If log generation still fails, emit a warning and continue.
+
+6. **Prepare plan-context.json via context-manager-for-judges**
+   - Launch `@code:context-manager-for-judges` with `artifact_type=plan`.
```
shafty023: **[P1]** Invoke context manager from judges plugin

After migration, `context-manager-for-judges` now lives in the `judges` plugin, but this flow still instructs the run to launch `@code:context-manager-for-judges`. In environments where `code` no longer exports that agent (as in this patch), plan/code judge orchestration will fail during context preparation because the required `plan-context.json`/`code-context.json` cannot be produced.

plugins/judges/schemas/judge-input.schema.json
```diff
@@ -0,0 +1,113 @@
...
+          "description": "Whether artifact is mandatory for this run."
+        },
+        "description": {
+          "type": "string"
+        }
+      },
+      "additionalProperties": false
+    }
+  },
+  "additionalProperties": false
```
shafty023, line 121: **[P1]** `source_of_truth` field missing from schema but referenced in 10+ files

Multiple files (SKILL.md, preambles, judge agents) reference `source_of_truth` as a field in judge-input.json, but the schema defines `additionalProperties: false` at the top level and does not include this property. Any envelope containing this field would fail schema validation. Referenced in SKILL.md lines 296, 329, 454, 462, 818; plan_preamble.md:46; code_preamble.md:58; dry-judge.md:380; goal-alignment-judge.md:363; kiss-judge.md:366.



> **Suggestion:** Either add `source_of_truth` as an array property to the schema, or remove all references from SKILL.md, preambles, and judge agent files. Given the semantic overlap with `primary_artifact` + `supporting_artifacts` ordering, consider whether this field is redundant.

plugins/judges/skills/run-judges/SKILL.md
shafty023: **[P1]** Stale `@code:context-manager-for-judges` reference after agent move

The context-manager-for-judges agent was moved from the `code` plugin to the `judges` plugin in this PR. However, SKILL.md line 280 and line 819 still reference it as `@code:context-manager-for-judges`. Since the agent now lives in the judges plugin, this reference will fail at runtime.



> **Suggestion:** Change `@code:context-manager-for-judges` to `context-manager-for-judges` (no plugin prefix needed since it's within the same plugin) or `@judges:context-manager-for-judges`.

plugins/code/scripts/run-loop.sh
```diff
@@ -388,6 +388,27 @@ write_runs_log_entry() {
...
+  if [[ ! -f "$changed_files" ]]; then
+    echo "0"
+    return 1
+  fi
+  local count
+  count=$(jq '[.[] | select(
+    (endswith("plan.json") | not) and
+    (endswith("plan.md") | not) and
+    (endswith("prd.md") | not) and
+    (endswith("judges.json") | not) and
```
shafty023, line 390: **[P2]** `has_code_changes` filter doesn't exclude `code-judges.json`

The jq filter excludes files ending with `judges.json` but uses `endswith("judges.json")` which would match `code-judges.json` as well since it ends with `judges.json`. Wait — actually `code-judges.json` does end with `judges.json`, so this is fine. Let me re-check... Yes, `endswith("judges.json")` would match both `judges.json` and `code-judges.json`, so this is actually not an issue.

plugins/judges/skills/run-judges/SKILL.md
shafty023, line 1: **[P2]** Backward compatibility section describes breaking changes

The 'Backward Compatibility Preservation' section was updated to say plan mode SHOULD launch context-manager-for-judges, build judge-input.json, and prepend preambles. This contradicts the previous behavior where plan mode explicitly did NOT do these things. The section title implies no breaking change, but the content describes new mandatory steps for plan-mode orchestration.



> **Suggestion:** Rename the section to 'Plan Mode Execution Flow' or add a migration note explaining that plan mode now requires context preparation and envelope construction, which is a behavioral change from previous versions.

plugins/judges/skills/run-judges/SKILL.md
```diff
@@ -702,7 +814,13 @@ Before marking this task complete, verify:
 | "report_id should end with '-code-judges'" | Incorrect ID format for code | Use pattern: `{RUN_ID}-code-judges` for code artifacts |
 | "Judge {name} has no metrics" | Empty metrics array | Each CaseScore must have ≥1 MetricStatistics entry |
 | "Context preparation failed" | context-manager-for-judges failed | Check context-manager agent output; verify artifact files exist |
-| "Preamble file not found" | Missing preamble .md file | Verify skills/artifact-type-tailored-context/preambles/{artifact_type}_preamble.md exists |
+| "judge-input.json missing" | Orchestrator did not generate envelope | Build `$CLOSEDLOOP_WORKDIR/judge-input.json` before launching judges |
+| "judge-input schema invalid" | Missing required envelope fields | Ensure required fields: `evaluation_type`, `task`, `primary_artifact`, `supporting_artifacts`, `source_of_truth`, `fallback_mode`, `metadata` |
```
shafty023, line 820: **[P2]** Troubleshooting table lists `source_of_truth` as required field

The troubleshooting error message for 'judge-input schema invalid' lists `source_of_truth` among required fields, but this field does not exist in the schema. This would confuse users debugging validation failures.



> **Suggestion:** Remove `source_of_truth` from the listed required fields in the troubleshooting table, or add it to the schema if intended.

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.0.5",
+  "version": "1.0.8",
```
shafty023: **[P3]** Version jumps from 1.0.5 to 1.0.8 with no 1.0.6/1.0.7 changelog entries

The code plugin version jumps from 1.0.5 to 1.0.8 but the changelog only documents 1.0.8. Versions 1.0.6 and 1.0.7 are skipped. This could confuse consumers tracking version history.



> **Suggestion:** Either use 1.0.6 as the single patch bump for this PR's changes, or add changelog entries for the intermediate versions if they correspond to separate commits.

plugins/judges/schemas/judge-input.schema.json
```diff
@@ -0,0 +1,113 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://closedloop.ai/schemas/judge-input.schema.json",
+  "title": "Judge Input Envelope",
+  "description": "Canonical orchestrator-to-judge context envelope. Defines evaluation task, artifact mappings, fallback semantics, and run metadata.",
+  "type": "object",
+  "required": [
```
shafty023, line 7: **[P2]** Add `source_of_truth` to the judge-input schema

Any workflow that validates `judge-input.json` against this new canonical schema will reject the `source_of_truth` field that the updated skill, preambles, and judges are now told to produce/read. Because the schema never defines that property and also sets top-level `additionalProperties: false`, the contract is impossible to satisfy consistently: either validation fails, or the ordering data referenced throughout `run-judges` is missing.

plugins/code/scripts/run-loop.sh
aponamarev: lol

plugins/judges/skills/run-judges/SKILL.md
aponamarev: fixed

plugins/judges/schemas/judge-input.schema.json
aponamarev: Fixed

plugins/judges/skills/run-judges/SKILL.md
aponamarev: Fixed

plugins/judges/skills/run-judges/SKILL.md
aponamarev: Fixed

plugins/judges/skills/run-judges/SKILL.md
aponamarev: Fixed

plugins/code/.claude-plugin/plugin.json
aponamarev: Fixed

aponamarev: Fixed

plugins/judges/schemas/judge-input.schema.json
aponamarev: Fixed

plugins/judges/schemas/judge-input.schema.json
```diff
@@ -0,0 +1,121 @@
...
```
shafty023, line 7: **[P2]** `source_of_truth` is optional in schema but treated as required in docs

The `required` array lists 6 fields but omits `source_of_truth`. However, SKILL.md:818 troubleshooting lists it as required, and multiple judge agents/preambles reference `source_of_truth` ordering as mandatory. A schema-valid envelope that omits this field would leave judges without evidence prioritization.



> **Suggestion:** Add `"source_of_truth"` to the `required` array in the schema, or clearly document it as optional with a default behavior (e.g., fall back to primary_artifact > supporting_artifacts order).

plugins/code/scripts/run-loop.sh
```diff
@@ -388,6 +388,27 @@ write_runs_log_entry() {
...
+# changed-files.json exists, 1 otherwise. Excludes plan artifacts and .learnings/ paths.
+has_code_changes() {
+  local workdir="$1"
+  local changed_files="$workdir/.learnings/changed-files.json"
+  if [[ ! -f "$changed_files" ]]; then
+    echo "0"
+    return 1
+  fi
+  local count
+  count=$(jq '[.[] | select(
```
shafty023: **[P3]** `has_code_changes` doesn't filter out `code-judges.json`

The jq filter excludes `judges.json` but not `code-judges.json`. If code judge output appears in `changed-files.json`, it would be counted as a code change, potentially triggering unnecessary code judge reruns.



> **Suggestion:** Add `(endswith("code-judges.json") | not) and` to the jq select filter.

plugins/judges/skills/artifact-type-tailored-context/preambles/common_input_preamble.md
```diff
@@ -0,0 +1,38 @@
+# Shared Judge Input Contract (SSOT)
+
+This preamble defines the canonical input-reading contract for all judges.
+
+## Required Read Order
+
+You MUST follow this sequence before analysis:
+
+1. Read `$CLOSEDLOOP_WORKDIR/judge-input.json` first.
+2. Parse envelope fields: `evaluation_type`, `task`, `primary_artifact`, `supporting_artifacts`, `fallback_mode`, `metadata`.
```
shafty023: **[P3]** Shared preamble omits `source_of_truth` from field enumeration

Line 10 instructs judges to parse `evaluation_type`, `task`, `primary_artifact`, `supporting_artifacts`, `fallback_mode`, `metadata` but does not mention `source_of_truth`. The plan and code preambles both reference this field for evidence prioritization, creating an inconsistency in the SSOT contract.



> **Suggestion:** Add `source_of_truth` to the enumerated fields in step 2 of the Required Read Order section.

plugins/judges/skills/run-judges/SKILL.md
```diff
@@ -734,14 +858,16 @@ If a single judge Task call fails during execution:
 - Continue with remaining judges in batch and subsequent batches
```
shafty023, line 860: **[P3]** Backward compatibility claim contradicted by new mandatory steps

The text says 'This ensures existing workflows and orchestrators continue to function without modification' but plan mode now requires context-manager launch, judge-input.json construction, and preamble injection — all new. The compatibility fallback only activates on context-preparation failure, not as a general backward-compat path.



> **Suggestion:** Reword to clarify this is the *new* standard plan mode flow, and that the compatibility fallback exists only for context-preparation failures, not for orchestrators that haven't been updated.

plugins/judges/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "judges",
   "description": "LLM Judges plugin",
-  "version": "1.0.0",
+  "version": "1.2.0",
```
shafty023: **[P3]** Version jump from 1.0.0 to 1.2.0 skips 1.1.0

The plugin version goes directly from 1.0.0 to 1.2.0 without a 1.1.0 release. While valid semver, this is unconventional and may confuse changelog readers expecting sequential minor versions.



> **Suggestion:** Use 1.1.0 unless there was an unreleased 1.1.0, in which case add a note in the changelog explaining the gap.

plugins/judges/schemas/judge-input.schema.json
aponamarev: Fixed

plugins/code/scripts/run-loop.sh
aponamarev: Rejected - Somehow LLM struggles with this - `(endswith("judges.json") | not)` in line 391 filters out `code-judges.json`.

plugins/judges/skills/artifact-type-tailored-context/preambles/common_input_preamble.md
aponamarev: Fixed

plugins/judges/skills/run-judges/SKILL.md
aponamarev: Fixed

plugins/judges/.claude-plugin/plugin.json
aponamarev: Fixed

plugins/code/scripts/run-loop.sh
```diff
@@ -410,15 +432,7 @@ run_judges_if_needed() {
     emit_skipped_step "$CLOSEDLOOP_JUDGES_STEP" "code_judges"
     return 0
   fi
-  local changed_count
```
shafty023, line 413: @aponamarev  At some point we need to move this judge logic out of the run-loop so that users can choose to spend the tokens on this evaluation. Can you create a ticket so we can follow up on this?

shafty023 (approved):

plugins/code/scripts/run-loop.sh
aponamarev: I will ask @kaiticarp to prioritize

PR #18, opened 2026-03-11
Judges 2.0 - consolidation into rubrics

aponamarev:
# Judge Consolidation into Rubrics

This PR consolidates 24 individual judge agents (13 plan + 11 code) into 4 rubric-based judges and introduces manifest-driven orchestration. Plan and code evaluation now run fewer, broader judges—`code-quality-judge`, `design-principles-judge`, `plan-evaluation-judge`, and `solid-principles-judge`—with judge sets and output routing defined in `judge-manifest.json` instead of hardcoded batches.

## Key Changes

**Core Feature:**
- Replaced 13 single-metric plan judges and 11 code judges with 4 consolidated rubric agents that evaluate multiple criteria in a single pass
- Added `agents/judge-manifest.json` to define per-category judge lists, output files, and report IDs—plan mode runs 2 judges, code mode runs 3
- Centralized judge input construction via `build_judge_input.py` and preflight checks via `preflight_judges.sh`
- Standardized verdict semantics with dataclass-based validation in `judge_report_contract.py` (replacing Pydantic)
- Integrated eval-cache logic into `run-judges`; removed standalone `eval-cache` skill; Phase 1.3 now invokes `check_eval_cache.sh` directly

**Supporting Changes:**
- Moved `validate_plan.py` from `plugins/code/scripts/` to `plugins/code/tools/python/`; added `test_validate_plan.py`
- Fixed `check_eval_cache.sh`: pass eval JSON path as CLI arg (avoids quote interpolation issues), output `selected_critics` as valid JSON
- Added `judge-output-validation` skill and `judges:judge-output-validation` reference for reusable report validation

**Impact:**
- API: Judge agent IDs changed—13 agents removed, 4 new consolidated agents; orchestrators must use manifest-defined judge names
- Dependencies: `judges` v2.0.0, `code` v1.1.3


plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.1.2",
```
shafty023, line 4: This is a huge change. Bump this to 1.2.0

plugins/code/skills/plan-validate/SKILL.md
```diff
@@ -20,10 +20,16 @@ Activate this skill **instead of** launching `@code:plan-validator` at every pla
 
 ## Usage
 
-Run the validation script. The `scripts/` directory is relative to this skill's base directory (shown above as "Base directory for this skill"):
+Run the validation script from the code plugin's tools directory:
```
shafty023: an agent activating this skill will not know what this means. It will have to stop and figure out what the code plugin is because when a skill is activated, the LLM is told what the base directory for that skill is but it doesn't have access to `$CLAUDE_PLUGIN_ROOT` which we can use in non-skills such as commands/agents/hooks. You'll need to resolve this and it can't be by hardcoding a path or instructing the LLM to go exploring in `~/.claude/plugins/cache` and trying to figure out where the code plugin root is. You'll want to thoroughly test in isolation any change you make.

plugins/judges/skills/run-judges/scripts/judge_report_contract.py
```diff
@@ -0,0 +1,686 @@
...
+
+    if not results_path.exists():
+        raw_scores: dict[str, CaseScore] = {}
+    else:
+        try:
+            raw_payload = json.loads(results_path.read_text())
+        except (json.JSONDecodeError, OSError) as err:
+            print(f"Warning: results file unreadable ({err}); all judges will be error CaseScores", file=sys.stderr)
+            raw_payload = {}
+        raw_scores = _extract_case_scores(raw_payload)
```
shafty023: **[P1]** Unhandled ValueError in aggregate_results for non-dict/non-list JSON payloads

aggregate_results catches JSONDecodeError and OSError when reading results files, but _extract_case_scores(raw_payload) is called outside the try-except. If the file contains valid JSON that is not a dict or list (e.g., null, 42, "string"), _extract_case_scores raises ValueError that propagates unhandled — contradicting the graceful-degradation design intent.



> **Suggestion:** Wrap the _extract_case_scores call in its own except block: try: raw_scores = _extract_case_scores(raw_payload) except (ValueError, TypeError) as err: print(warning); raw_scores = {}

plugins/code/skills/plan-validate/scripts/validate_plan.py
shafty023, line 1: **[P1]** Missing reverse-direction sync check for answeredQuestions

All other sync pairs (pendingTasks, completedTasks, manualTasks, openQuestions) check both directions: JSON→content AND content→JSON. The answeredQuestions block only checks json_answered_q - content_answered_q, missing the reverse. If plan.json has empty answeredQuestions but the markdown content contains checked answered-question items, the sync validation silently passes.



> **Suggestion:** Add the reverse direction check: for qid in content_answered_q - json_answered_q: issues.append(f"Sync error: content has answered question {qid} but not in answeredQuestions array")

plugins/judges/skills/run-judges/scripts/test_aggregate_judge_report.py
```diff
@@ -0,0 +1,304 @@
...
+from pathlib import Path
+from typing import Any
+
+from judge_report_contract import (
+    aggregate_results,
+    load_manifest,
+    main,
+    validate_report,
+)
+from test_validate_judge_report import create_valid_casescore
```
shafty023: **[P2]** Cross-test-module import creates cascading collection failure

Imports create_valid_casescore directly from test_validate_judge_report, which reads the manifest at module level. If the manifest is absent, both test modules fail to collect with confusing import errors.



> **Suggestion:** Extract the helper into a shared conftest.py or duplicate it inline.

plugins/judges/skills/run-judges/scripts/resolve_investigation_log.sh
```diff
@@ -0,0 +1,71 @@
...
+# call again with --pre-explorer-failed to get the fallback action.
+#
+# Output (stdout JSON):
+#   resolution: already_exists | try_pre_explorer | internal_fallback | continue_without
+#   instructions: human-readable instruction for the orchestrator
+#   canonical_sections: (only for internal_fallback) section headers to use
+#
+# Exit 0 always (best-effort, never blocks).
+
+set -euo pipefail
```
shafty023, line 16: **[P2]** set -euo pipefail contradicts stated 'Exit 0 always' contract

The script header states 'Exit 0 always (best-effort, never blocks)' but set -euo pipefail will cause non-zero exits if jq fails. No trap 'exit 0' EXIT guard enforces the stated contract.



> **Suggestion:** Add trap 'exit 0' EXIT after set -euo pipefail.

plugins/judges/skills/run-judges/scripts/test_preflight_judges.sh
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/env bash
+# Test harness for preflight_judges.sh.
+# Creates temp directories with various file combinations, runs the script,
+# and asserts JSON output via jq.
+#
+# Usage: bash test_preflight_judges.sh
+# Exit 0 = all passed, exit 1 = at least one failure.
+
+set -euo pipefail
```
shafty023, line 9: **[P2]** Test harness aborts remaining tests on unexpected preflight exit

set -euo pipefail without || true guards means if preflight_judges.sh exits non-zero unexpectedly, all subsequent test functions are silently skipped.



> **Suggestion:** Wrap run_preflight calls with || true or use set +e around the run.

plugins/judges/skills/run-judges/scripts/test_preflight_judges.sh
```diff
@@ -0,0 +1,316 @@
...
+  echo "--- test_plan_cache_hit ---"
+  local w; w=$(setup_workdir)
+  echo '# PRD' > "$w/prd.md"
+  echo '{}' > "$w/plan.json"
+  sleep 1
+  cat > "$w/plan-evaluation.json" <<'EOF'
+{"simple_mode": true, "selected_critics": ["a","b"], "evaluation_summary": "All good"}
+EOF
+
+  run_preflight plan "$w"
```
shafty023: **[P2]** test_plan_cache_hit depends on cross-plugin script without skip guard

The test depends on plugins/code/scripts/check_eval_cache.sh existing. In isolated test environments without the code plugin, the test fails silently with no skip message.



> **Suggestion:** Add an existence guard that skips the test when the cross-plugin script is unavailable.

CHANGELOG.md
shafty023, line 1: **[P2]** CHANGELOG documents four-level fallback but code implements three

Changelog says 'four-level fallback strategy' including 'repo-relative path', but resolve_judges_agents_dir() in run-loop.sh implements only three levels — the 'repo-relative path' step is absent.



> **Suggestion:** Correct the CHANGELOG to say 'three-level fallback' or implement the missing level.

plugins/judges/tools/python/test_aggregate_judge_report.py
```diff
@@ -0,0 +1,304 @@
...
+
+from judge_report_contract import (
+    aggregate_results,
+    load_manifest,
+    main,
+    validate_report,
+)
+from test_validate_judge_report import create_valid_casescore
+
+MANIFEST_PATH = Path(__file__).resolve().parents[2] / "agents" / "judge-manifest.json"
```
shafty023: **[P2]** MANIFEST_PATH duplicated across test files instead of using existing helper

Both test_aggregate_judge_report.py and test_validate_judge_report.py define the identical MANIFEST_PATH computation inline, while judge_report_contract.get_default_manifest_path() already provides this exact functionality.



> **Suggestion:** Replace inline path with from judge_report_contract import get_default_manifest_path.

plugins/code/.claude-plugin/plugin.json
aponamarev: Fixed

plugins/code/skills/plan-validate/SKILL.md
aponamarev: Great point. This means that this python script needs to be moved to skill's script folder. This will enable to reference this script relative to the skill's folder. Fixed

plugins/judges/skills/run-judges/scripts/judge_report_contract.py
aponamarev: Fixed

plugins/judges/skills/run-judges/scripts/test_aggregate_judge_report.py
aponamarev: Fixed

plugins/judges/skills/run-judges/scripts/resolve_investigation_log.sh
aponamarev: Fixed

plugins/judges/skills/run-judges/scripts/test_preflight_judges.sh
aponamarev: Fixed

plugins/judges/skills/run-judges/scripts/test_preflight_judges.sh
aponamarev: Removed this dependency

plugins/judges/tools/python/test_aggregate_judge_report.py
aponamarev: Fixed

CHANGELOG.md
aponamarev: Fixed

plugins/code/skills/plan-validate/scripts/validate_plan.py
aponamarev: Fixed

plugins/judges/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "judges",
   "description": "LLM Judges plugin",
-  "version": "1.1.0",
+  "version": "2.0.1",
```
shafty023, line 4: This should be `2.0.0`

aponamarev: Huh... unexpected. This is outdated. The version is set to 2.0.0.

shafty023:
**[P1]** Aggregation step reads combined results file that is never created

**plugins/judges/skills/run-judges/SKILL.md:272**

The per-judge validation section instructs agents to write each judge's output to $CLOSEDLOOP_WORKDIR/judge-output-{judge_id}.json (one file per judge). The aggregation step then calls judge_report_contract.py aggregate --results-path "$CLOSEDLOOP_WORKDIR/judge-results-${ARTIFACT_TYPE}.json". There is no instruction that collects the individual per-judge files into the combined judge-results-${ARTIFACT_TYPE}.json before aggregation. The aggregate script handles a missing file by marking all judges as errored (final_status=3), so every run will produce a report where all judges appear to have failed.



> **Suggestion:** Add an explicit step between validation and aggregation that collects all validated judge-output-{judge_id}.json payloads into $CLOSEDLOOP_WORKDIR/judge-results-${ARTIFACT_TYPE}.json in the format {"results": [{"judge_id": "...", "result": <CaseScore>}, ...]}

shafty023:
**[P1]** Skill usage instruction drops base-directory prefix, breaking path resolution

**plugins/code/skills/plan-validate/SKILL.md:196**

The PR removes the explanatory text and <base_directory>/ prefix from the usage instruction, changing python3 <base_directory>/scripts/validate_plan.py <WORKDIR> to python3 scripts/validate_plan.py <WORKDIR>. Every other skill uses the <base_directory>/scripts/... pattern. An agent running this skill from the repo root will get a file-not-found error.



> **Suggestion:** Restore the explicit path: python3 <base_directory>/scripts/validate_plan.py <WORKDIR> with the context note, or use the full repo-relative path.

shafty023:
**[P2]** Relative path in usage example lost base-directory context hint

**plugins/code/skills/plan-validate/SKILL.md:22**

Same root cause as the HIGH finding — the bare relative path scripts/validate_plan.py in the usage section is insufficient without the removed <base_directory>/ prefix context.



> **Suggestion:** See HIGH finding for SKILL.md:196.

shafty023:
**[P2]** test_registry_values_are_deduplicated asserts a trivially true inequality

**plugins/judges/skills/run-judges/scripts/test_validate_judge_report.py:845**

The assertion assert len(JUDGE_REGISTRY[category]) <= len(config['judges']) is always True because JUDGE_REGISTRY[category] is built as set(config['judges']), and len(set(L)) <= len(L) holds for any list. The test passes unconditionally regardless of duplicates.



> **Suggestion:** Change <= to == so the assertion actually verifies no duplicates exist in the manifest.

shafty023:
**[P2]** fallback_mode.reason hardcoded incorrectly when plan-context.json exists

**plugins/judges/skills/run-judges/scripts/build_judge_input.py:188**

When compatibility_mode=True and plan-context.json exists on disk, _build_plan_compatibility hardcodes reason='plan-context.json unavailable' which is factually incorrect.



> **Suggestion:** Compute the reason dynamically: 'plan-context.json unavailable; ...' if not plan_ctx.is_file() else 'compatibility mode explicitly requested; using raw plan.json + prd.md'.

PR #20, opened 2026-03-11
Export token usage from stream_formatter.py

mikeangstadt:
## Summary
- The harness-agent parses stdout for per-model token usage lines to populate `tokensUsed` and `tokensByModel` in completed events
- For PLAN/EXECUTE commands, `run-loop.sh` uses `--output-format stream-json` piped through `stream_formatter.py`, which never printed usage summary lines
- This resulted in all loops having zero token counts and $0.00 cost in the UI

`stream_formatter.py` now accumulates usage from `assistant` events and prints a per-model summary plus totals at the end, matching the regex format the harness expects:
```
Model: claude-opus-4-6  Input: 12345  Output: 6789  Cache creation: 100  Cache read: 200
Total input tokens: 12345
Total output tokens: 6789
```

## Test plan
- [ ] Run a PLAN or EXECUTE loop and verify token usage appears in completed event
- [ ] Verify the loops table shows non-zero token counts for new loops
- [ ] Verify cost calculation works with the tokensByModel data

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### code v1.1.3
+
+#### Added
+- `stream_formatter.py` now accumulates per-model token usage from assistant events and prints a summary in the format the harness expects, fixing zero token counts for PLAN/EXECUTE loops
+
+#### Fixed
+- `stream_formatter.py` returns early on `BrokenPipeError` before printing usage summary, preventing tracebacks when used in pipelines with early-exit consumers
+
 ### code v1.1.2
 
 #### Fixed
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.1.2",
+  "version": "1.1.3",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.2.1",
+  "version": "1.1.3",
   "author": {
     "name": "ClosedLoop",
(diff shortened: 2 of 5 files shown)


plugins/code/tools/python/stream_formatter.py
```diff
@@ -219,6 +272,8 @@ def main() -> int:
     except BrokenPipeError:
         pass
 
+    _print_usage_summary(tokens_by_model)
+
```
shafty023, line 276: [P2] Return immediately after a broken pipe before printing the summary.

When `stream_formatter.py` is used in a pipeline whose consumer stops reading early (for example `... | head`), the `BrokenPipeError` from the main event loop is swallowed, but this unconditional `_print_usage_summary()` call still writes to closed stdout. That produces a traceback instead of exiting quietly, which is a regression for a command used as a Unix filter.

mikeangstadt: Fixed in 890c7fa — BrokenPipeError now returns early before the summary.

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.1.2",
+  "version": "1.2.1",
```
shafty023: Why did we jump from 1.1.2 -> 1.2.1 ? Should have been 1.2 if anything

CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### code v1.1.3
```
shafty023, line 9: This does not match the plugin version in `plugins/code/.claude-plugin/plugin.json`

plugins/code/.claude-plugin/plugin.json
mikeangstadt: Fixed in ecd12aa — version corrected to 1.1.3 (PATCH bump from 1.1.2, since this is a bug fix).

CHANGELOG.md
mikeangstadt: Fixed in ecd12aa — plugin.json now reads 1.1.3, matching this CHANGELOG entry.

mikeangstadt: other place got outdated, both places match now.

shafty023 (approved):

PR #27, opened 2026-03-24
FEAT-151: add feedback-explorer delta caching and deferral detection

shafty023:
Add delta caching to feedback-explorer so it reuses prior context between debate rounds instead of re-fetching everything from scratch. Cache uses normalized reference keys (file:, symbol:, pattern:, critical-file:) with a stamp (round/session/HEAD) for validation.

Also adds deferral detection to plan-with-codex -- scans plans for "Deferred", "Out of Scope", "Future Work" items and requires explicit user approval before excluding work. Consolidates per-plugin CHANGELOGs into the root CHANGELOG.md.


aponamarev (approved):
LGTM

mikeangstadt (approved):
Do we want to emit any new usage metrics around this for the stored/used cached context?

shafty023:
> Do we want to emit any new usage metrics around this for the stored/used cached context?

@mikeangstadt if you're talking about token counts the LLM doesn't know about its token usage. That's received only in a hook input so something we'd handle in a separate PR

PR #33, opened 2026-03-26
FEAT-181: remove self-learning write references from agent prompts

shafty023:
Remove learning capture sections, Organization Learnings sections, and self-learning:learning-quality skill references from 10 agent prompts. Delete 3 learning guidance prompt files that are no longer referenced.

These were missed when self-learning was disabled by default in 95a26b3. Agents were still wasting tokens reading guidance docs and writing learning JSON files that are never processed.


mikeangstadt (approved):

aponamarev (commented):
LGTM

PR #35, opened 2026-03-27
FEAT-190: add PR auto-detection, fix partition cap, add small-diff fast path

shafty023:
Local mode now auto-detects open PRs via `gh pr view` and uses the PR diff scope instead of `main...HEAD`, preventing scope explosion on diverged branches. The partition cap enforcement loop is fixed with a force-merge fallback that guarantees partition count never exceeds max_bha_agents. Small diffs (<=150 LOC, <=5 files) now use a single Sonnet reviewer instead of the full 5-agent fleet.

Also adds a round-aware severity gate to the codex-review script so later debate rounds only flag functionally wrong behavior, and fixes truncated-response detection in verdict extraction.

These changes together reduce review time for small PRs from ~33 minutes to ~1-2 minutes.

### DIFF
--- a/plugins/code-review/tools/python/code_review_helpers.py
+++ b/plugins/code-review/tools/python/code_review_helpers.py
@@ -93,6 +93,10 @@ DEFAULT_MAX_BHA_AGENTS = 5
 REBALANCE_LOC_BUDGET = 1200
 MIXED_PARTITION_SPLIT_THRESHOLD = 50
 
+# Fast-path routing thresholds
+FAST_PATH_MAX_LOC = 150
+FAST_PATH_MAX_FILES = 5
+
 # Validation thresholds
 CONFIDENCE_DISCARD_THRESHOLD = 0.5
 LINE_TOLERANCE = 3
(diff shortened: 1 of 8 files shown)


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -93,6 +93,10 @@
 REBALANCE_LOC_BUDGET = 1200
 MIXED_PARTITION_SPLIT_THRESHOLD = 50
 
+# Fast-path routing thresholds
+FAST_PATH_MAX_LOC = 150
```
mikeangstadt, line 97: we should make these configurable.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -744,40 +805,60 @@ def _flush_partition() -> None:
...
-                # Remove the two originals (remove higher index first)
-                remove_indices = sorted([idx_a, idx_b], reverse=True)
-                for ri in remove_indices:
-                    partitions.pop(ri)
-                partitions.append(new_part)
-                merged = True
-                break
-        if not merged:
-            break  # No valid same-type merge possible, stop
+            for ai in range(len(same_type)):
```
mikeangstadt, line 819: hash table makes this O(n) instead of O(n^2)

shafty023: Not a concern here -- n is the partition count, which is capped at max_bha_agents (5-6). Even in the worst case before cap enforcement, we are looking at ~30 partitions, so the nested loop iterates over maybe 15-30 pairs per iteration.

More importantly, a hash table would not help with this problem. We are finding the minimum-LOC pair that satisfies two constraints (merged_loc <= 1200 AND merged_file_count <= max_files). That is a constrained minimum search, not a lookup -- you cannot hash your way out of checking pairs against two constraints.

Total work across all while-loop iterations is O(n^3) worst case with n <= 30, which is sub-millisecond. Not worth adding complexity for.

mikeangstadt (approved):

plugins/code-review/tools/python/code_review_helpers.py
shafty023: Good idea. Tracked as [FEAT-191](https://app.closedloop.ai/features/FEAT-191) -- we will add CLI args for all hardcoded routing/partition constants: --fast-path-max-loc, --fast-path-max-files, --size-small, --size-medium, --rebalance-loc-budget, --trivial-partition-threshold, and --mixed-partition-split-threshold. Same pattern as the existing --loc-budget and --max-files args. Keeps the helpers stateless with the orchestrator as single source of truth for tuning. Low priority since no one has needed to tune these yet and defaults preserve current behavior.

PR #40, opened 2026-04-02
feat: add one-line bootstrap install script

mikeangstadt:
## Summary
- Adds `install.sh` bootstrap script that installs all 6 plugins globally via `curl | bash` one-liner
- Script checks prerequisites (Claude CLI, Python 3.11+, jq), registers the marketplace, and installs all plugins at user scope with auto-update
- Updates README Quick Start to lead with the one-liner and reorganizes the section for clarity

## Test plan
- [ ] Run `bash install.sh` locally and verify all 6 plugins install
- [ ] Verify `curl -fsSL https://raw.githubusercontent.com/closedloop-ai/claude-plugins/main/install.sh | bash` works after merge
- [ ] Confirm plugins appear in `claude /plugin` installed tab
- [ ] Verify auto-update picks up new versions on next session start

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,126 @@
+#!/usr/bin/env bash
+# ClosedLoop Claude Plugins - One-Line Installer
+#
+# Usage:
+#   curl -fsSL https://raw.githubusercontent.com/closedloop-ai/claude-plugins/main/install.sh | bash
+#
+# What this does:
+#   1. Checks prerequisites (Claude Code CLI, Python 3.11+, jq)
+#   2. Registers the closedloop-ai plugin marketplace
+#   3. Installs all 6 plugins globally (user scope)
+#   4. Auto-update is enabled by default — plugins stay current automatically
+#
+set -euo pipefail
+
+# ── Colors ───────────────────────────────────────────────────────────────────
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+BLUE='\033[0;34m'
+BOLD='\033[1m'
+NC='\033[0m'
+
+info()  { echo -e "${GREEN}[✓]${NC} $1"; }
+warn()  { echo -e "${YELLOW}[!]${NC} $1"; }
+err()   { echo -e "${RED}[✗]${NC} $1"; }
+step()  { echo -e "${BLUE}[→]${NC} ${BOLD}$1${NC}"; }
+
+# ── Constants ────────────────────────────────────────────────────────────────
+MARKETPLACE_SOURCE="closedloop-ai/claude-plugins"
+MARKETPLACE_NAME="closedloop-ai"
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -0,0 +1,126 @@
...
+    exit 1
+fi
+info "Claude Code CLI found: $(claude --version 2>/dev/null || echo 'unknown version')"
+
+# Python 3.11+
+if command -v python3 &>/dev/null; then
+    PY_VERSION=$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
+    PY_MAJOR=$(echo "$PY_VERSION" | cut -d. -f1)
+    PY_MINOR=$(echo "$PY_VERSION" | cut -d. -f2)
+    if [[ "$PY_MAJOR" -ge 3 && "$PY_MINOR" -ge 11 ]]; then
```
shafty023: **[P3]** Accept Python 4+ in version prerequisite check

The version gate `[[ "$PY_MAJOR" -ge 3 && "$PY_MINOR" -ge 11 ]]` incorrectly marks Python 4.x as below the `3.11+` requirement because it still requires `minor >= 11` even when `major > 3`. On environments that ship Python 4 in the future, the installer will emit a false warning about unsupported Python, which is misleading for users and makes the prerequisite check inaccurate.

install.sh
```diff
@@ -0,0 +1,126 @@
...
+step "Installing plugins (user scope)..."
+
+INSTALLED=0
+FAILED=0
+
+for plugin in "${PLUGINS[@]}"; do
+    PLUGIN_REF="${plugin}@${MARKETPLACE_NAME}"
+    if claude plugin install "$PLUGIN_REF" --scope user 2>/dev/null; then
+        info "Installed: $plugin"
+        ((INSTALLED++))
```
peterulsteen: A quirk of `((INSTALLED++))` is that if the initial value of `INSTALLED` is 0, the expression evaluates to 0, which results in a non-zero exit status. Combined with `set -euo pipefail` on line 13, this will kill the script on the first successful install. Same issue for `((FAILED++))`.

peterulsteen (changes requested):

install.sh
mikeangstadt: Good catch — fixed in d286719. Changed to `[[ "$PY_MAJOR" -gt 3 || ( "$PY_MAJOR" -eq 3 && "$PY_MINOR" -ge 11 ) ]]` so Python 4+ passes the check.

install.sh
mikeangstadt: Nice find — that `set -euo pipefail` interaction is subtle. Fixed in d286719 by replacing all `((VAR++))` with `VAR=$((VAR + 1))` which always returns exit status 0.

peterulsteen (approved):

PR #44, opened 2026-04-08
FEA-107: Add PR validation gates, lockfile, and flaky test quarantine

shafty023:
- Create requirements-dev.txt with pinned direct dev/CI dependencies (anthropic, mcp, pyright, pytest, PyYAML, ruff). Every CI job now installs via pip install -r requirements-dev.txt followed by pip check, which validates the resolved dependency graph is internally consistent and satisfies FEA-107 AC-004.1 (install must succeed without warnings or missing packages). File header documents the regeneration procedure for manual bumps.
- Add [tool.pytest.ini_options] to pyproject.toml registering two markers: contract (repo-level contract tests) and quarantine (flaky, non-blocking). Registering the markers also silences pytest's PytestUnknownMarkWarning so future runs are warning-free. Also sets testpaths = ["plugins"].
- Rewrite .github/workflows/ci.yml with six jobs:
  - Lint: ruff check .
  - Type Check: pyright
  - Tests: pytest plugins/ -m 'not quarantine' (quarantined tests are now excluded from the blocking run)
  - Contract tests: pytest plugins/ -m contract, wrapped to treat exit code 5 (no tests collected) as success -- passes vacuously until tests are marked
  - Quarantined tests (non-blocking): pytest plugins/ -m quarantine with continue-on-error: true, also handling exit 5 gracefully
  - PR gates: aggregator with needs: [lint, typecheck, test, contract-tests] and if: always(); lets branch protection require one check long-term instead of four.
- Add a concurrency group so stacked PR updates cancel in-flight runs.
- Add docs/flaky-test-quarantine.md documenting the @pytest.mark.contract and @pytest.mark.quarantine conventions, how to quarantine/un-quarantine, and the policy for rolling review.
- Add docs/ci-gates.md documenting the gate matrix and the gh api commands to update ruleset 13555155 once the new check contexts have been observed on a real PR run. The ruleset currently has zero required status checks.

Testing: Ran the full suite locally against requirements-dev.txt in a Python 3.14 venv: pip check clean, ruff check . clean, pyright clean (0 errors, 0 warnings), pytest plugins/ -m 'not quarantine' = 577 passed, pytest plugins/ -m contract = exit 5 (wrapper treats as success), same for -m quarantine.


.github/workflows/ci.yml
```diff
@@ -36,5 +42,74 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - run: pip install -r requirements-dev.txt
+      - run: pip check
+      - name: Run contract tests
+        shell: bash
+        run: |
+          set +e
+          pytest plugins/ -m contract
```
peterulsteen: A test decorated with both `@pytest.mark.contract` and `@pytest.mark.quarantine` is correctly excluded from the blocking `test` job on line 47, but will still run here in `contract-tests`, which is also blocking according to line 103.

Should we change this to `pytest plugins/ -m 'contract and not quarantine'` so the quarantine override is consistent? Tests marked with both would then only run in the non-blocking `quarantine` job.

wongk (changes requested):
same comments here regarding contract tests and quarantine.

.github/workflows/ci.yml
shafty023: ended up reverting the contract-tests and quarantine jobs entirely per other review feedback (commit 3030535), so this inconsistency no longer exists.

wongk (approved):
i believe this repo also does not have code review?

PR #45, opened 2026-04-09
PLN-202: Add multi-repo support for planning and exploration

aponamarev:
## Summary

- Add `--add-dir` flag to `run-loop.sh` enabling multi-repository planning workflows. Secondary repos are passed via `CLOSEDLOOP_ADD_DIRS` and `CLOSEDLOOP_REPO_MAP` environment variables.
- Extend `pre-explorer` agent to generate per-repo code maps (`code-map-{name}.json`) for secondary repositories, and `plan-draft-writer` to produce multi-repo plans with `## Repositories` table and `@{repo}:path` file prefixes.
- Enhance `discover-repos.sh` with Tier 0 explicit directory support, path deduplication helpers, and a `local: true` marker on peers added via `--add-dir` (consumed by `cross-repo-prd-writer` to exclude local peers from PRD generation).
- Add `repo` field to plan schema for multi-repo task targeting and update cross-repo agents for enhanced context. Multi-repo behavior is owned by the agents themselves — the orchestrator prompt is unchanged.
- Add tests for `discover-repos.sh` and `setup-closedloop.sh` scripts.
- Added `CLOSEDLOOP_STATE_DIR` as a single source of truth for state directory name — `.closedloop-ai`.

## Changed Files

| Area | Files |
|------|-------|
| CLI / orchestration | `run-loop.sh`, `setup-closedloop.sh` |
| Agents | `pre-explorer.md`, `plan-draft-writer.md`, `plan-evaluator.md`, `cross-repo-coordinator.md`, `cross-repo-prd-writer.md` |
| Schema | `plan-schema.json` |
| Discovery | `discover-repos.sh` |
| Hooks | `subagent-start-hook.sh` |
| Tests | `test_discover_repos.py`, `test_setup_closedloop.py`, `test_validate_plan.py` (new) |
| Config | `plugin.json` (version bump to 1.10.0) |

## Test plan

- [x] Run `pytest plugins/code/tools/python/test_discover_repos.py` — verify discover-repos tests pass
- [x] Run `pytest plugins/code/tools/python/test_setup_closedloop.py` — verify setup-closedloop tests pass
- [x] Run `run-loop.sh --add-dir /path/to/secondary-repo --prd test.md` — verify multi-repo env vars are exported
   - Mono-repo approach [plan.json](https://github.com/user-attachments/files/26613483/plan.json) and [plan.md](https://github.com/user-attachments/files/26613484/plan.md)
   - Multi-repo approach [plan.json](https://github.com/user-attachments/files/26613511/plan.json) and [plan.md](https://github.com/user-attachments/files/26613512/plan.md)

- [x] Verify pre-explorer generates `code-map-{name}.json` for secondary repos

---
Loop ID: 019d7291-c472-706c-bf30-19de89cc4afe
Artifact: https://app.closedloop.ai/implementation-plans/PLN-202


plugins/code/scripts/setup-closedloop.sh
```diff
@@ -159,5 +215,24 @@ CLAUDE_PLUGIN_ROOT="$PLUGIN_ROOT"
 CLOSEDLOOP_PROMPT_FILE="$CLOSEDLOOP_PROMPT_FILE"
 EOF
 
+# Build pipe-joined multi-repo variables (empty strings when no extra repos)
+add_dirs_joined=""
+add_dir_names_joined=""
+repo_map_joined=""
+if [[ ${#RESOLVED_ADD_DIRS[@]} -gt 0 ]]; then
+    add_dirs_joined="$(IFS='|'; echo "${RESOLVED_ADD_DIRS[*]}")"
```
shafty023, line 284: **[P2]** Reject duplicate repo names before building CLOSEDLOOP_REPO_MAP

If two `--add-dir` repos resolve to the same short name — either because they share a basename or because their `.repo-identity.json` files both say `name: service` — these lines emit duplicate `service=...` entries. Downstream multi-repo agents treat `{name}` as the unique repo key, the `@{name}:path` prefix, and the stem of `code-map-{name}.json`, so the second repo silently overwrites the first and the resulting plan can point tasks at the wrong repository.

plugins/code/scripts/setup-closedloop.sh
```diff
@@ -51,13 +62,29 @@ while [[ $# -gt 0 ]]; do
...
+    if ! abs_path="$(cd "$raw_dir" 2>/dev/null && pwd)"; then
+        echo "Error: --add-dir path does not exist or is not a directory: $raw_dir" >&2
+        exit 1
+    fi
+    identity_file="$abs_path/.closedloop-ai/.repo-identity.json"
+    repo_name="$(jq -r '.name // empty' "$identity_file" 2>/dev/null || true)"
+    if [[ -z "$repo_name" ]]; then
+        repo_name="$(basename "$abs_path")"
+    fi
+    RESOLVED_ADD_DIRS+=("$abs_path")
```
shafty023, line 173: **[P2]** Drop the primary workdir from resolved --add-dir repos

If the caller passes `--add-dir` with the same directory as `WORKDIR` (for example `--add-dir .`), we still append it here and later publish it in `CLOSEDLOOP_REPO_MAP`. `discover-repos.sh` explicitly skips the current repo, but `pre-explorer` and `plan-draft-writer` read `CLOSEDLOOP_REPO_MAP` directly, so the primary repo gets explored a second time and its files can be mislabeled with a secondary `@repo:` prefix in the generated plan.
The new multi-repo flow has two correctness gaps in `setup-closedloop.sh`: it does not guarantee unique secondary repo identifiers, and it does not filter the primary workdir out of `--add-dir`. Both issues can cause downstream agents to read or label the wrong repository during plan generation.

aponamarev: Added repo path and name deduplication. Fixed

plugins/code/scripts/setup-closedloop.sh
aponamarev: Added repo path and name deduplication

CHANGELOG.md
```diff
@@ -14,6 +14,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Updated agent output path references from `.claude/runs/` to `.closedloop-ai/runs/` in `agent-prompt-generator`
 - Updated bootstrap configuration documentation in `agent-bootstrap.md` to reference `.closedloop-ai/` state directory
 
+### code v1.7.0
```
shafty023: Looks like we missed 1.7.0 and 1.8.0 here from a previous commit. The current version is 1.8.0 and your PR needs to bump that to 1.9.0

plugins/code/scripts/discover-repos.sh
```diff
@@ -17,20 +21,57 @@ else
...
+            repo_type=$(jq -r '.type // "unknown"' "$identity_file")
+        fi
+        repo_name="${repo_name:-$(basename "$path")}"
+
+        PEER_JSONS+=("{\"name\": \"$repo_name\", \"type\": \"$repo_type\", \"path\": \"$path\", \"discoveryMethod\": \"add_dir\"}")
+    done
+fi
 
 # Tier 1: Environment variable
+DISCOVERY_METHOD="sibling_scan"
```
shafty023, line 74: **[P2]** Root-level discoveryMethod incorrectly reports 'sibling_scan' for Tier 0 only discovery

DISCOVERY_METHOD='sibling_scan' is set as the default before checking CLAUDE_WORKSPACE_REPOS (Tier 1). When CLOSEDLOOP_ADD_DIRS is set but CLAUDE_WORKSPACE_REPOS is not, only Tier 0 peers are discovered but the root JSON field still reads 'sibling_scan'.



> **Suggestion:** After the Tier 0 loop, set DISCOVERY_METHOD='add_dir' if PEER_JSONS was populated by Tier 0 and no Tier 1 env var is present.

plugins/code/scripts/run-loop.sh
```diff
@@ -655,10 +671,10 @@ STOPPING:
 
 MONITORING:
   # View current iteration:
-  grep '^iteration:' .closedloop-ai/closedloop-loop.local.md
+  grep '^iteration:' $CLOSEDLOOP_STATE_DIR/closedloop-loop.local.md
```
shafty023: **[P2]** $CLOSEDLOOP_STATE_DIR not expanded in single-quoted heredoc in show_help()

The show_help() function uses a single-quoted heredoc delimiter (cat << 'EOF'), which suppresses all variable expansion. The diff replaces three literal .closedloop-ai strings with $CLOSEDLOOP_STATE_DIR references that will appear verbatim in help output instead of expanding. Users copying MONITORING commands from --help will get non-functional commands.



> **Suggestion:** Change the heredoc delimiter from << 'EOF' to << EOF (unquoted) so that $CLOSEDLOOP_STATE_DIR expands at runtime, or revert to the literal string .closedloop-ai for these three lines.

plugins/self-learning/scripts/bootstrap-learnings.sh
```diff
@@ -10,11 +10,14 @@
...
+CLOSEDLOOP_STATE_DIR=".closedloop-ai"
+
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-LEARNINGS_DIR="${1:-.closedloop-ai/learnings}"
+LEARNINGS_DIR="${1:-$CLOSEDLOOP_STATE_DIR/learnings}"
 
 # Derive PROJECT_DIR for gitignore updates (only if using default path)
-if [[ "$LEARNINGS_DIR" == ".closedloop-ai/learnings" ]] || [[ "$LEARNINGS_DIR" == *"/.closedloop-ai/learnings" ]]; then
+if [[ "$LEARNINGS_DIR" == "$CLOSEDLOOP_STATE_DIR/learnings" ]] || [[ "$LEARNINGS_DIR" == *"/$CLOSEDLOOP_STATE_DIR/learnings" ]]; then
     PROJECT_DIR="${LEARNINGS_DIR%/.closedloop-ai/learnings}"
```
shafty023: **[P2]** Incomplete SSOT migration: hardcoded .closedloop-ai in string-trimming after variable introduced

The PR introduces CLOSEDLOOP_STATE_DIR='.closedloop-ai' on line 14 as the SSOT and updates line 20 to use the variable, but line 21's string-trimming operation (${LEARNINGS_DIR%/.closedloop-ai/learnings}) still contains the hardcoded literal. If CLOSEDLOOP_STATE_DIR ever changes, the suffix removal will silently fail.



> **Suggestion:** Change line 21 to: PROJECT_DIR="${LEARNINGS_DIR%/$CLOSEDLOOP_STATE_DIR/learnings}"

plugins/self-learning/tools/python/test_compute_success_rates.py
```diff
@@ -17,6 +16,8 @@
     serialize_toon,
 )
 
+CLOSEDLOOP_STATE_DIR = ".closedloop-ai"
```
shafty023: **[P2]** CLOSEDLOOP_STATE_DIR duplicated in test file instead of imported from shared module

test_compute_success_rates.py independently defines CLOSEDLOOP_STATE_DIR = '.closedloop-ai' at line 19. The same constant already exists in compute_success_rates.py. The code plugin solved this pattern with conftest.py, but self-learning has no equivalent, creating two independent copies.



> **Suggestion:** Either import the constant from compute_success_rates (from compute_success_rates import CLOSEDLOOP_STATE_DIR) or add a conftest.py at plugins/self-learning/tools/python/conftest.py.

plugins/code/prompts/overlays/README.md
```diff
@@ -0,0 +1,60 @@
...
+  repositories.
+- `CLOSEDLOOP_ADD_DIRS` — pipe-separated absolute paths of local peer
+  repos.
+- `CLOSEDLOOP_ADD_DIR_NAMES` — pipe-separated names matching
+  `CLOSEDLOOP_ADD_DIRS` by index.
+
+The overlay introduces the `@{repo-name}:path` file-reference convention
+for secondary repos (primary-repo files need no prefix).
+
+`run-loop.sh --add-dir` auto-selects `--prompt multi-repo` when the user
```
shafty023: **[P2]** README claims --add-dir auto-selects --prompt multi-repo but no such code exists

The README states 'run-loop.sh --add-dir auto-selects --prompt multi-repo when the user does not pass --prompt explicitly.' Inspection of run-loop.sh shows no auto-selection logic -- PROMPT_NAME is only set when --prompt is passed explicitly. Users following this documentation will silently miss multi-repo amendments.



> **Suggestion:** Either add the auto-selection logic to run-loop.sh (if ADD_DIRS is non-empty and PROMPT_NAME is empty, set PROMPT_NAME to 'multi-repo') or correct the README to remove the claim.

aponamarev: Auto-selection logic does exist; it just lives in a different file than the one you inspected.

**Step 1 — How `run-loop.sh` handles `--prompt` when the user omits it**

You're correct that `run-loop.sh` itself has no auto-selection logic. `PROMPT_NAME` is initialized empty at and is only set when the user passes `--prompt` explicitly (line 731).

**Step 2 — What `run-loop.sh` does with `--add-dir`**

In `create_state_file()`, `run-loop.sh` builds the command it will hand to `claude -p`:

```bash
local prompt="/code:code $WORKDIR"
if [[ -n "$PROMPT_NAME" ]]; then
  prompt="$prompt --prompt $PROMPT_NAME"
fi
...
for add_dir in "${ADD_DIRS[@]+"${ADD_DIRS[@]}"}"; do
  prompt="$prompt --add-dir $add_dir"
done
```

Key observation: when the user omits `--prompt`, the `--prompt` flag is *not* appended — but every `--add-dir` *is*. The command that reaches Claude is therefore `/code:code <workdir> --add-dir <path> ...` with no `--prompt`.

**Step 3 — The `/code:code` slash command invokes `setup-closedloop.sh`**

`plugins/code/commands/code.md` executes `setup-closedloop.sh $ARGUMENTS`, forwarding every flag — including `--add-dir` and the (now absent) `--prompt`.

**Step 4 — `setup-closedloop.sh` is where the auto-selection actually lives**

At `plugins/code/scripts/setup-closedloop.sh`, `PROMPT_NAME_EXPLICIT` defaults to `false` and is only flipped to `true` when the argument parser sees an explicit `--prompt`.:

```bash
# Step 3: Auto-select prompt based on whether extra repos were provided
if [[ "$PROMPT_NAME_EXPLICIT" == false ]]; then
    if [[ ${#RESOLVED_ADD_DIRS[@]} -gt 0 ]]; then
        PROMPT_NAME="multi-repo"
    else
        PROMPT_NAME="${PROMPT_NAME:-prompt}"
    fi
fi
```

So `run-loop.sh --add-dir X` (without `--prompt`) → `/code:code <workdir> --add-dir X` → `setup-closedloop.sh` with no explicit `--prompt` → auto-selects `multi-repo`. The end-to-end claim in the README is correct.

**Why I'm not adding the logic to `run-loop.sh`**

Duplicating the auto-selection into `run-loop.sh` would create two sources of truth for prompt resolution and risk them drifting apart. `setup-closedloop.sh` is the single entry point that every invocation path (`run-loop.sh`, direct `/code:code`, and resumed loops) funnels through, so it's the right home for this logic.

aponamarev: Follow-up: tightened the README wording in `be337ef` per the concession in my previous reply. The section now reads:

> When `run-loop.sh --add-dir` is used without `--prompt`,
> `setup-closedloop.sh` auto-selects the `multi-repo` overlay (see the
> `PROMPT_NAME_EXPLICIT` branch in `setup-closedloop.sh`). `run-loop.sh`
> itself performs no prompt resolution — it forwards `--add-dir` and
> `--prompt` through the `/code:code` slash command, which invokes
> `setup-closedloop.sh` as the single source of truth for overlay
> selection.

This names the correct file so a future reviewer who audits `run-loop.sh` in isolation won't hit the same dead end. Resolving this thread. Thanks again for the careful read.

plugins/self-learning/tools/python/test_compute_success_rates.py
shafty023: Addressed -- the test file now imports `CLOSEDLOOP_STATE_DIR` from `compute_success_rates` instead of defining its own copy (line 11).

plugins/code/hooks/subagent-start-hook.sh
```diff
@@ -150,6 +153,9 @@ CLOSEDLOOP_PRD_FILE=${CLOSEDLOOP_PRD_FILE:-}
 CLOSEDLOOP_PLAN_FILE=${CLOSEDLOOP_PLAN_FILE:-}
 CLOSEDLOOP_MAX_ITERATIONS=${CLOSEDLOOP_MAX_ITERATIONS:-10}
 CLAUDE_PLUGIN_ROOT=$PLUGIN_ROOT
+CLOSEDLOOP_ADD_DIRS=${CLOSEDLOOP_ADD_DIRS:-}
```
shafty023, line 156: **[P1]** Export multi-repo env vars before Bash-based agents use them

In `--add-dir` sessions, `setup-closedloop.sh` only writes `CLOSEDLOOP_REPO_MAP` and `CLOSEDLOOP_ADD_DIR_NAMES` to `config.env`; they are never exported into the Claude process. This hook now advertises those values in `<closedloop-environment>`, but the Bash snippet still only exports `CLOSEDLOOP_WORKDIR` and `CLAUDE_PLUGIN_ROOT`, so the new multi-repo steps that read `CLOSEDLOOP_REPO_MAP` from Bash (for example `pre-explorer` and `plan-draft-writer`) will see it as empty and skip the secondary repos.

plugins/code/scripts/run-loop.sh
```diff
@@ -827,6 +856,15 @@ emit_skipped_step() {
   )"
 }
 
+# Export CLOSEDLOOP_ADD_DIRS as pipe-joined string from ADD_DIRS array
+export_add_dirs() {
```
shafty023: **[P2]** Persist the resolved add-dir list instead of raw CLI args

This exports the raw `--add-dir` arguments before `setup-closedloop.sh` canonicalizes them. Downstream consumers do not read the canonicalized `config.env`; they read `CLOSEDLOOP_ADD_DIRS` from the Claude process, so paths that setup intentionally filtered out (current repo, parent of the workdir, duplicates, or cwd-relative paths) still change behavior. A concrete case is `run-loop.sh .closedloop-ai/work --add-dir .`: setup produces an empty repo map, but the new `plan-evaluator` signal still sees one add-dir and forces `simple_mode=false` anyway.
The multi-repo feature is not wired through the runtime environment consistently. Secondary-repo metadata never reaches Bash-based agents, and the exported add-dir list can disagree with the filtered config, so `--add-dir` runs can silently skip repos or take the wrong planning path.

plugins/code/hooks/subagent-start-hook.sh
aponamarev: The current setup works because the agents are not depending on shell-level environment inheritance for multi-repo context. Instead, these variables and their values are being passed directly to sub-agents via prompt injection orchestrated by `claude-output.jsonl`.

In this session (`claude-output.jsonl`), the orchestrator launched `code:pre-explorer` with an explicit prompt that included:

- `WORKDIR=/Users/alex.ponamarev/Projects/closed_loop/symphony-alpha/.closedloop-ai/work`
- `Additional repos context: CLOSEDLOOP_REPO_MAP is set to "closedloop-electron=/Users/alex.ponamarev/Projects/closed_loop/closedloop-electron"`

The same pattern was used for `code:plan-draft-writer`. Therefore, the subagents then used the concrete repo path from the prompt, not `$CLOSEDLOOP_REPO_MAP` expansion inside Bash. The session confirms this: `pre-explorer` ran Bash commands directly against `/Users/alex.ponamarev/Projects/closed_loop/closedloop-electron/...`, then successfully wrote `code-map-closedloop-electron.json`. `plan-draft-writer` later produced a `plan.json` that included a `## Repositories` section for `closedloop-electron`, a `repositories` object with that repo path, and cross-repo tasks using the `@closedloop-electron:` prefix. In other words, the multi-repo plan succeeded because the needed values were already embedded in the prompt as plain text and the agents used them as data.

So the current flow works because prompt injection is carrying the multi-repo context end-to-end, and the observation of actual `claude-output.jsonl` confirms this mechanism. So practically speaking environment vairalbes `CLOSEDLOOP_REPO_MAP` and `CLOSEDLOOP_ADD_DIR_NAMES` are currently unused due to the fact that `subagent-start-hook.sh` is injecting them directly into the prompt. However, I will export `CLOSEDLOOP_REPO_MAP` and `CLOSEDLOOP_ADD_DIR_NAMES` in case future changes require these variables to be present in the agent environment directly.

plugins/code/scripts/run-loop.sh
aponamarev: subagent-start-hook.sh sources config.env (line 43), which overwrites any process-inherited env vars with the canonicalized values from setup-closedloop.sh. Agents always see the filtered list.

That said, the real issue is the fact that export_add_dirs() in run-loop.sh exports CLOSEDLOOP_ADD_DIRS into the parent process environment where nothing consumes it. This dead export will be removed.

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.8.0",
+  "version": "1.9.1",
```
shafty023: 1.9.0. Also update `CHANGELOG.md`

plugins/code/agents/cross-repo-coordinator.md
```diff
@@ -59,6 +59,19 @@ Parse the JSON output to get:
 
 **Write the discovery result to `$CLOSEDLOOP_WORKDIR/.workspace-repos.json`** so other agents can access it.
 
+### Step 1.5: Local Repos (--add-dir)
```
shafty023: If this is truly nested under Step 1 then this should be:

    #### Step 1.1: Local Repos (--add-dir)

plugins/code/agents/cross-repo-coordinator.md
```diff
@@ -59,6 +59,19 @@ Parse the JSON output to get:
...
+
+After running `discover-repos.sh`, check `CLOSEDLOOP_ADD_DIRS` from the environment. This variable contains pipe-separated paths passed via `--add-dir` flags, representing local repositories that are already part of the current task plan. Example: `CLOSEDLOOP_ADD_DIRS="/path/to/a|/path/to/b"`.
+
+For each path in `CLOSEDLOOP_ADD_DIRS`:
+1. Normalize the path (resolve symlinks, trailing slashes)
+2. Find the matching entry in the `peers[]` array from the discovery output by comparing the `path` field
+3. If a match is found, mark that peer with `"local": true` in the entry written to `.cross-repo-needs.json`
```
shafty023: not sure I follow here. If `--add-dirs` is passed that means we are giving claude permission to access those additional repos. So we're taking that as meaning these additional repos are important for either planning or execution. So what does this "local" vs "external" repo logic do? If a match is found why would it be "local" where we don't generate a PRD/plan?

plugins/code/.claude-plugin/plugin.json
aponamarev: Fixed

plugins/code/agents/cross-repo-coordinator.md
aponamarev: Fixed

plugins/code/agents/cross-repo-coordinator.md
aponamarev: When we set `local: true` in `.cross-repo-needs.json` that is signaling to section 2 of cross-repo-prd-writer — this peer is already in scope for the current run, so we should not generate a separate cross-repo PRD for it.

shafty023: @aponamarev but can't we just have `cross-repo-prd-writer` filter on `discoveryMethod === "add_dir"` rather than add `local: true` ?

shafty023:
@aponamarev 

Spent some time tracing how the multi-repo context actually reaches the agents and want to flag a few concerns before this merges. Not bug blockers, but worth addressing while the surface is small.

1. The overlay system looks like over-engineering

This PR introduces a new prompt-overlay abstraction (prompts/overlays/ dir, 65-line README, assembly branch in setup-closedloop.sh, auto-selection on --add-dir) to support exactly one consumer: `multi-repo.overlay.md`. And when you trace what that overlay actually does, all three of its amendments restate things the target agents' own system prompts already cover:
- Amendment to Phase 0 tells the orchestrator to remind pre-explorer about `CLOSEDLOOP_REPO_MAP`, but `pre-explorer.md` already has a ~90-line "Multi-Repo Exploration" section that reads the env var directly, iterates entries, and writes code-map-{name}.json.
- Amendment to Phase 1 tells the orchestrator to remind `plan-draft-writer` about the @{repo-name}:path convention, but `plan-draft-writer.md` already has a full "Multi-Repository Plans" section (M1-M5) with that convention, the ## Repositories table, and the repositories field in plan.json.
- Amendment to Phase 1.4 is framed as a "NOTE" about local peers, but `discover-repos.sh` → `cross-repo-coordinator` Step 1.1 → `cross-repo-prd-writer` Step 2 already handle this deterministically via the `local: true` flag. The orchestrator never branches on it.

Net effect: we built an extension point, used it once, and the one use adds zero behavioral delta while spending ~38 lines of tokens per iteration and creating a drift risk (if the @repo:path convention in `plan-draft-writer.md` changes later, nothing catches a stale overlay).

If there's no second overlay lined up, my vote is to delete the overlay infrastructure entirely (overlays dir, README, the assembly and auto-select branches in `setup-closedloop.sh`) along with the overlay content. Multi-repo still works end-to-end because the agents own the behavior and read the env vars directly. It's trivial to restore from git if a real second consumer shows up later, and much easier to delete now than after it calcifies.

Side note: the PR description lists `prompt-multi-repo.md` (new) under Prompts, but what shipped is the overlay system. Worth updating so future archaeologists aren't looking for a file that doesn't exist.

2. local vs discoveryMethod is a minor SSOT smell

`discover-repos.sh:67` stamps `add-dir` peers with both "discoveryMethod": "add_dir" and "local": true, and no other discovery method produces local peers. `cross-repo-coordinator.md` Step 1.1 then has to manually preserve the local field when writing `.cross-repo-needs.json`. Two fields, one fact, with a hand-written preservation step in between.

The mechanism of tagging at discovery time is the right shape (much better than path-matching inside the agent), but the representation should pick one:
- Drop local and filter downstream on discoveryMethod == "add_dir", or
- Drop discoveryMethod: "add_dir" and keep local.

Either way, Step 1.1 in `cross-repo-coordinator.md` becomes unnecessary.

shafty023 (changes requested):
finished review

CHANGELOG.md
```diff
@@ -14,6 +14,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Updated agent output path references from `.claude/runs/` to `.closedloop-ai/runs/` in `agent-prompt-generator`
 - Updated bootstrap configuration documentation in `agent-bootstrap.md` to reference `.closedloop-ai/` state directory
 
+### code v1.10.0
```
shafty023: plugin.json says `1.9.0` but CHANGELOG says `1.10.0`. Need to fix

aponamarev: Fixed

shafty023 (approved):

PR #47, opened 2026-04-09
FEA-372: Add post-loop code review and fix skill

wongk:
## Summary

- Adds comprehensive code review as the final step in the run-loop pipeline, replacing the lightweight in-loop `code:code-reviewer` with the full `code-review:start` pipeline after implementation completes
- Creates a new `code-review:fix` skill that verifies findings, fixes confirmed issues serially, runs build validation, and supports configurable re-review cycles
- Writes `CLOSEDLOOP_START_SHA` to `config.env` so the review can scope its diff to only this run's changes

**Implementation plan:** [PLN-233](https://app.closedloop.ai/implementation-plans/PLN-233)

### `plugins/code/prompts/prompt.md`
- Remove Phase 5 Step 2 (lightweight single-agent code reviewer)
- Renumber Step 3 → Step 2
- Rename phase from "Testing and Code Review" to "Testing and Validation"

### `plugins/code/scripts/run-loop.sh`
- Add `run_post_loop_review()` function (~130 lines) invoked after both completion-promise and max-iterations exit paths
- Detects changes via working tree + staged + committed diffs against `START_SHA`
- Runs `code-review:start --base $START_SHA`, reads `verdict.json` from the session directory
- If verdict is not `approve`, invokes `code-review:fix` with `--max-cycles 2`
- Write `CLOSEDLOOP_START_SHA` to `config.env` in `create_state_file()`

### `plugins/code-review/skills/fix/SKILL.md` (new)
- 6-step pipeline: parse findings → verify with sonnet agents (parallel) → fix serially → build-validator → conditional re-review → summary
- Only acts on BLOCKING/HIGH severity findings; MEDIUM/LOW are logged
- Auto-discovers most recent `cr-*` session directory when not provided
- Capped at configurable `--max-cycles` (default 2) to prevent infinite loops

### `plugins/code-review/commands/fix.md` (new)
- Thin command wrapper for manual `/code-review:fix` invocation

### Version bumps
- `code`: 1.6.0 → 1.7.0
- `code-review`: 1.4.0 → 1.5.0

## Test plan

- [ ] Run `run-loop.sh` on a project with an intentional bug — verify post-loop review catches it and fix skill addresses it
- [ ] Run on clean code — verify review returns `approve` and no fix cycle triggers
- [ ] Run `/code-review:fix` manually on a prior review session directory
- [ ] Verify `CLOSEDLOOP_START_SHA` is written to `config.env`
- [ ] Verify Ctrl+C during post-loop review is handled gracefully
- [ ] Verify Phase 5 no longer launches `code:code-reviewer`

### DIFF
--- a/plugins/code-review/skills/fix/SKILL.md
+++ b/plugins/code-review/skills/fix/SKILL.md
@@ -1,0 +1,166 @@
+---
+name: fix
+description: Verify and fix code review findings, then run project verification. Use this skill whenever the user wants to address, fix, or remediate code review findings after running /code-review:start, when there are BLOCKING or HIGH severity issues to resolve, or when the user says things like "fix the review comments", "address the findings", or "remediate code review issues". Only applies when a prior code review session directory exists. Accepts a review session directory path and optional --max-cycles flag.
+---
+
+# Fix Code Review Findings
+
+Verify and remediate BLOCKING and HIGH severity findings from a prior `/code-review:start` run, then run project verification (test, lint, typecheck, build). Caps at a configurable number of review-fix cycles to prevent infinite loops.
+
+## Usage
+
+```
+/code-review:fix .closedloop-ai/code-review/cr-abc123
+/code-review:fix .closedloop-ai/code-review/cr-abc123 --max-cycles 1
+```
+
+**Finding the session directory:** The most recent review session is the latest `cr-*` directory under `.closedloop-ai/code-review/`:
+```bash
+ls -td .closedloop-ai/code-review/cr-* | head -1
+```
+
+---
+
+## Step 1: Parse Arguments and Load Findings
+
+Parse `$ARGUMENTS` to extract:
+- **CR_DIR** (required positional): Path to the code-review session directory
+- **--max-cycles N** (optional, default: 2): Maximum review-fix cycles. Initialize `CYCLE = 1`.
+
+Validation:
+- If CR_DIR is missing: auto-discover by running `ls -td .closedloop-ai/code-review/cr-* | head -1`. If no session directories exist, print error "No code review session found. Run /code-review:start first." and exit
+- If `<CR_DIR>/validate_output.json` does not exist: print error "No validate_output.json found in <CR_DIR>. Run /code-review:start first." and exit
+
+### Load Findings
+
+- Read `<CR_DIR>/validate_output.json`
+- Extract the `validated` array
+- Filter to findings where `severity` is `BLOCKING` or `HIGH`
+- Count and log MEDIUM/LOW findings that will be skipped
+- If no BLOCKING/HIGH findings: print "No actionable findings (BLOCKING/HIGH). Review complete." and mark all todos completed and exit
+
+### Create Todo List
+
+```json
+TodoWrite([
+  {"content": "Parse arguments and load findings", "status": "completed", "activeForm": "Parsing arguments"},
+  {"content": "Verify findings", "status": "pending", "activeForm": "Verifying findings"},
+  {"content": "Fix confirmed findings", "status": "pending", "activeForm": "Fixing findings"},
+  {"content": "Run project verification", "status": "pending", "activeForm": "Running verification"},
+  {"content": "Re-review", "status": "pending", "activeForm": "Re-reviewing"},
+  {"content": "Print summary", "status": "pending", "activeForm": "Printing summary"}
+])
+```
+
+---
+
+## Step 2: Verify Each Finding
+
+For each BLOCKING/HIGH finding, launch a verification subagent to determine if it is a real bug or a false positive.
+
+For each finding, launch a Task with `model: sonnet` and tools `Read, Grep, Glob`:
+
+**Prompt template** (inline all values -- no shell variables):
+```
+Determine if this code review finding is a real bug or a false positive.
+
+**Finding:**
+- File: <file>
+- Line: <line>
+- Severity: <severity>
+- Category: <category>
+- Issue: <issue>
+- Explanation: <explanation>
+- Code snippet: <code_snippet>
+- Recommendation: <recommendation>
+
+Read the cited file and 50 lines of surrounding context. Check for error handling, type guards, intentional patterns, or inline justifications that would make this a false positive. Verify the issue actually exists at the cited line.
+
+Output exactly one of:
+CONFIRMED: <brief reasoning why this is a real bug>
+REJECTED: <brief reasoning why this is a false positive>
+```
+
+Launch all verification agents **in parallel** (multiple Task calls in a single message).
+
+### Collect Results
+
+- Parse each agent's response for `CONFIRMED` or `REJECTED`
+- Filter to only CONFIRMED findings
+- If no confirmed findings remain: print "All findings were false positives. No fixes needed." and mark remaining todos as completed and exit
+
+---
+
+## Step 3: Fix Confirmed Findings
+
+Group related confirmed findings by file. Findings in the same file should be fixed together.
+
+Fix each finding **one at a time, sequentially** -- do NOT launch fix agents in parallel. Separate findings may affect the same source files, so each fix must see the results of prior fixes to avoid conflicts.
+
+For each finding or group, launch a Task with `model: sonnet` and tools `Read, Write, Edit, Grep, Glob`:
+
+**Prompt template** (inline all values -- no shell variables):
+```
+Fix this code review finding. Apply the minimal change needed -- do NOT refactor surrounding code, add new features, or add unnecessary error handling.
+
+**Finding:**
+- File: <file>
+- Line: <line>
+- Issue: <issue>
+- Explanation: <explanation>
+- Recommendation: <recommendation>
+
+Read the file, apply the fix, and confirm what you changed.
+```
+
+Wait for each fix agent to complete before launching the next one. Track the list of modified files for re-review in Step 5.
+
+---
+
+## Step 4: Run Project Verification
+
+Launch a Task with `subagent_type: "code:build-validator"`:
+
+```
+Run all validation commands (test, lint, typecheck, build) for this project. Report VALIDATION_PASSED, VALIDATION_FAILED, or NO_VALIDATION.
+```
+
+- **VALIDATION_PASSED** or **NO_VALIDATION**: Proceed to Step 5
+- **VALIDATION_FAILED**: Delegate fix to a Sonnet subagent and retry the validator, up to 5 attempts total. Log a warning and proceed on persistent failure.
+
+---
+
+## Step 5: Re-review (Conditional)
+
+- If `CYCLE >= MAX_CYCLES`: log any remaining concerns as warnings, proceed to Step 6
+- If `CYCLE < MAX_CYCLES` AND fixes were applied in Step 3:
+  - Increment `CYCLE`
+  - Invoke the code review on only the modified files using the Skill tool: `Skill(skill="code-review:start", args="<file1> <file2> ...")`
+  - Find the new session directory (`ls -td .closedloop-ai/code-review/cr-* | head -1`) and read its `validate_output.json`
+  - If new BLOCKING/HIGH findings exist: loop back to Step 2 with the new findings
+  - If clean: proceed to Step 6
+
+---
+
+## Step 6: Summary
+
+Print a structured summary of the fix session:
+
+```markdown
+## Fix Summary
+
+| Metric | Value |
+|--------|-------|
+| Findings received | N |
+| Confirmed (not false positive) | N |
+| Fixed | N |
+| Remaining (warnings) | N |
+| Verification | PASSED/FAILED/NO_VALIDATION |
+| Cycles used | N of MAX |
+```
+
+Mark all todos as `completed`.
+
+---
+
+$ARGUMENTS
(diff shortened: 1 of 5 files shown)


plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
+---
+name: fix
+description: Verify and fix code review findings, then run project verification. Use this skill whenever the user wants to address, fix, or remediate code review findings after running /code-review:start, when there are BLOCKING or HIGH severity issues to resolve, or when the user says things like "fix the review comments", "address the findings", or "remediate code review issues". Only applies when a prior code review session directory exists. Accepts a review session directory path and optional --max-cycles flag.
```
shafty023: add [argument-hint](https://code.claude.com/docs/en/skills#frontmatter-reference) here to let users know what arguments this supports. Don't mention it in the `description` field.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+---
+
+# Fix Code Review Findings
```
shafty023, line 7: Nit: activate `/context-engineering` skill and then ask claude to optimize this skill

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+
+Verify and remediate BLOCKING and HIGH severity findings from a prior `/code-review:start` run, then run project verification (test, lint, typecheck, build). Caps at a configurable number of review-fix cycles to prevent infinite loops.
```
shafty023: I'd be careful about mentioning another slash command here. We don't want the agent activating this skill to either try to run this slash command (not allowed, will fail) or go and read that slash command to find out what it contains (pollutes context).

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Load Findings
+
+- Read `<CR_DIR>/validate_output.json`
+- Extract the `validated` array
+- Filter to findings where `severity` is `BLOCKING` or `HIGH`
+- Count and log MEDIUM/LOW findings that will be skipped
+- If no BLOCKING/HIGH findings: print "No actionable findings (BLOCKING/HIGH). Review complete." and mark all todos completed and exit
+
+### Create Todo List
```
shafty023: This `TodoWrite` should be mentioned before the "load findings" section since it references it in the `TodoWrite`.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+  {"content": "Re-review", "status": "pending", "activeForm": "Re-reviewing"},
+  {"content": "Print summary", "status": "pending", "activeForm": "Printing summary"}
+])
+```
+
+---
+
+## Step 2: Verify Each Finding
+
+For each BLOCKING/HIGH finding, launch a verification subagent to determine if it is a real bug or a false positive.
```
shafty023: If you don't mention which subagent to use it might pick whatever it thinks it should use in that moment. That means whatever someone has available in their CC session that appears to be a good match it will use which is not what you actually want. In this case a [general-purpose](https://code.claude.com/docs/en/sub-agents#general-purpose) subagent probably will do. Have claude update this prompt so it writes it correctly to use that subagent. The best way is for it to actually write the `Task` tool command so there's no discretion as to what to call here. 

Also, since you decided to make this a skill instead of a slash command, that means the agent activating this skill can't actually launch a subagent. Subagents can't launch subagents. You'll want to verify this even works as written. You might have to change this to a slash command for this step to even work.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+
+For each finding, launch a Task with `model: sonnet` and tools `Read, Grep, Glob`:
+
+**Prompt template** (inline all values -- no shell variables):
+```
+Determine if this code review finding is a real bug or a false positive.
```
shafty023: Have you looked at `plugins/code-review/tools/prompts/shared_prompt.txt` yet? That was created from that article you'd shared that Meta wrote -> https://venturebeat.com/orchestration/metas-new-structured-prompting-technique-makes-llms-significantly-better-at . That's more geared towards the actual review but maybe you can employ some of those tactics here for verification.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+- Recommendation: <recommendation>
+
+Read the cited file and 50 lines of surrounding context. Check for error handling, type guards, intentional patterns, or inline justifications that would make this a false positive. Verify the issue actually exists at the cited line.
+
+Output exactly one of:
+CONFIRMED: <brief reasoning why this is a real bug>
+REJECTED: <brief reasoning why this is a false positive>
+```
+
+Launch all verification agents **in parallel** (multiple Task calls in a single message).
```
shafty023: Another mention here as on line 59 that generically mentions launching a verification agent. We only need it mentioned once.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+
+For each BLOCKING/HIGH finding, launch a verification subagent to determine if it is a real bug or a false positive.
+
+For each finding, launch a Task with `model: sonnet` and tools `Read, Grep, Glob`:
+
+**Prompt template** (inline all values -- no shell variables):
+```
+Determine if this code review finding is a real bug or a false positive.
+
+**Finding:**
```
shafty023: Coming up with a json schema for the results would ensure each subagent will always return the results in the same format every time. Tweaks to CC binary won't mess it up.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+Output exactly one of:
+CONFIRMED: <brief reasoning why this is a real bug>
+REJECTED: <brief reasoning why this is a false positive>
+```
+
+Launch all verification agents **in parallel** (multiple Task calls in a single message).
+
+### Collect Results
+
+- Parse each agent's response for `CONFIRMED` or `REJECTED`
```
shafty023: Explain exactly where in the findings schema the orchestrator should look for this.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+Wait for each fix agent to complete before launching the next one. Track the list of modified files for re-review in Step 5.
+
+---
+
+## Step 4: Run Project Verification
+
+Launch a Task with `subagent_type: "code:build-validator"`:
+
+```
+Run all validation commands (test, lint, typecheck, build) for this project. Report VALIDATION_PASSED, VALIDATION_FAILED, or NO_VALIDATION.
```
shafty023: It's not clear that this here is intended to be passed to the `build-validator` subagent. Let's make that explicit so the orchestrator doesn't launch that agent then also do these steps itself

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+- **VALIDATION_PASSED** or **NO_VALIDATION**: Proceed to Step 5
+- **VALIDATION_FAILED**: Delegate fix to a Sonnet subagent and retry the validator, up to 5 attempts total. Log a warning and proceed on persistent failure.
```
shafty023: Be explicit which subagent should be used. Likely `general-purpose` subagent should be fine.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+| Remaining (warnings) | N |
+| Verification | PASSED/FAILED/NO_VALIDATION |
+| Cycles used | N of MAX |
+```
+
+Mark all todos as `completed`.
+
+---
+
+$ARGUMENTS
```
shafty023:
```
## Arguments

$ARGUMENTS
```

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,166 @@
...
+- **VALIDATION_FAILED**: Delegate fix to a Sonnet subagent and retry the validator, up to 5 attempts total. Log a warning and proceed on persistent failure.
+
+---
+
+## Step 5: Re-review (Conditional)
+
+- If `CYCLE >= MAX_CYCLES`: log any remaining concerns as warnings, proceed to Step 6
+- If `CYCLE < MAX_CYCLES` AND fixes were applied in Step 3:
+  - Increment `CYCLE`
+  - Invoke the code review on only the modified files using the Skill tool: `Skill(skill="code-review:start", args="<file1> <file2> ...")`
```
shafty023: This can't work. `/code-review:start` is a slash command. You can't launch a slash command from inside a skill. This is why I came up with `run-loop.sh` in the first place. So after planning/execution finished I could optionally run other things afterwards since I can't do that from within a running slash command.

plugins/code/scripts/run-loop.sh
```diff
@@ -436,6 +436,151 @@ run_background_pruning() {
...
+  ) &
+  wait $! 2>/dev/null || true
+  local review_exit
+  review_exit=$(cat "$review_exit_file" 2>/dev/null || echo 1)
+  rm -f "$review_exit_file"
+  set -e
+
+  # Find the most recent CR_DIR (code-review session directory)
+  local cr_dir
+  cr_dir=$(ls -td .closedloop-ai/code-review/cr-* 2>/dev/null | head -1)
```
shafty023: **[P1]** Stale cr-* directory verdict consumed when code review fails

In `run_post_loop_review()`, after running the code review subprocess, the function discovers the session directory using `ls -td .closedloop-ai/code-review/cr-* | head -1` with no verification it was created by the current invocation. The captured `review_exit` is only used for telemetry, never as a gate. If `claude` exits non-zero without creating a new cr-* directory, and a prior cr-* directory exists with a non-approve verdict.json, the function reads the stale verdict and triggers the fix cycle against old findings.



> **Suggestion:** Guard on review_exit: add `if [[ "$review_exit" -ne 0 ]]; then log_progress "Post-loop review failed (exit $review_exit). Skipping fix."; return 0; fi` immediately after capturing the exit code. Alternatively, snapshot existing cr-* directories before the review and verify a new one appeared.

shafty023 (changes requested):
finished review

plugins/code-review/skills/fix/SKILL.md
wongk: i already used the Anthropic skill-creator skill to review this skill twice.

plugins/code-review/skills/fix/SKILL.md
wongk: i did verify it is working

plugins/code-review/skills/fix/SKILL.md
wongk: yeah, i noticed this myself. but i verified it yesterday and it worked... somehow claude figured it out. i'm not sure how. but i'll get this fixed up anyway.

shafty023: We don't want to do this. Here's why:

orchestrator activates skill `/code-review:fix`
- 166 lines of prompt are injected into the orchestrator
- any file reads or subagent responses are read by the orchestrator
- task list is created

Near the end you ask the orchestrator to now launch `/code-review:start`
- 1,250 lines of prompt are injected into the orchestrator
- 13 tasks including multi-agent spawning, disk i/o, caching, validation, etc
- All of the context from the fix workflow is still in context with its own set of tasks it was managing

This is a recipe for degraded quality. We don't want to chain them that way.

shafty023: Each orchestrator in a workflow should have one job and its job is to follow the workflow exactly as written and never deviate. Doing that deterministically is the hard part. Especially as workflows get more complex. You have to run these things over and over, see where the orchestrator is doing something it shouldn't be doing (going rogue) and then update the prompt's accordingly. Since it's that difficult just on one workflow to get right, the last thing you'd want to do is chain workflows in a single context window.

wongk: the loop is moved to `run-loop.sh`

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,150 @@
...
+name: fix
+description: Verify and fix BLOCKING/HIGH code review findings from a prior review session, then run project verification.
+argument-hint: "<cr-dir>"
+---
+
+# Fix Code Review Findings
+
+Single verify-and-fix pass for BLOCKING/HIGH findings from a prior code review. The caller controls re-review cycles.
+
+## Arguments
```
shafty023, line 11: Did the context-engineering skill move this up here or did you? Generally it's best for the orchestrator to see this last as you had it before but if that skill moved it up here I guess

wongk: i don't know if it was claude itself, or the skill. i had claude pull in all your feedback, create a plan to address it all, and then i executed it.

plugins/code-review/skills/fix/SKILL.md
```diff
@@ -0,0 +1,150 @@
+---
+name: fix
+description: Verify and fix BLOCKING/HIGH code review findings from a prior review session, then run project verification.
+argument-hint: "<cr-dir>"
+---
+
+# Fix Code Review Findings
```
shafty023, line 7: Changes here look really good

shafty023 (approved):

PR #48, opened 2026-04-13
refactor(code): compress orchestrator prompt from ~10.4K to ~5.1K tokens

mikeangstadt:
## Summary

Compresses `plugins/code/prompts/prompt.md` by 51% (from ~10,444 to ~5,143 tokens) with no meaning lost. Two commits:

**Pass 1 — mechanical extraction:**
- Extract 2 inline bash blocks into standalone scripts (`stamp_cross_repo_cache.sh`, `stamp_critic_cache.sh`)
- Define `PLAN_VALIDATION_SEQUENCE` procedure once, referenced 3x (Phase 1, 2.6, 2.7)
- Define `AWAITING_USER_SEQUENCE` procedure once, referenced 2x (Phase 1.1, Phase 5)

**Pass 2 — editorial compression:**
- Collapse 6 WRONG/CORRECT examples into 2 compact lines
- Convert skills section from prose to table
- Compress state tracking from 10-row table to base schema + extensions
- Replace 17-item TodoWrite JSON literal with compact phase list
- Remove per-phase "Update state.json" lines (stated once as a rule)
- Remove scratchpad section
- Compress subagent prompt strings throughout all phases

## Test plan

- [ ] Verify `stamp_cross_repo_cache.sh` produces same hash output as the original inline block
- [ ] Verify `stamp_critic_cache.sh` produces same hash output as the original inline block
- [ ] Run a full `/code:code` loop to confirm no behavioral regression
- [ ] Verify all procedure references (PLAN_VALIDATION_SEQUENCE, AWAITING_USER_SEQUENCE) resolve correctly in orchestrator flow

### DIFF
--- a/plugins/code/prompts/prompt.md
+++ b/plugins/code/prompts/prompt.md
@@ -50,7 +50,27 @@ Use this sequence at any hard-stop that requires user action before continuing:
 
 ## Required TodoWrite
 
-**MANDATORY first action:** Create a TodoWrite entry for each phase below (0.9, 1, 1.1, 1.2, 1.2a, 1.3, 1.4, 1.4.1, 1.4.2, 1.4.3, 2.5, 2.6, 2.7, 3, 4, 5, 6, 7), all `pending`. Mark `in_progress` when starting, `completed` when done.
+**MANDATORY first action:** Create a TodoWrite entry for each phase, all `pending`. Use the content and activeForm below. Mark `in_progress` when starting, `completed` when done.
+
+| content | activeForm |
+|---|---|
+| Phase 1: Planning | Planning |
+| Phase 1.1: Plan review checkpoint | Awaiting plan review decision |
+| Phase 1.2: Process answered questions | Processing answered questions |
+| Phase 1.2a: Process addressed gaps | Processing addressed gaps |
+| Phase 1.3: Simple mode evaluation | Evaluating plan complexity |
+| Phase 1.4: Cross-repo coordination | Coordinating cross-repo |
+| Phase 1.4.1: Discover peers | Discovering peers |
+| Phase 1.4.2: Verify capabilities | Verifying capabilities |
+| Phase 1.4.3: Generate PRDs | Generating cross-repo PRDs |
+| Phase 2.5: Critic validation | Running critic reviews |
+| Phase 2.6: Plan refinement | Merging critic feedback |
+| Phase 2.7: Plan finalization | Finalizing plan |
+| Phase 3: Implementation | Implementing |
+| Phase 4: Code simplification | Simplifying code |
+| Phase 5: Testing and Validation | Testing |
+| Phase 6: Visual inspection | Inspecting visuals |
+| Phase 7: Logging and completion | Completing |
 
 ## State Tracking
 
--- a/plugins/code/prompts/prompt.md
+++ b/plugins/code/prompts/prompt.md
@@ -54,6 +54,7 @@ Use this sequence at any hard-stop that requires user action before continuing:
 
 | content | activeForm |
 |---|---|
+| Phase 0.9: Pre-exploration | Pre-exploring |
 | Phase 1: Planning | Planning |
 | Phase 1.1: Plan review checkpoint | Awaiting plan review decision |
 | Phase 1.2: Process answered questions | Processing answered questions |
@@ -140,8 +141,6 @@ Here are the key phases you must complete:
 
 **PHASE 1.4: CROSS-REPO COORDINATION**
 
-- If `simple_mode` is true, this phase was already marked complete. Skip to Phase 3.
-
 **Phase 1.4.1: Discover peers**
 - Activate `code:cross-repo-cache` skill. On `CACHE_HIT` with `NO_CROSS_REPO_NEEDED`: mark 1.4.x complete, skip to Phase 2.5. On `CAPABILITIES_IDENTIFIED`: skip to 1.4.2.
 - On `CACHE_MISS`: Launch @code:cross-repo-coordinator with `WORKDIR` and `PLAN_PATH=$CLOSEDLOOP_WORKDIR/plan.json`
@@ -160,9 +159,8 @@ Here are the key phases you must complete:
 - Generates PRDs for missing capabilities, updates plan.json with cross-repo tags
 - Proceed to Phase 2.5
 
-**PHASE 2.5: CRITIC VALIDATION** (skipped if simple_mode = true)
+**PHASE 2.5: CRITIC VALIDATION**
 
-- Skip if `simple_mode = true`
 - Activate `code:critic-cache` skill. On `CACHE_HIT`: skip to Phase 2.6. On `CACHE_MISS`: continue.
 - `mkdir -p $CLOSEDLOOP_WORKDIR/reviews`
 - Launch Task() **in parallel** for each critic: "WORKDIR=$CLOSEDLOOP_WORKDIR. Review plan as {critic_name} specialist. Read plan.md, investigation-log.md, PRD. Write to reviews/{critic_name}.review.json with findings: {severity, description, recommendation, affectedTasks}."
@@ -174,9 +172,8 @@ Here are the key phases you must complete:
 - After plan-writer completes, run **PLAN_VALIDATION_SEQUENCE**
 - Proceed to Phase 2.7
 
-**PHASE 2.7: PLAN FINALIZATION** (skipped if simple_mode = true)
+**PHASE 2.7: PLAN FINALIZATION**
 
-- If `simple_mode` is true, skip to Phase 3
 - Launch @code:plan-writer with `WORKDIR`, FINALIZE MODE: enrich task descriptions with implementation details (code patterns, signatures, edge cases). Do NOT add/remove/renumber tasks.
 - After plan-writer completes (outputs `<promise>PLAN_WRITER_COMPLETE</promise>`), run **PLAN_VALIDATION_SEQUENCE**
 - Proceed to Phase 3
@@ -207,12 +204,9 @@ Here are the key phases you must complete:
 
 **PHASE 5: TESTING AND VALIDATION**
 
-
 **Step 1: Write tests for implemented code**
-- If code was implemented in Phase 3, launch @test-engineer with prompt:
-  "WORKDIR=$CLOSEDLOOP_WORKDIR. Write tests for the code changes made in this session. Focus on the implemented tasks from the plan."
-- The test-engineer will identify testable code and write appropriate unit/integration tests
-- Skip this step only if: (a) no code was implemented, or (b) the project has no test framework
+- If code was implemented in Phase 3, launch @test-engineer with `WORKDIR=$CLOSEDLOOP_WORKDIR` to write tests for the changes
+- Skip if no code was implemented or the project has no test framework
 
 **Step 2: Run validation via build-validator agent:**
 1. Launch @code:build-validator with `WORKDIR=$CLOSEDLOOP_WORKDIR`
@@ -243,11 +237,8 @@ Here are the key phases you must complete:
    - If `BUILD_CACHE_MISS`: Launch @code:build-validator with `WORKDIR=$CLOSEDLOOP_WORKDIR`
    - If `VALIDATION_FAILED`:
      1. Log "Final build validation failed. Loop will continue."
-     2. Update state.json:
-        ```bash
-        echo '{"phase": "Phase 7: Logging and completion", "status": "IN_PROGRESS", "reason": "Final build validation failed", "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json
-        ```
-     3. **Do NOT output `<promise>COMPLETE</promise>`** - end naturally, loop will restart
+     2. Update state.json with `"reason": "Final build validation failed"` (base schema + reason)
+     3. **Do NOT output `<promise>COMPLETE</promise>`** — end naturally, loop will restart
    - If `VALIDATION_PASSED` or `NO_VALIDATION`: Continue to step 2
 
 2. **Task and question check:** Activate `code:plan-validate` skill (runs Python script against $CLOSEDLOOP_WORKDIR) — semantic check is unnecessary since plan content hasn't changed since last semantic validation
@@ -257,12 +248,8 @@ Here are the key phases you must complete:
 
 - **If `pending_tasks` is NOT empty (work remains):**
   1. Log: "Pending tasks remain: [task IDs]. Loop will continue."
-  2. Update state.json:
-     ```bash
-     echo '{"phase": "Phase 7: Logging and completion", "status": "IN_PROGRESS", "reason": "Pending tasks remain", "pendingTasks": ["T-X.Y", ...], "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json
-     ```
-  3. **Do NOT output `<promise>COMPLETE</promise>`** - just end your response naturally
-  4. The external loop will automatically restart a fresh iteration
+  2. Update state.json with `"reason": "Pending tasks remain"` and `"pendingTasks": [...]` (base schema + fields)
+  3. **Do NOT output `<promise>COMPLETE</promise>`** — end naturally, loop will restart
 
 - **If all clear:** Write state.json with `"status": "COMPLETED"`, THEN output `<promise>COMPLETE</promise>`. Never output the promise without writing state.json first.
 
--- a/plugins/code/prompts/prompt.md
+++ b/plugins/code/prompts/prompt.md
@@ -260,3 +260,5 @@ Here are the key phases you must complete:
 4. Do not over-engineer. Only ask questions for critical missing information.
 5. Document all changes in $CLOSEDLOOP_WORKDIR/log.md.
 6. Output `<promise>COMPLETE</promise>` ONLY when ALL phases done and `pending_tasks` is empty. If tasks remain, end naturally — the loop will restart.
+7. **Self-check before ANY tool use:** "Am I about to read or edit a file? If yes, delegate to a subagent instead."
+8. **Self-check before ANY `<promise>` output:** "Did I write state.json with the correct status?" If you output the promise WITHOUT writing state.json first, external systems will show "IN_PROGRESS" forever.
--- a/plugins/code/prompts/prompt.md
+++ b/plugins/code/prompts/prompt.md
@@ -109,6 +109,25 @@ This orchestrator also has access to the **iterative-retrieval** skill for refin
 
 See the skill documentation for the 4-phase protocol (Initial Dispatch → Sufficiency Evaluation → Refinement Request → Loop).
 
+## Reusable Procedures
+
+### PLAN_VALIDATION_SEQUENCE
+
+Use this sequence whenever a phase needs full plan validation (structural + semantic):
+1. Activate `code:plan-validate` skill (runs Python script against $CLOSEDLOOP_WORKDIR)
+2. If `FORMAT_ISSUES`: launch @code:plan-writer to fix format issues, then re-activate `code:plan-validate`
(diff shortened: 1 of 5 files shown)


plugins/code/scripts/stamp_cross_repo_cache.sh
```diff
@@ -0,0 +1,25 @@
...
+# Computes a hash of peer repo HEADs (from .workspace-repos.json) or
+# falls back to hashing .cross-repo-needs.json. Writes to <WORKDIR>/.cross-repo-hash.
+
+set -euo pipefail
+
+WORKDIR="${1:?Usage: stamp_cross_repo_cache.sh <WORKDIR>}"
+
+if [ -f ".workspace-repos.json" ]; then
+  python3 -c "import json; [print(r['path']) for r in json.load(open('.workspace-repos.json')) if r.get('path')]" 2>/dev/null \
+    | while IFS= read -r repo_path; do
```
mikeangstadt: **[HIGH]** Correctness

[P1] Cross-repo cache hash format differs from check script, causing permanent cache miss

`stamp_cross_repo_cache.sh` hashes repo entries as newline-separated sorted lines via printf/sort/shasum pipeline. However, `check_cross_repo_cache.sh` accumulates entries as space-separated values and hashes that single-line string. The two scripts produce different SHA-256 hashes for the same repositories, making the cache non-functional.

**Recommendation:** Match the hash computation format used by `check_cross_repo_cache.sh`, or update both scripts to use the same format.

plugins/code/prompts/prompt.md
```diff
@@ -232,66 +126,27 @@ Here are the key phases you must complete:
...
-    shasum -a 256 "$CLOSEDLOOP_WORKDIR/.cross-repo-needs.json" > "$CLOSEDLOOP_WORKDIR/.cross-repo-hash"
-  fi
-  ```
-- Handle return status:
-  - `NO_CROSS_REPO_NEEDED`: Mark 1.4.x phases complete, proceed to Phase 2.5
-  - `CROSS_REPO_SKIPPED`: Mark 1.4.x phases complete, proceed to Phase 2.5
-  - `CAPABILITIES_IDENTIFIED`: Continue to Phase 1.4.2
+- Activate `code:cross-repo-cache` skill. On `CACHE_HIT` with `NO_CROSS_REPO_NEEDED`: mark 1.4.x complete, skip to Phase 2.5. On `CAPABILITIES_IDENTIFIED`: skip to 1.4.2.
+- On `CACHE_MISS`: Launch @code:cross-repo-coordinator with `WORKDIR` and `PLAN_PATH=$CLOSEDLOOP_WORKDIR/plan.json`
+- Stamp cache: `bash scripts/stamp_cross_repo_cache.sh $CLOSEDLOOP_WORKDIR`
```
mikeangstadt: **[MEDIUM]** Correctness

[P2] Scripts invoked with bare relative path instead of $CLAUDE_PLUGIN_ROOT/scripts/

The orchestrator runs with CWD = user's project directory. These scripts live at `$CLAUDE_PLUGIN_ROOT/scripts/`, so bare `scripts/...` won't resolve. The established convention in this codebase is `${CLAUDE_PLUGIN_ROOT}/scripts/...` (see cross-repo-coordinator.md:52, repo-coordinator.md:20).

**Recommendation:** Change to `bash "$CLAUDE_PLUGIN_ROOT/scripts/stamp_cross_repo_cache.sh" "$CLOSEDLOOP_WORKDIR"` and same for `stamp_critic_cache.sh` at line 169.

plugins/code/scripts/stamp_critic_cache.sh
```diff
@@ -0,0 +1,18 @@
...
+# Usage: stamp_critic_cache.sh <WORKDIR>
+#
+# Hashes plan.json (+ critic-gates.json if present) and writes to
+# <WORKDIR>/reviews/.plan-hash so subsequent runs can skip critics.
+
+set -euo pipefail
+
+WORKDIR="${1:?Usage: stamp_critic_cache.sh <WORKDIR>}"
+
+if [ -f ".closedloop-ai/settings/critic-gates.json" ]; then
```
mikeangstadt: **[MEDIUM]** Correctness

[P2] Stamp misses second critic-gates.json probe path used by check script

`check_critic_cache.sh` probes two locations for critic-gates.json: `.closedloop-ai/settings/critic-gates.json` (CWD-relative) and `$(dirname "$WORKDIR")/settings/critic-gates.json` as fallback. This stamp script only checks the first path. When critic-gates.json lives at the fallback path only, the stamp hashes only plan.json but the check hashes plan.json + critic-gates.json, causing permanent cache miss.

**Recommendation:** Add the same fallback probe to this script.

plugins/code/prompts/prompt.md
```diff
@@ -406,53 +217,23 @@ Here are the key phases you must complete:
...
-           ```
-        2. **ONLY AFTER state.json is written** - Output `<promise>COMPLETE</promise>`
-        3. Tell the user: "Validation failed after 20 attempts. Fix issues manually and run `/code:code $ARGUMENTS` to continue."
-        4. **HARD STOP** - Do not continue.
-        </awaiting_user_sequence>
+     a. Delegate fixes to subagents (test failures → @test-engineer, other → sonnet subagent)
+     b. Re-run @code:build-validator. Repeat until VALIDATION_PASSED (max 20 attempts)
+     e. If still failing after 20 attempts: Execute **AWAITING_USER_SEQUENCE** with: phase="Phase 5: Testing and Validation", reason="Validation failed after 20 attempts", file=null, command="/code:code $ARGUMENTS". Tell the user: "Validation failed after 20 attempts. Fix issues manually and run `/code:code $ARGUMENTS` to continue."
```
mikeangstadt: **[MEDIUM]** Code Quality

[P3] VALIDATION_FAILED sub-step labels jump from b to e, implying missing steps

The ordered sub-steps under VALIDATION_FAILED are labeled a, b, e. The old version had five steps a through e; the refactoring condensed them but kept the original label 'e'. The gap in alphabetical labels could cause the LLM orchestrator to hallucinate missing intermediate steps.

**Recommendation:** Relabel step 'e' to 'c' so the sub-steps read a, b, c.

plugins/code/scripts/stamp_cross_repo_cache.sh
mikeangstadt: Fixed in 9e264bd. Rewrote the hash computation to match `check_cross_repo_cache.sh` exactly: space-separated `path:hash ` entries (no sort), hashed via `echo | shasum`, stored as hash-only (`cut -d' ' -f1`). Also added error handling for `.workspace-repos.json` parse failures.

plugins/code/prompts/prompt.md
mikeangstadt: Fixed in 9e264bd. Both invocations now use `$CLAUDE_PLUGIN_ROOT/scripts/` prefix, matching the convention in cross-repo-coordinator.md, repo-coordinator.md, and plan-importer.md.

plugins/code/scripts/stamp_critic_cache.sh
mikeangstadt: Fixed in 9e264bd. Added the same two-path probe as `check_critic_cache.sh`: tries `.closedloop-ai/settings/critic-gates.json` first, then `$(dirname "$WORKDIR")/settings/critic-gates.json` as fallback. Also added `cut -d' ' -f1` to store hash-only (matching the check script's comparison).

plugins/code/prompts/prompt.md
mikeangstadt: Fixed in 9e264bd. Relabeled `e.` to `c.` so sub-steps read a, b, c.

plugins/code/scripts/stamp_cross_repo_cache.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ -f ".workspace-repos.json" ]; then
+  repo_paths=$(python3 -c "import json; [print(r['path']) for r in json.load(open('.workspace-repos.json')) if r.get('path')]" 2>/dev/null) || {
+    echo "CROSS_REPO_CACHE_STAMP_FAILED"
+    echo "reason: failed to parse .workspace-repos.json"
+    exit 1
+  }
+
+  repo_hashes=""
+  while IFS= read -r repo_path; do
```
thadeusb, line 23: **File: stamp_cross_repo_cache.sh, line 33**

The old inline version used `LC_ALL=C sort` before hashing the repo paths, but this script dropped the sort. Without it, the hash is order-dependent on however python3 iterates the JSON array. Two runs with the same repos could produce different hashes if the array order changes, causing false cache misses.

Suggest adding `| LC_ALL=C sort` before the shasum pipe on line 33.

plugins/code/prompts/prompt.md
```diff
@@ -476,57 +248,15 @@ Here are the key phases you must complete:
...
-
-1. Follow the phases sequentially - do not skip ahead
-2. After initial plan creation (Phase 1), you MUST wait for human approval at the Phase 1.1 checkpoint before continuing. Subsequent automated plan edits (gap incorporation, critic merge, finalization, checkbox updates) proceed without additional approval and must follow the phase-specific scope constraints.
-3. All validation checks must pass before completion (or user must explicitly skip)
-4. Use build-validator to discover and run project-specific validation commands - do not hardcode commands
-5. Do not over-engineer solutions
-6. Only ask questions when there are drastically different options or critical missing information
-7. Document all changes in $CLOSEDLOOP_WORKDIR/log.md
-
-Before taking action in each phase, use your scratchpad to think through what needs to be done:
```
thadeusb, line 508: The compression removed the scratchpad reasoning block and self-check reminders ("Am I about to read a file? Delegate instead," "Did I write state.json before the promise tag?"). These acted as guardrails against the two most common orchestrator failure modes. The token savings are real (~5K tokens), but worth monitoring whether orchestrator reliability regresses without those explicit self-check prompts. Consider keeping at least the state.json self-check as a one-liner at the bottom of the RULES section.

plugins/code/prompts/prompt.md
```diff
@@ -3,206 +3,111 @@
 
 **FIRST ACTION RULE:** After reading this prompt, your very first action must be TodoWrite to create the phase list. Do NOT read project files (PRD, plan.json, code, etc.). Start with TodoWrite, then `ls` to check if plan exists.
```
thadeusb, line 4: I found a few more functional losses in the compression:

- the self-check "Am I about to read a file? Delegate instead" was a one-liner that prevented the most common orchestrator mistake. gone now.

- the state.json WARNING that said "if you output the promise WITHOUT writing state.json first, external systems will show IN_PROGRESS forever" was removed. the new version says it is a bug but drops the specific consequence. models respond better to concrete consequences than abstract rules.

- the output format guidance at the bottom (what the orchestrator response should look like, status updates per phase, separating reasoning from output) was removed entirely. the model now has no instruction on how to structure its responses.

- the <critical_requirement> XML tags around the state.json mandate gave that section extra weight. compressed version uses plain bold text which carries less emphasis.

the scratchpad + self-checks are the biggest risk. the others are smaller but they all compound. I would keep a "self-checks" one-liner section at the bottom of the RULES list.

thadeusb (approved):

plugins/code/scripts/stamp_cross_repo_cache.sh
mikeangstadt: Fixed in 1ab5ad2. Added `LC_ALL=C sort` to both `stamp_cross_repo_cache.sh` and `check_cross_repo_cache.sh` so hashes are order-independent and both scripts agree. (The check script also lacked the sort, so adding it to only the stamp would have broken cache matching — fixed both sides.)

plugins/code/prompts/prompt.md
mikeangstadt: Good call — restored both self-checks in 1ab5ad2 as rules 7-8 in the RULES section. The state.json check now includes the concrete consequence ("external systems will show IN_PROGRESS forever"). Two one-liners, minimal token cost.

plugins/code/prompts/prompt.md
mikeangstadt: Both restored in 1ab5ad2. Rule 7 adds back the file-read self-check ("Am I about to read or edit a file? Delegate instead.") and rule 8 adds the state.json warning with the specific consequence ("external systems will show IN_PROGRESS forever"). These are two lines at the bottom of RULES — negligible token cost for meaningful guardrails.

shafty023:
Three issues to address in a follow-up. Issue #3 has been updated — my original recommendation for it was wrong and the underlying problem is larger than I initially thought.

### 1. AWAITING_USER_SEQUENCE template produces invalid JSON when `file=null`

**File:** `plugins/code/prompts/prompt.md:46`

The template has `"file": "<path or null>"` with the placeholder inside quotes. At the Phase 5 call site (line 219), `file=null` is passed, which requires the LLM to drop the quotes when substituting — a subtle template quirk it may get wrong, producing `"file": "null"` (string literal) instead of `"file": null` (JSON null). The pre-PR Phase 5 echo wrote unquoted `"file": null` directly, so this is a regression for external systems that poll state.json and distinguish "no file" from a file literally named `"null"`.

**Fix:** Either split AWAITING_USER_SEQUENCE into two variants (with-file for Phase 1.1, no-file for Phase 5), or change the template to place the placeholder outside the quotes and document explicitly at each call site what to substitute:

```
"file": <"path-string" or null>
```

### 2. AWAITING_USER_SEQUENCE call sites never supply the `description` field

**Files:** `plugins/code/prompts/prompt.md:119` (Phase 1.1), `plugins/code/prompts/prompt.md:219` (Phase 5)

The template at line 46 has `"description": "<what user should do>"` but both call sites pass only `phase`, `reason`, `file`, `command`. The LLM must invent the description — it may pass the placeholder string through verbatim, copy the user-facing sentence, or omit the field entirely, producing different state.json shapes each run.

**Fix:** Add the missing parameter at both call sites so the procedure has everything it needs:

- Phase 1.1 (line 119): `description="Review the plan and run the command when ready"`
- Phase 5 (line 219): `description="Fix validation issues manually and run the command to continue"`

These match what the pre-PR literal echo commands wrote to state.json.

### 3. All three cache stamps in `prompt.md` are non-functional (including the 9e264bd fix)

**Files:** `plugins/code/prompts/prompt.md:147` (cross-repo), `plugins/code/prompts/prompt.md:167` (critic), `plugins/code/prompts/prompt.md:214` (build)

The 9e264bd commit replaced bare `scripts/...` with `$CLAUDE_PLUGIN_ROOT/scripts/...` for the cross-repo and critic stamp calls. I initially thought this was correct and proposed the same fix for the Phase 5 build cache line. **It's not correct.** `$CLAUDE_PLUGIN_ROOT` is not set in the Bash tool's environment when the orchestrator is running inside `/code:code`, so all three stamp invocations resolve to empty-prefix paths and silently fail.

**Why the var is empty:** Per the [plugins reference docs](https://code.claude.com/docs/en/plugins-reference#environment-variables), `${CLAUDE_PLUGIN_ROOT}` is "substituted inline anywhere they appear in **skill content, agent content, hook commands, and MCP or LSP server configs**" and "exported as environment variables to **hook processes and MCP or LSP server subprocesses**." That list is exhaustive. `plugins/code/prompts/*.md` is none of those categories (it's a sidecar file loaded via `CLOSEDLOOP_PROMPT_FILE`), so nothing substitutes the var at load time. And the orchestrator's Bash tool runs in the main Claude Code session — the docs do not list that environment as a substitution or export target.

**Empirical verification:** I ran a test in-session against the cached plugin (v1.8.1) to confirm this rather than reason from docs alone.

1. Created a minimal test prompt at `plugins/code/prompts/env-test.md`:

   ```markdown
   You are a test harness. Run this EXACT bash command (copy it verbatim without any substitution, use it as-is in a single Bash tool call):

   ```
   echo "TEST1: $CLAUDE_PLUGIN_ROOT"
   echo "TEST2: $CLOSEDLOOP_WORKDIR"
   echo "TEST3: ${CLAUDE_PLUGIN_ROOT}/scripts/stamp_cross_repo_cache.sh"
   ls "${CLAUDE_PLUGIN_ROOT}/scripts/" 2>&1 | head -5
   ```

   Copy the command VERBATIM into a single Bash tool call. Do not edit, substitute, or modify the command before calling the tool.
   ```

2. Invoked the orchestrator through the real `/code:code` command (so it goes through `setup-closedloop.sh` and the actual plugin substitution path, not a plain `claude -p`):

   ```bash
   cd /tmp/env-test-workdir
   claude -p "/code:code /tmp/env-test-workdir --prompt env-test" \
     --allowed-tools=Bash,Read \
     --max-turns 6 \
     --output-format stream-json --verbose
   ```

3. Bash tool output:

   ```
   TEST1:
   TEST2:
   TEST3: /scripts/stamp_cross_repo_cache.sh
   ls: /scripts/: No such file or directory
   ```

4. Separately ran `env | grep -iE 'CLAUDE|CLOSED' | sort` from the same Bash tool call. Result included only `CLAUDE_CODE_*` runtime vars and `CLAUDECODE=1`. **No `CLAUDE_PLUGIN_ROOT`. No `CLOSEDLOOP_WORKDIR`. No `CLOSEDLOOP_PROMPT_FILE`.**

This confirms the issue empirically: the 9e264bd fix ships with `$CLAUDE_PLUGIN_ROOT/scripts/...` expressions that dereference to empty strings at runtime. Every cache stamp fails silently, hash files are never written, and the corresponding check skills return `CACHE_MISS` every iteration — so `cross-repo-coordinator`, the critics, and `build-validator` all re-run every loop iteration, defeating the caching layer entirely.

The same `env | grep` output also explains why **Issue #3's original recommendation was wrong**: my suggested replacement `bash "$CLAUDE_PLUGIN_ROOT/skills/build-status-cache/scripts/check_build_cache.sh"` for prompt.md:214 would fail identically.

**The reason prompt.md still functions at all** (writing state.json, launching subagents with `WORKDIR=/absolute/path`, etc.) is that the orchestrator LLM follows the rule at prompt.md:14 — "In subagent prompts, always use the **literal resolved path**, NEVER the string `$CLOSEDLOOP_WORKDIR`" — by extracting the path from the `setup-closedloop.sh` config.env output that appears in its context from the `/code:code` bash substitution. That substitution happens in the LLM, not the shell. The rule is scoped to subagent prompts, but in practice the LLM also does it for `echo > $CLOSEDLOOP_WORKDIR/state.json` lines, which is why those work. **It is not doing this substitution for the cache stamp script paths**, because the `bash` syntax reads like a literal shell invocation the LLM should pass through untouched.

**Fix options, in order of preference:**

1. **Move stamping into the cache skills themselves.** Add a `stamp` action to `code:cross-repo-cache`, `code:critic-cache`, and `code:build-status-cache` so the orchestrator activates the skill to stamp the same way it already activates them to check. Skill content is subject to `${CLAUDE_PLUGIN_ROOT}` substitution at load time (per docs), so `bash <base_directory>/scripts/stamp.sh <WORKDIR>` resolves correctly. This is how Phase 7's build cache check already works.

2. **Extend the literal-path substitution rule to bash commands.** Rewrite prompt.md:14 from "In subagent prompts, always use the literal resolved path" to "**In subagent prompts AND bash commands**, always use the literal resolved absolute path for `$CLOSEDLOOP_WORKDIR` and `$CLAUDE_PLUGIN_ROOT` — both are taken from the config.env output at the top of this session. NEVER pass either as a `$VAR` string to bash." Then update the three cache stamp lines to show the substituted form explicitly, e.g., `bash <CLAUDE_PLUGIN_ROOT>/scripts/stamp_cross_repo_cache.sh <CLOSEDLOOP_WORKDIR>` with a note that both must be substituted before emitting. This relies on LLM behavior but at least makes the contract explicit.

3. **Make the hook export the var into a source-able file.** Have `session-start-hook.sh` write `export CLAUDE_PLUGIN_ROOT=...` to a well-known location, and rewrite the cache stamp lines as `bash -c 'source /path/to/env && bash "$CLAUDE_PLUGIN_ROOT/scripts/stamp.sh" "$CLOSEDLOOP_WORKDIR"'`. Ugly, circular (the `source` path itself needs a literal), and introduces a cross-session dependency — I would not pick this, but noting it for completeness.

Option 1 is the cleanest and matches how the check side already works.

PR #55, opened 2026-04-29
PLN-376: wire decision-table into pipeline

shafty023:
### Primary benefit: intent preservation under iteration drift.

The orchestrator currently reads a plan, implements it, runs tests, and code-reviews, but every step is a fresh-context LLM. Each agent re-reads the plan and re-derives "what should this do?" from prose. That re-derivation is lossy: edge cases the planner had in mind never make it into the implementation prompt, the implementer fills the gap with reasonable-sounding defaults, code review checks for security/quality but not "did we build the behavior the planner intended," and tests verify what the test-writer thought of. By the time the loop completes, you've often built a working thing, not the thing the plan implied.

The decision-table pins down branch-by-branch externally-visible behavior at Phase 2.7 in a frozen, machine-checkable form (Current Code vs Intended Change rows with state axes, dependency outcomes, and test invariants). Phase 5.5's verifier then mechanically checks the implementation against those rows and routes drift back into the auto-fix loop. The plan stays a scope/intent document; the decision-table becomes the executable contract.

### Specific bug class it catches that the rest of the pipeline misses:

- Edge cases that "feel obvious" so they never get spelled out: retry vs first-attempt with expiring credentials, throw-capable preparation code in structured-result contracts, lifecycle re-entry under reconnect/restart, fire-and-forget side effects on serverless, dependency-throw vs dependency-null mapping to externally visible status codes, canonicalization on policy comparisons, signal-precedence conflicts between transport status and payload fields. These are exactly the bugs the patch you applied to SKILL.md was designed to surface, they're the long tail of post-incident learnings, not novel discoveries.
- Silent drift between iterations: if iteration N implements a behavior correctly and iteration N+1 refactors it (e.g., simplifier in Phase 4 collapses two branches), the verifier catches the regression because the artifact is frozen. Without the artifact, the "is the new code still correct?" check has nothing to compare against.
 - Test gap detection rather than test failure: the verifier surfaces "this row has no test covering its invariant" as test_drift. Today, the system only knows about tests that exist and fail — not about behavior that should have a test and doesn't.

### Where it doesn't help (and is correctly skipped):

Pure UI work, simple CRUD, doc-only changes, anything where there's no real branching behavior to map. That's why simple-mode skips the entire flow — the overhead doesn't pay back when the plan is "add a button" or "fix a typo." The 7 signals in plan-evaluator (PRD length, AC count, task count, forbidden terms, cross-repo keywords, etc.) are a reasonable proxy for "is the surface area worth mapping?" If those signals are wrong in either direction (too aggressive, too loose), the T-4.6 telemetry will show it: you'll see runs marked aligned after zero verifier iterations (overhead with no catches → skip threshold should be looser) or runs that escalate after exhausting retries on simple plans (skip threshold too tight). That's the feedback loop that lets us tune the gates against real data rather than guessing.

### Do we still need this even though our implementation plan gets fully fleshed out with code snippets and critic reviewed?

Yes, and the value isn't redundant with the enriched plan, it operates at a different layer. The enriched plan tells the implementation-subagent what to build (signatures, code patterns, integration points). The decision-table tells the verifier what to check after it's built. Even with code snippets in tasks, those snippets describe the happy path plus a handful of prose-mentioned edge cases, not an exhaustive enumeration of branch behavior.

#### Specific bug classes the enriched plan tends to miss but the decision-table catches:

1. State × dependency-outcome combinations. Plan task says "handle the dependency failure" with a snippet showing the catch block. The decision-table forces enumeration of (success / null / validation-fail / throw / reject) × (first-attempt / retry / restart), that's 15 cells, none of which a single code snippet naturally enumerates.
2. Signal-precedence conflicts. Snippet shows reading both transport status and body field. Doesn't answer what happens when they disagree. Decision-table forces a precedence row with the "wins when X, degrades when Y" axis.
3. Caller capability skew. Plan covers the current-version caller. Snippet doesn't enumerate older peers omitting newly required fields, or newer peers sending unknowns. The skill's call-site inventory step forces this.
4. Side-effect ordering across failure boundaries. Snippet shows persist → emit-event → return. Doesn't explicitly answer: if emit-event throws after persist succeeds, is persist rolled back? The side-effect-boundary row forces this.
5. Drift across the loop's own iterations. This is the one the enriched plan structurally can't catch. Iteration 1 implements correctly. Iteration 4's Phase 4 simplifier collapses two branches that were genuinely distinct per the plan but read as locally redundant. The simplifier's local view is fine; behavior regresses globally. The decision-table is the only artifact in the loop frozen at Phase 2.7 and verified again at 5.5 / on every later iteration — it's the contract that survives Phase 4 simplifier passes, Phase 5 code-reviewer fix loops, /code:amend-plan runs, and re-entries from AWAITING_USER.
6. Test-shape invariants. Plan says "test the validation logic." Without the decision-table, a test that triggers any 400 passes. Required Tests rows force "prove field-level binding, not just status-level rejection" — positive control plus one-axis mutation.

So ultimately: decision-table converts "the plan we approved" from prose-with-implicit-edge-cases into an executable contract, so iteration N+M can still verify behavior against iteration 1's intent without a human re-reading the original plan and re-judging. That's the value compounding over a long autonomous loop.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### code v1.10.0
+
+#### Added
+- New `decision-table` skill for generating code-grounded decision-table artifacts that map current vs. intended control-flow behavior, capturing recovery, retry, finalization, validation, and state-machine edge cases under `.closedloop-ai/decision-tables/`. Includes baseline/target table rules, behavioral edge-case expansion guidance (call-site inventory for shared surfaces, exception scope, serverless async side effects, testable invariants), post-implementation verification sections, contract-heavy review checklist, and a referenced artifact format template at `references/artifact-format.md`.
+- New `behavior-verifier` agent that activates the `decision-table` skill in verification-only mode (SKILL.md step 17), reads final code against the artifact's Intended Change rows, appends Verification Findings and Final Alignment Status, and emits a structured `ALIGNED` or `MISALIGNED` verdict with a typed `<drift_rows>` JSON block (`code_drift`, `test_drift`, `plan_ambiguity`) for orchestrator routing. Read-and-report only — never modifies code or tests.
+- Optional `decisionTable` property on the plan schema (`path` + `status` enum: `pending|aligned|aligned_with_clarifications|verification_failed`) so the orchestrator can persist artifact pointers and verification state across iterations.
+- Phase 5.5 Behavioral Verification loop in the orchestrator prompt with a 5-attempt cap, drift routing by kind (`code_drift` → `implementation-subagent`, `test_drift` → `test-engineer`, `plan_ambiguity` → haiku append), parse-failure circuit breaker, and per-run telemetry emit to `.closedloop-ai/decision-table-verifications.jsonl` (timestamp, final status, iteration count, drift counts, parse failures, phase duration).
+- `startSha` state-tracking field initialized once per run from `CLOSEDLOOP_START_SHA` in `config.env` and propagated on every `state.json` write so Phase 5.5 can scope the changed-file set without re-reading config.
+
+#### Changed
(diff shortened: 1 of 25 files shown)


wongk (approved):

thadeusb:
lgtm

PR #63, opened 2026-05-04
FEA-879: skip spurious-complete check on AWAITING_USER hard stops

shafty023:
detect_spurious_complete in run-loop.sh inspected only plan.json, so any <promise>COMPLETE</promise> emitted while pendingTasks was non-empty was treated as a Phase 7 contract violation. AWAITING_USER_SEQUENCE hard stops (notably the Phase 1.1 plan review checkpoint) use the same COMPLETE signal but are expected to have pending tasks and open questions, since the plan was just drafted. The runner now reads state.json.status and short-circuits when status is AWAITING_USER. Final-completion regressions (status COMPLETED with leftover pendingTasks) are still flagged.

Also moves the Phase 5.5 telemetry append target from .closedloop-ai/decision-table-verifications.jsonl to decision-table-verifications.jsonl at the workdir root, matching where plan.json, log.md, and state.json live.

Adds six tests covering detect_spurious_complete: missing plan, no pending tasks, pending-with-questions, pending-without-questions, AWAITING_USER skip, and COMPLETED-with-pending positive case.


mikeangstadt (approved):

aponamarev (approved):
LGTM

PR #69, opened 2026-05-05
PLN-465: feat(code): add PERF_V2 token aggregation to subagent-stop-hook

peterulsteen:
## Summary

Producer-side enrichment (FEA-888) of [PRD-254](https://app.closedloop.ai/prds/PRD-254) — extends the existing FEA-764 `agent` perf event in `perf.jsonl` so each sub-agent invocation carries `model`, `parent_session_id`, `command`, all four token-count fields (`input_tokens`, `output_tokens`, `cache_creation_input_tokens`, `cache_read_input_tokens`), and a per-turn `total_context_tokens` high-water mark. Implements [PLN-465](https://app.closedloop.ai/implementation-plans/PLN-465).

This PR also populates the `command:` field on **every** `perf.jsonl` event row produced by the orchestrator and producer scripts, not just the new `agent` event — so phase, iteration, and pipeline_step rows can also be filtered by slash command in Datadog.

Out of scope: tool/skill/spawn hooks (FEA-889 — see PR #70), the desktop scanner that ships these to Datadog (FEA-886).

**Merged:** squash commit [`88b78d2`](https://github.com/closedloop-ai/claude-plugins/commit/88b78d2cf06a71e4d73528141e5040e7e7b1d73b).

### Agent event token enrichment (FEA-888)

- **`subagent-stop-hook.sh`** — parses `agent_transcript_path` JSONL via a single `jq -s` reduction, mirroring the accumulation pattern from `stream_formatter._accumulate_usage`. The four cumulative fields sum across all assistant turns; `total_context_tokens` is the per-turn maximum of any single entry's full usage (input + output + cache_creation + cache_read), so a context-pressure spike is visible as the peak rather than buried in a monotonic running sum. Reads `model` and `parent_session_id` from the SubagentStop hook payload, `command` from `CLOSEDLOOP_COMMAND` (defaulting to `"interactive"` to match `record_phase.sh` / `run-loop.sh`'s `emit_perf_event` helper). Emits `null` for absent `model`/`parent_session_id`. Defaults all numeric fields to `0` on transcript-missing or jq-parse failure. Hook always exits 0 — no error path aborts a Loop.
- **`perf_summary.py`** — schema comment updated: `command:` is now a required field on `agent` and `phase` rows.
- **`test_subagent_stop_hook.py`** — 13 tests across `TestAgentEventTokenAggregation`, `TestAgentEventMissingTranscript`, `TestAgentEventAttributionFields`, `TestAgentEventModelAndMetadata`. Fixtures use the real Claude Code transcript shape (`{"type": "assistant", "message": {"role": "assistant", "usage": ...}}`) so the bash port can't drift silently from the Python reference.

### Producer/orchestrator `command:` field

- **`run-loop.sh::emit_perf_event()`** — adds `command:` to every event flowing through the helper (`phase`, `iteration`, `pipeline_step`). Single `jq -n -c` filter per event — no extra `jq` invocation cost.
- **`record_run.sh`** — emits the singular per-Loop `run` event on every fresh-start invocation, with `command:` populated.
- **`record_phase.sh`** — `command:` field is always part of the emitted JSON.
- **`test_record_run.py` and `test_record_phase.py`** — assertions updated to verify `command:` is present on every event; the `TestRecordPhaseCommandField` class pins the contract.
- **`plugins/code/README.md`** — `record_run.sh` description simplified to "Emitted unconditionally and fails open".

### Versions

- `code` 1.11.8 → **1.11.9** (PATCH).
- `self-learning` 1.2.2 → **1.2.3** (coordination bump alongside `code`).

## Why

PRD-254 hypothesis #1 ("identify top phase × agent combos driving wall-clock and token cost") and the example query "For PLAN Loops, which sub-agent is the bottleneck — and on which model?" are both gated on having `command:` populated on every event row and the four token-count fields appearing on `agent` rows. Without this PR, agent rows in Datadog would carry timing only, and `phase`/`iteration`/`pipeline_step` rows would have no slash-command attribution.

## Pre-merge fixes from review

Codex `/review` flagged four issues; all addressed before mark-as-ready:

| Severity | Finding | Fix |
|---|---|---|
| BLOCKING | jq filter selected `.role == "assistant"` — never matches actual transcript shape, every agent event emitted **0 tokens** | `select(.type == "assistant") \| .message.usage`, mirroring `stream_formatter._accumulate_usage` |
| HIGH | `total_context_tokens` computed from cumulative running sum — monotonic, just equals the final total, no peak signal | Per-turn HWM: max of any single entry's full usage |
| MEDIUM | `command` defaulted to `""` while phase/iteration/pipeline_step rows default to `"interactive"` — broke joins by command in Datadog | Final fallback `PERF_COMMAND="${PERF_COMMAND:-interactive}"` |
| LOW | Test fixtures used the same wrong shape, masking the BLOCKING bug | Rewrote to real Claude Code shape; updated HWM assertion (5050 → 1750); added malformed-JSONL test |

## Test plan

- [x] `pytest plugins/code/tools/python/` — 261/261 passing
- [x] `ruff check` + `pyright` clean on all modified Python files
- [x] CI: Lint / Tests / Type Check all green
- [x] Verified cache-read transcript → all four token fields sum correctly; HWM = max single-turn total
- [x] Verified missing/malformed transcript → exit 0, agent event still emitted with token fields = 0
- [x] Pre-merge dogfooding — confirmed `perf.jsonl` carries the new `agent` event fields and `command:` on every other event row end-to-end

## Risks

The hooks/scripts now write additional fields on every Loop and every SubagentStop. Mitigations:
- **Schema additivity** — existing `perf.jsonl` consumers ignore unknown fields; no breaking changes to FEA-764 timing semantics.
- **Fail-open contract** — `trap 'exit 0' ERR` plus explicit numeric defaults (all token fields default to `0` on transcript-missing or jq-parse failure) — a hook bug can never abort a Loop or corrupt `perf.jsonl`.
- **Per-turn HWM math** — the only new computation is a max over already-validated integer fields; no division, no string formatting that could silently produce malformed JSON.
- **`command:` field is one short string** — at typical Loop sizes (<100 KB perf.jsonl), the extra bytes are negligible; Datadog ingestion costs scale with event count, not field count.

---
Loop ID: 019dfa29-03ce-749c-88fb-e4d006674e2d
Artifact: https://app.closedloop.ai/implementation-plans/PLN-465


plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.7",
+  "version": "1.11.8",
+  "claudeCodeMinimumVersion": "1.0.0",
```
shafty023: Is this hallucinated? i'm not finding this in any docs nor is this set to a valid version

mikeangstadt:
## Code Review Summary

**Status:** Approved

**Reviewers:** Bug Hunter A (x4), Bug Hunter B, Unified Auditor, Premise Reviewer
**Model Routing:** Medium (1,926 LOC) — BHA: Opus, BHB: Sonnet, Auditor: Sonnet, Premise: Opus

| Severity | Count |
|----------|-------|
| Blocking | 0 |
| High | 0 |
| Medium | 3 |

### MEDIUM Issues (consider)

1. **[P2] [CHANGELOG.md:28] Convention Violation** — CHANGELOG.md manually edited in violation of CLAUDE.md convention. CLAUDE.md states: *Do NOT manually edit CHANGELOG.md or README.md. After all code changes are finalized, run /update-documentation to generate documentation updates.* **Recommendation:** Revert the manual edits and run `/update-documentation`.

2. **[P2] [plugins/code/scripts/run-loop.sh:435] Correctness** — `write_runs_log_entry` defaults command to `self_learning` but main-loop callsites don't pass an explicit command. On fresh starts before any review runs, `LAST_CLAUDE_COMMAND` is empty and the fallback `self_learning` is used — misattributing plan_execute events in Datadog. **Recommendation:** Either pass `plan_execute` explicitly to all main-loop `write_runs_log_entry` callsites, or initialize `LAST_CLAUDE_COMMAND` to `plan_execute` at script start.

3. **[P2] [plugins/code/scripts/run-loop.sh:1073] Correctness** — `emit_perf_event` PERF_V2 path exits script via `set -e` if `json_line` is empty. When `json_line` is empty, `jq` exits non-zero. Under `set -euo pipefail`, this terminates the entire loop script. **Recommendation:** Guard the jq call: `if [[ -n "$json_line" ]]; then ... fi`.

### Validation Stats

- **Total findings from agents:** 7
- **Validated:** 3
- **Discarded — line not changed:** 3
- **Discarded — duplicate:** 1
- **Agent failures:** 0

**Recommendation:** Approve — consider addressing medium items for robustness

> Note: Findings #2 and #3 reference `run-loop.sh` which is in the full three-dot diff but not in GitHub's PR diff (likely already merged to main). Inline comments could not be posted for these.

plugins/self-learning/tools/python/perf_summary.py
```diff
@@ -21,7 +21,12 @@
 # pipeline_step: {event, run_id, iteration, step, step_name, started_at, ended_at, duration_s,
 #                 exit_code, skipped, [sub_step], [sub_step_name]}
 # agent:         {event, run_id, iteration, agent_id, agent_type, agent_name,
```
shafty023, line 23: **[P2]** Version the self-learning plugin change

Because this edit changes a file under `plugins/self-learning`, the repo's plugin version-management contract applies to `self-learning` as well. This PR only bumps the `code` plugin metadata/changelog, so publishing or release validation would ship this self-learning change while `plugins/self-learning/.claude-plugin/plugin.json` still reports the old version; either remove this edit or add the corresponding self-learning version/changelog update.

plugins/code/tools/python/test_subagent_stop_hook.py
```diff
@@ -146,3 +161,364 @@ def test_acknowledgment_enforcement_reachable(
...
+        assert evt["parent_session_id"] == "parent-session-xyz"
+        assert evt["command"] == "code"
+
+    def test_missing_model_emits_null(
+        self, session_env: tuple[Path, Path, str]
+    ) -> None:
+        """When model is missing from hook payload, emit model: null."""
+        cwd, workdir, session_id = session_env
+
+        # Re-create agent-type file (cleaned up by previous tests)
```
shafty023: **[P2]** Agent-type file recreation inlined in 3 tests instead of calling _recreate_agent_type_file

The PR introduces helper _recreate_agent_type_file(workdir) at line 182 specifically to avoid repeating agent-type file setup logic. It is correctly called in test_token_aggregation_sums_correctly and test_malformed_transcript_emits_zero_tokens_and_does_not_abort. However, three other tests in the same PR inline the identical 4-line block: test_missing_model_emits_null (lines 422-427), test_missing_parent_session_emits_null (lines 452-457), and test_missing_command_defaults_to_interactive (lines 489-493). All use the same hard-coded content string and mkdir/write_text pattern.



> **Suggestion:** Replace the three inline agent-type recreation blocks with calls to _recreate_agent_type_file(workdir).

plugins/code/hooks/subagent-stop-hook.sh
```diff
@@ -436,21 +436,115 @@ if [[ -n "$AGENT_ID" ]] && [[ -n "$CLOSEDLOOP_WORKDIR" ]] && [[ -f "$AGENT_TYPES
...
+            # Read command from env var, falling back to config.env, then to "interactive"
+            # to match the default used by record_phase.sh and run-loop.sh's emit_perf_event
+            # helper. Without this final fallback, agent rows can carry command="" while
+            # phase/iteration/pipeline_step rows carry command="interactive", breaking joins
+            # by command in Datadog (PRD-254 attribution discipline).
+            PERF_COMMAND="${CLOSEDLOOP_COMMAND:-}"
+            if [[ -z "$PERF_COMMAND" ]]; then
+                CLOSEDLOOP_CONFIG_PATH="$CLOSEDLOOP_WORKDIR/$CLOSEDLOOP_STATE_DIR/config.env"
+                if [[ -f "$CLOSEDLOOP_CONFIG_PATH" ]]; then
+                    PERF_COMMAND=$(grep '^CLOSEDLOOP_COMMAND=' "$CLOSEDLOOP_CONFIG_PATH" 2>/dev/null | head -1 | cut -d'=' -f2- || echo "")
```
shafty023: **[P2]** config.env CLOSEDLOOP_COMMAND fallback is dead code, the key is never written there

Lines 458-462 read CLOSEDLOOP_COMMAND from config.env as a fallback when the env var is not set. However, CLOSEDLOOP_COMMAND is never written to config.env: setup-closedloop.sh does not include it, and no script appends CLOSEDLOOP_COMMAND= to that file. Additionally, config.env is already sourced at line 44 of this same hook, so any key present would already be reflected in ${CLOSEDLOOP_COMMAND:-} at line 457, making the grep redundant. The comment at lines 69-73 says 'falling back to config.env' as if this is a working code path, but in practice the grep always returns empty and control falls through to the 'interactive' default.



> **Suggestion:** Remove the config.env grep block (lines 458-462) and update the comment to accurately describe the two-step fallback (env var, then 'interactive'). Alternatively, write CLOSEDLOOP_COMMAND into config.env from setup-closedloop.sh if the fallback is intended.

plugins/self-learning/tools/python/perf_summary.py
```diff
@@ -21,7 +21,12 @@
...
-#                 started_at, ended_at, duration_s}
+#                 started_at, ended_at, duration_s,
+#                 [command], [model], [parent_session_id],
+#                 [input_tokens], [output_tokens],
+#                 [cache_creation_input_tokens], [cache_read_input_tokens],
+#                 [total_context_tokens]}
+#                 Token/model fields are present when CLOSEDLOOP_PERF_V2=1.
```
shafty023: **[P2]** Schema comment says 'Token/model fields' but command field is also PERF_V2-only

The added comment at line 29 reads 'Token/model fields are present when CLOSEDLOOP_PERF_V2=1.' However, command is also listed as an optional field ([command]) in the schema block above it, and in the hook code the command field is emitted only in the PERF_V2 branch (subagent-stop-hook.sh lines 145/528), absent from the baseline FEA-764 event. The comment omits command, making the annotation incomplete.



> **Suggestion:** Update the comment to 'Token/model/command fields are present when CLOSEDLOOP_PERF_V2=1.' or 'Optional fields (marked []) are present only when CLOSEDLOOP_PERF_V2=1.'

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.7",
+  "version": "1.11.8",
```
shafty023: **[P3]** Version bump may be PATCH where MINOR is warranted (advisory, validation-discarded)

Per CONTRIBUTING.md semver rules in CLAUDE.md ('MINOR for new agents/skills/commands, MAJOR for breaking changes'), and given that this PR adds new hook capability (token aggregation, four new token fields, model/parent_session_id/command fields, claudeCodeMinimumVersion manifest field), the change may warrant 1.12.0 (MINOR) rather than 1.11.8 (PATCH). Note: the auditor flagged this finding on the wrong line (28 vs the actual changed line 4-5) so the deterministic line-range gate discarded it. Surfacing as advisory only.



> **Suggestion:** Consider bumping plugins/code/.claude-plugin/plugin.json to 1.12.0 to reflect new hook capability per CONTRIBUTING.md semver rules.

plugins/code/.claude-plugin/plugin.json
peterulsteen: Good catch — verified, removed in `52065ef`.

You're right: the Claude Code plugin manifest schema (https://code.claude.com/docs/en/plugins-reference) doesn't include any minimum-version field. Supported top-level keys are limited to `name`, `version`, `description`, `author`, `homepage`, `repository`, `license`, `keywords`, `$schema`, `dependencies`, and component-path fields. The loader silently ignores unrecognized fields, so `claudeCodeMinimumVersion` provided zero runtime enforcement.

The field came from PLN-465's task list (which derived "pin Claude Code minimum version" from PRD-254's NFR), but the field name was speculative — neither the plan nor the implementer (me) verified against the actual schema. Replaced with a CHANGELOG note documenting the SubagentStop payload contract we rely on, plus the graceful-degrade behavior on older Claude Code versions.

Same fix applied to sister PR #70 in `5d34150` (it had the variant `minClaudeCodeVersion` — same root issue).

plugins/self-learning/tools/python/perf_summary.py
peterulsteen: Right call — bumped self-learning 1.2.2 → 1.2.3 in `52065ef` and added a `self-learning v1.2.3` entry to the root CHANGELOG describing the schema-docstring update in `tools/python/perf_summary.py` (the only self-learning file this PR touches). Per CLAUDE.md's plugin-version-bump rule, any change to a plugin's files needs a version bump in that plugin's manifest — that contract was not honored here originally.

peterulsteen:
@mikeangstadt re your code-review summary at #issuecomment-4391122910 — status update on the three findings (final state, post-merge in [`88b78d2`](https://github.com/closedloop-ai/claude-plugins/commit/88b78d2cf06a71e4d73528141e5040e7e7b1d73b)):

### 1. [P2] CHANGELOG.md manual edit — convention violation
**Addressed.** The CHANGELOG entry was regenerated via the `/update-documentation` skill recipe, then merged with main and rebased through several iterations. The final v1.11.9 entry on main reflects what the skill produces against this PR's actual diff vs `main`, not editorial prose. Per CLAUDE.md: *"Do NOT manually edit CHANGELOG.md ... run /update-documentation."*

### 2. [P2] `run-loop.sh:435` — `write_runs_log_entry` defaults command to `self_learning`
### 3. [P2] `run-loop.sh:1073` — `emit_perf_event` PERF_V2 path may abort under `set -e`

**Out of scope for PR #69** — both reference `plugins/code/scripts/run-loop.sh` lines that weren't in PR 69's diff (those changes shipped in PLN-482 / #68 and now live on main, as you noted yourself).

Filed both as a single follow-up feature in ClosedLoop: **[FEA-936](https://app.closedloop.ai/features/FEA-936)** — *"run-loop.sh: fix command-attribution default and emit_perf_event empty-input crash"*. The FEA captures the recommended fix for each (init `LAST_CLAUDE_COMMAND="plan_execute"` at script start; add `[[ -n "$json_line" ]]` guard to the jq call), the rationale, and acceptance criteria for both. Whoever picks it up next can drive a focused PR against `main` with both fixes plus the regression tests in `test_run_loop_failure_marker.py`.

*Note*: PR #69's `emit_perf_event` was rewritten in this PR to add `command:` to every event unconditionally (no env-var branch), so finding #3 specifically applies to the legacy code path on main, not to anything PR #69 lands.

plugins/code/tools/python/test_subagent_stop_hook.py
peterulsteen: Good catch — fixed in `fddabfa`. Replaced the inline blocks in all three tests (`test_missing_model_emits_null`, `test_missing_parent_session_emits_null`, `test_missing_command_defaults_to_interactive`) with `_recreate_agent_type_file(workdir)`. This also resolves PLN-465 task T-3.5 which had been deferred.

Net: -12 lines of duplication. 13/13 tests still passing.

plugins/code/hooks/subagent-stop-hook.sh
peterulsteen: Verified and removed in `fddabfa`. You're right on both counts:

1. `CLOSEDLOOP_COMMAND` is never written to `config.env` anywhere in the repo — `run-loop.sh` exports it as an env var directly and persists `command:` to `state.json`, never to `config.env`.
2. Even if it *were* written there, line 44 of this same hook already `source`s `config.env` into the env, so any value would already be in `${CLOSEDLOOP_COMMAND:-}` by the time we reach line 457.

Simplified to the two-step fallback you suggested:

```bash
PERF_COMMAND="${CLOSEDLOOP_COMMAND:-interactive}"
```

Updated the surrounding comment to describe the actual two-step path and explain that the redundant grep was removed.

plugins/self-learning/tools/python/perf_summary.py
peterulsteen: Right call — reworded the docstring to a more general phrasing that matches the schema block above. Final wording in `perf_summary.py`:

> `model`, `parent_session_id`, and the token-count fields are marked optional ([...]) because they fall back to `null`/`0` when the SubagentStop payload or transcript is missing or unparseable.

This way the comment matches the schema block and won't drift the next time we add an optional field.

plugins/code/.claude-plugin/plugin.json
peterulsteen: Going with **PATCH** for both PRs. Reasoning:

**Per the literal CLAUDE.md rule:** *"PATCH for bug fixes/prompt wording, MINOR for new agents/skills/commands, MAJOR for breaking changes to orchestration/hook API/skill interfaces."* Neither PR adds a new agent, skill, or command. PR #69 extends an existing perf event additively; PR #70 adds two hooks but no new agents/skills/commands. Both fit the strict PATCH read.

**Final versions (post-merge):**
- PR #69 → **1.11.9** (PATCH from main's 1.11.8)
- PR #70 → **1.11.10** (PATCH above PR #69's 1.11.9, applied after rebase)

No collision, no churn on already-published changelog content.

Two follow-ups worth considering, separate from these PRs:
1. Update `CLAUDE.md` to spell out the rule for additive perf-schema / hook-payload changes (PATCH or MINOR?).
2. The `claudeCodeMinimumVersion` / `minClaudeCodeVersion` field family was removed from both PRs — neither is in Claude Code's plugin manifest schema, and the loader silently ignores unrecognized fields.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code v1.11.8
```
peterulsteen: **Update**: regenerated via `/update-documentation` rather than keeping the manual content. The CHANGELOG entries on this branch are now skill-generated (commit `7b30a4f` regen + `21956f6` merge resolution). Per CLAUDE.md the workflow contract is followed — entries reflect what `/update-documentation`'s recipe produces against this PR's actual diff.

Mike's M2 and M3 reference `plugins/code/scripts/run-loop.sh` which isn't in PR #69's diff (those changes shipped in PLN-482 / #68 and now in main, as Mike noted) — they'll be a separate follow-up PR against `main`. Full reply posted on the issue thread above.

plugins/code/.claude-plugin/plugin.json
peterulsteen: Final versions (post-merge):

- PR #69 → **1.11.9** (PATCH from main's 1.11.8)
- PR #70 → **1.11.10** (PATCH above PR #69's 1.11.9, applied after rebase)

No collision, no churn on already-published changelog content. Marking this thread resolved on my end. Thanks for the careful flag.

mikeangstadt (approved):

PR #70, opened 2026-05-06
PLN-466: feat(code): add PreToolUse/PostToolUse perf-tracking hooks

peterulsteen:
## Summary

Producer-side telemetry (FEA-889) of [PRD-254](https://app.closedloop.ai/prds/PRD-254) — adds `PreToolUse` and `PostToolUse` Claude Code hooks that emit `tool`, `skill`, and `spawn` perf events to `perf.jsonl` for every tool invocation inside a Loop. Implements [PLN-466](https://app.closedloop.ai/implementation-plans/PLN-466).

Out of scope: agent-event token enrichment (FEA-888 — see PR #69, merged), the desktop scanner that ships these to Datadog (FEA-886).

**Merged:** squash commit [`9f41576`](https://github.com/closedloop-ai/claude-plugins/commit/9f4157656bf925688b19ae06d829ea52fed7f961).

## What changed

- **`pre-tool-use-hook.sh`** (new) — On every tool invocation, writes a sentinel file at `$CLOSEDLOOP_WORKDIR/.tool-calls/{TOOL_USE_ID}` containing call-time attribution: `started_at`, `tool_name`, `agent_id`, `run_id`, `command`, `iteration`. For `Agent` tool calls, additionally appends a `spawn` event to `perf.jsonl` with `parent_session_id`, `parent_agent_id`, and `planned_subagent_type` from the hook payload — supporting one-level parent→child reconstruction in Datadog.
- **`post-tool-use-hook.sh`** (new) — Reads the sentinel, computes `duration_s` (POSIX-portable date math for both macOS and Linux), and appends a `tool` event with `event, run_id, command, iteration, agent_id, tool_name, started_at, ended_at, duration_s, ok`. For `Skill` tool calls, additionally appends a `skill` event with `skill_name` (extracted from `tool_input.skill` with `tool_input.command` fallback per PRD GAP-004). Sentinel is deleted on success; orphaned sentinels remain for the desktop scanner (FEA-886) to detect SIGTERM-killed Loops per PRD AC-5.c.
- **`hooks.json`** — registers both new hooks; the existing `pretooluse-hook.sh` (security/learning injection) is unaffected.
- **Bash test suite** under `plugins/code/hooks/tests/` — 5 files covering tool/skill/spawn event emission, fail-open behavior, and end-to-end pre→post correlation:
  - `test_helpers.sh` — shared pass/fail counters, `assert_field_present`, `assert_field_equals`, `setup_temp_env`, `create_sentinel`
  - `test_tool_event.sh` — post-hook emits a complete `tool` event with sentinel-based attribution overrides
  - `test_skill_event.sh` — post-hook emits both `tool` and `skill` events for `Skill` tool calls, with `skill > command` priority and empty-fallback paths
  - `test_spawn_event.sh` — pre-hook emits a `spawn` event for `Agent` tool calls and writes a sentinel for non-Agent tools
  - `test_fail_open.sh` — both hooks exit 0 and don't corrupt `perf.jsonl` on internal errors (read-only sentinel dirs, missing/corrupted sentinels, exit-1 stub replacements, malformed JSON)
  - `test_correlation.sh` — end-to-end pre→post run, sentinel-attribution-wins regression, corrupt-sentinel suppression, missing-correlation-id silent skip, non-numeric iteration normalization
- **Version:** `code` 1.11.9 → **1.11.10** (PATCH).

## Why

PRD-254's example question "Which skills are getting invoked the most inside Phase 3 across all Loop commands?" is gated on having `skill` events with `skill_name`. The `spawn` events provide the `parent_session_id` correlation key that PR #69's `agent` events match against, enabling one-level parent→child agent tree reconstruction in Datadog.

## Pre-merge fixes from review

Codex `/review` flagged 7 issues (verified 6, 1 speculative); all verified items addressed before mark-as-ready. A 9-agent multi-agent review then surfaced 6 additional findings — all addressed:

| Severity | Finding | Fix |
|---|---|---|
| BLOCKING ×2 | `.tool-call-counter` fallback in both hooks is race-prone under parallel tool calls; post may read a counter already advanced by a later pre-hook | Counter fallback removed entirely; both hooks `exit 0` silently when both `tool_use_id` and `tool_call_id` are absent |
| HIGH | Corrupt sentinel JSON swallowed by `\|\| echo ""`; tool event still emitted with empty `started_at` and `duration_s: 0` | Validate `STARTED_AT` after sentinel parse; suppress event and remove sentinel if empty |
| HIGH | Pre-hook recorded run_id/command/iteration in sentinel but post-hook used env vars (drift risk if iteration advances mid-call) | Sentinel becomes authoritative for all attribution fields; env vars are fallback only |
| HIGH | `test_fail_open.sh` Tests 3 & 4 wrapped stub hook with their own `trap 'exit 0' ERR` and asserted that wrapper exits 0 — self-validating | Rewritten to feed malformed JSON to the real pre/post hooks; the hook's own trap is now what's exercised |
| MEDIUM | `post-tool-use-hook.sh` `OK` derivation aborted entire `eval` block when `tool_response` was a non-object scalar (string/array) — silent emission skip with orphaned sentinel | Type-guard with `if (.tool_response \| type) != "object" then "true" elif ... end` |
| MEDIUM | `--argjson iteration "${CLOSEDLOOP_ITERATION:-0}"` aborts on non-numeric value, sentinel lost via `trap exit 0 ERR` | Numerically validate against `^[0-9]+$`; normalize invalid values to 0 |
| MEDIUM | `test_correlation.sh` Test 5 only exercised pre-hook iteration normalization; post-hook's own `^[0-9]+$` was dead code | Test 6 added: hand-crafts a sentinel without `iteration`, sets non-numeric env, asserts post-hook normalizes to 0 |
| MEDIUM | `test_skill_event.sh` had no test for `skill > command` priority or both-fields-absent behavior | Tests 5 and 6 added |
| MEDIUM | `tool_input.skill` vs `tool_input.command` ambiguity (PRD GAP-004) | Kept fallback chain; comment documents the rationale |
| MEDIUM | All 5 test files used `set -euo pipefail` — `set -e` aborts before assertion `fail()` reporters can run inside `{ ... }` blocks | Dropped `-e`; kept `-u` and `-o pipefail` |
| LOW | Tests pre-create sentinels; never exercise the real pre→post correlation path | New `test_correlation.sh` exercises real pre→post flow with sentinel-wins regression coverage |

The skipped HIGH ("only PostToolUse registered, failed tool executions via a separate failure hook would never emit") is speculative — Claude Code has no separate failure hook; PostToolUse fires for both success and failure paths.

## Test plan

- [x] `bash plugins/code/hooks/tests/test_*.sh` — 102 tests passing across 5 files
  - `test_correlation.sh` — 21
  - `test_fail_open.sh` — 13
  - `test_skill_event.sh` — 26
  - `test_spawn_event.sh` — 25
  - `test_tool_event.sh` — 17
- [x] `pytest plugins/code/tools/python/` — 261/261 passing
- [x] CI: Lint / Tests / Type Check all green
- [x] Verified the actual pre→post correlation path: pre-hook writes sentinel, post-hook reads it, tool event carries sentinel attribution
- [x] Verified concurrency safety: missing correlation id skips silently with no counter file written
- [x] Pre-merge dogfooding — confirmed `perf.jsonl` carries `tool`/`skill`/`spawn` events end-to-end

## Risks

The two new hooks now write to `perf.jsonl` on every tool invocation. Mitigations:
- **Schema additivity** — the new event types (`tool`, `skill`, `spawn`) are append-only and ignored by downstream code that doesn't recognize them; no breaking changes to existing consumers.
- **Fail-open contract** — both hooks use `trap 'exit 0' ERR`, so any hook bug can never break a tool invocation or the Loop (PRD-254 AC-6.hooks).
- **Sentinel-attribution-wins** — post-hook uses sentinel-recorded attribution (run_id, command, iteration) over env vars, so iteration-advance mid-call cannot misattribute the event.
- **Race-free correlation** — counter fallback removed; missing `tool_use_id` AND `tool_call_id` results in silent skip rather than ambiguous attribution.

---
Loop ID: 019dfa32-db85-707f-8da5-33b3cf06ccd3
Artifact: https://app.closedloop.ai/implementation-plans/PLN-466


mikeangstadt:
## Code Review — 3 Medium Findings

All code quality / DRY — no correctness or security issues.

### [P2] Test helper duplication (~150 lines)
**File:** `plugins/code/hooks/tests/test_fail_open.sh:78` (and 3 other test files)

`setup_temp_env`, `pass`/`fail`, `create_sentinel`, and `assert_field_equals`/`assert_field_present` are copy-pasted across all 4-5 test files with identical bodies. When the sentinel format changes (as it did in this PR with run_id/command/iteration), every copy must be updated independently.

**Fix:** Extract into `plugins/code/hooks/tests/test_helpers.sh` and source from each test file.

---

### [P2] assert_eq vs assert_field_equals naming inconsistency
**File:** `plugins/code/hooks/tests/test_correlation.sh:33`

`assert_eq` has an identical body to `assert_field_equals` in the other test files. Different name breaks the visual pattern across the test suite.

**Fix:** Rename to `assert_field_equals` for consistency.

---

### [P3] Stdin pipe inconsistency between hooks
**File:** `plugins/code/hooks/pre-tool-use-hook.sh:25`

Pre-hook uses `echo "$INPUT" | jq` while post-hook uses `jq ... <<< "$INPUT"` (here-string). Symmetric hooks should use the same idiom.

**Fix:** Change to here-string form to match post-hook.

peterulsteen:
Thanks Mike — all three landed in `a1853fa`:

- **P2 DRY**: extracted `plugins/code/hooks/tests/test_helpers.sh` (pass/fail, assert_field_*, setup_temp_env, create_sentinel). Each test file now sources it. `setup_temp_env` also picked up a `$RANDOM` suffix so back-to-back calls in the same file can't collide on session id. Net **-182 lines** across the suite.
- **P2 naming**: renamed `assert_eq` → `assert_field_equals` and `setup_env` → `setup_temp_env` in `test_correlation.sh` to match the rest of the suite. My bad — I introduced that drift when adding the file.
- **P3 stdin**: pre-hook now uses `<<< "$INPUT"` to match the post-hook idiom.

103/103 tests still passing locally + on CI. Appreciate the careful read.

peterulsteen:
## 9-Agent Code Review Re-Run — Findings

Re-ran `/code-review:start` against this branch with tighter agent prompts (the prior fleet stalled). All 9 agents completed cleanly this time.

**3 false positives** (BLOCKING-class agent hallucinations): the agents Grep'd the local filesystem to verify newly-added files (`test_helpers.sh`, `pre-tool-use-hook.sh`) and got empty results because the review ran from a sibling worktree, not this PR's branch. The files exist on `symphony/pln-466` (commit `a1853fa`). Worth flagging for the platform team that the review skill should auto-`cd` into the PR's worktree.

**7 real findings, all addressing in the next commit:**

### HIGH

**[H1] `tests/test_fail_open.sh:165-200`** — Tests 3 & 4 wrap the stub hook with their own `trap 'exit 0' ERR` and assert that wrapper exits 0. The trap is what the test added — so the test is self-validating and would still pass if `trap 'exit 0' ERR` were removed from the *real* `pre-tool-use-hook.sh`. Fix: invoke the real hook with malformed input that exercises its own internal trap.

### MEDIUM

**[M1] `post-tool-use-hook.sh:39`** — `OK` derivation aborts the entire `eval` block when `tool_response` is a non-object scalar. `(.tool_response // {})` only saves null; if `tool_response` is a string or array, `.error` errors in jq, and all parsed env vars (TOOL_NAME, SESSION_ID, etc.) are lost — leading to silent emission skip with orphaned sentinel. Fix: type-guard with `if (.tool_response | type) != "object"`.

**[M2] `tests/test_correlation.sh:283`** — Test 5 only exercises pre-hook iteration normalization. The post-hook reads sentinel iteration (already normalized to 0 by pre-hook), so the post-hook's own `^[0-9]+$` check is dead code under the sentinel-wins design. Add a test that creates a sentinel without iteration to force the env fallback.

**[M3] `tests/test_skill_event.sh`** — No test asserts the documented `skill > command` priority when both `tool_input.skill` and `tool_input.command` are present.

**[M4] `tests/test_skill_event.sh`** — No test verifies behavior when `tool_input` has neither `skill` nor `command` (skill_name should be empty).

**[M5] `plugin.json:5`** — `minClaudeCodeVersion` has zero precedent in this repo or marketplace.json. May be silently ignored by Claude Code's loader. Sister PR 69 uses `claudeCodeMinimumVersion` for the same intent — neither is verified canonical. Documenting as advisory; PR 69 will need to align.

**[M6] `tests/test_*.sh`** — All 5 test files use `set -euo pipefail`. If a non-explicit-error-checked command fails inside a `{ ... }` test block (e.g., `grep` returning 1 on no match), `set -e` aborts before the assertion's `fail()` reporter runs — silent test loss. Fix: drop `-e` from test files (explicit error handling is everywhere).

### Why earlier reviews missed these

Codex focused on production correctness; it caught the race conditions and attribution bugs. Mike caught the most visible DRY/naming. The remaining gap was **test design quality** — assertions that don't actually assert what they claim, missing edge cases. That's exactly what the multi-agent fleet (BHB / Auditor) is meant to catch, and it stalled on PR 69 / wasn't run on PR 70 in the original review pass.

Pushing fixes now.

peterulsteen:
## ✅ All 7 findings addressed

| # | Severity | Fix |
|---|---|---|
| H1 | HIGH | `test_fail_open.sh` Tests 3 & 4 rewritten to feed malformed JSON to the **real** pre/post hooks (vs. self-validating wrappers). Real `trap 'exit 0' ERR` is now exercised; assertions also verify no sentinel / no perf event leaks on bad input. |
| M1 | MEDIUM (production bug) | `post-tool-use-hook.sh` `OK` derivation now type-guards `tool_response`: `if (.tool_response \| type) != "object" then "true" elif ... end`. Non-object scalars no longer abort the entire `eval`. |
| M2 | MEDIUM | `test_correlation.sh` Test 6 added: hand-crafts a sentinel **without** an `iteration` field, sets `CLOSEDLOOP_ITERATION="garbage"`, asserts post-hook normalizes to 0. The post-hook's `^[0-9]+$` guard is now actually exercised. |
| M3 | MEDIUM | `test_skill_event.sh` Test 5 added: payload with **both** `tool_input.skill` and `tool_input.command` set; asserts `skill` wins. |
| M4 | MEDIUM | `test_skill_event.sh` Test 6 added: payload with **neither** field set; asserts `skill_name` is empty and the hook doesn't crash. |
| M5 | MEDIUM | `minClaudeCodeVersion` flagged as advisory then **removed** entirely — the field has no precedent in this repo or `marketplace.json` and is silently ignored by Claude Code's loader. PR #69 also dropped its `claudeCodeMinimumVersion` variant for the same reason. |
| M6 | MEDIUM | All 5 test files: `set -euo pipefail` → `set -uo pipefail` (kept `-u` and `-o pipefail`, dropped `-e`). Tests use `\|\| capture` and `if ... else fail` everywhere; `set -e` would have caused silent test loss when an unguarded command failed inside a `{ ... }` block. |

**Final test count (post-merge):** 102 hook tests passing across 5 files (test_correlation: 21, test_fail_open: 13, test_skill_event: 26, test_spawn_event: 25, test_tool_event: 17). Counts settled lower than the 109 reached during the multi-agent fix pass after a final cleanup round dropped a handful of defensive regression tests that the user judged unnecessary. CI green: Lint ✅ Tests ✅ Type Check ✅.

The 3 false-positive BLOCKINGs (test_helpers.sh / pre-tool-use-hook.sh "doesn't exist") were a worktree-mismatch artifact — agents Grep'd from a sibling worktree where the new files didn't exist yet. No action needed for this PR.

mikeangstadt (approved):

PR #72, opened 2026-05-07
PLN-502: fix(code): prevent false-positive terminal-failure from benign rate-limit heartbeats

aponamarev:
- Tighten rate_limit_signal jq predicate so rate_limit_event entries only fire when rate_limit_info.status or overageStatus is non-allowed (allowed/missing → benign heartbeat)
- Source failure message from the triggering entry's result/error string instead of unrelated assistant text; preserve stderr fallback paths
- Scope auth_challenge_signal to is_error / isApiErrorMessage envelopes so assistant prose mentioning auth keywords no longer fires
- Guard rename_orphan_output_on_start against cross-workdir prev_run_id reuse via state.workdir comparison

Testing: pytest plugins/code/tools/python/test_run_loop_failure_marker.py; pytest plugins/code/tools/python/; bash -n plugins/code/scripts/run-loop.sh; ruff check .

Risks: Low; predicates are tightened (fewer false positives), real exhaustion / auth / context-limit / stderr-fallback paths covered by added scenarios A1-J27. PLN-500 canonical heartbeat-only fixture now returns {} rather than emitting loop-error.json.


shafty023:
Thanks for jumping on this. The change addresses the immediate heartbeat false positive by no longer treating allowed rate-limit posture events as terminal failures, and the added coverage around heartbeat-only streams is the right direction.

One limitation I’d call out before we merge this shape is that the classifier is still broader than the contract we want long-term. Some rate-limit detection still flows through generic error-text aggregation and broad “error-shaped” matching, which means assistant or tool-adjacent prose can still become classification input in cases that are not strict Claude CLI terminal-error envelopes. The fix is much safer than the merged FEA-932 behavior, but it still leaves a larger false-positive surface than necessary.

I’d suggest tightening the JSONL path so it classifies only positive failure evidence: a terminal result envelope marked as an error, a synthetic API-error envelope, or a rate-limit event whose status fields are explicitly non-allowed. Marker messages should always come from the specific triggering entry rather than any earlier text in the stream. Auth and context-limit detection should follow the same triggering-entry pattern so future prose about those topics cannot accidentally surface as user-visible runner failures.

There are also a few unrelated cleanup changes in this PR. They may be valid, but they make the terminal-failure fix harder to review and risk coupling this urgent regression fix to unrelated behavior. I’d prefer splitting those out or at least documenting why they need to ship with this PR.

Overall, this is a good incident fix direction; I’d just like us to narrow the classifier boundary further so this class of false positive does not come back in a different form.

aponamarev:
@shafty023 Good catch — you were right that the tightening was asymmetric. Pushed `873ea8d` on this branch which scopes `rate_limit_signal` and `context_limit_signal` to the same envelope-bound pattern that `auth_challenge_signal` already used.

**What changed in `detect_claude_terminal_failure` (run-loop.sh):**

- `rate_limit_signal`: dropped the `error_shaped and (text_blob | test(...))` branch. The keyword regex now only fires inside an `is_error: true` envelope (matching `.result`) or an `isApiErrorMessage: true` envelope (matching `.error`). The three positive-evidence branches — `rate_limit_event` with non-`allowed` status, `error: "rate_limit_error"`, and `apiErrorStatus: 429` — are unchanged.
- `context_limit_signal`: same treatment. Previously bare `error_shaped and (text_blob | test(...))`; now envelope-scoped to `.result` / `.error` only.

`text_blob` (which aggregated `.result`, `.message.content[].text`, `.message.content`, and `.error` per entry) is no longer reachable from any signal predicate. All marker messages were already sourced from the triggering entry — that part was unchanged.

**Concrete false-positive example now blocked:**

Before the patch, an assistant entry like this would classify as a terminal rate-limit failure and write `loop-error.json`:

```jsonl
{"type":"assistant","is_error":true,"result":"","message":{"content":[{"type":"text","text":"Note: you've hit your limit on test fixture cardinality."}]}}
```

The entry satisfied `error_shaped` (because `is_error: true`), and `text_blob` pulled "you've hit your limit" out of `.message.content[].text`. Pure prose — no `rate_limit_event`, no 429, no `error: "rate_limit_error"` — got runner-fataled.

After the patch the same entry returns `{}` because the rate-limit regex now only inspects `.result` (empty here), not `.message.content`. A symmetric case now blocked for context-limit:

```jsonl
{"type":"assistant","is_error":true,"result":"","message":{"content":[{"type":"text","text":"Discussing why prompt is too long is a common topic."}]}}
```

Both cases are pinned by new scenarios in `test_run_loop_failure_marker.py` (`test_rate_limit_prose_in_message_content_does_not_trigger`, `test_context_limit_prose_in_message_content_does_not_trigger`), with paired positive tests confirming that prose in `.result` / `.error` on a real error envelope still fires correctly. Full failure-marker suite: 52/52 green; broader `plugins/code/tools/python/` suite: 253/253 green.

Out of scope for this commit (happy to file follow-ups): removing the now-unused `user_texts` / `text_blob` / `first_user_text` / `first_error_text` helper defs.

PR #73, opened 2026-05-07
fix(self-learning): CLOSEDLOOP_ITERATION env var as fallback, not override

aponamarev:
## Summary
- Make `CLOSEDLOOP_ITERATION` a **fallback** in `evaluate_reduce_failures` instead of an unconditional override of the iteration count parsed from `runs.log`.
- Bumps `self-learning` to v1.2.4.

## What was broken

`evaluate_reduce_failures` scores the `reduce-failures` goal by reading how many iterations a given `run_id` took to complete. It produces two outputs that flow directly into the self-learning pipeline:

- `success: bool` — `iterations <= target`
- `score: float` — `1.0 - (iterations / (target * 2))`, clamped to `[0, 1]`

Before this fix, the function read the correct iteration count for `run_id` from `runs.log`, then **unconditionally overwrote it** with `os.environ['CLOSEDLOOP_ITERATION']` if that env var was set:

```python
# Old (buggy) precedence
if runs_log.exists():
    # ...parse iterations for this run_id from runs.log...

env_iterations = os.environ.get('CLOSEDLOOP_ITERATION')
if env_iterations and env_iterations.isdigit():
    iterations = int(env_iterations)   # always wins
```

`runs.log` is keyed by `run_id` (per-run, authoritative). `CLOSEDLOOP_ITERATION` is a process-global env var exported by the outer orchestrator loop. Whenever an evaluation ran in the same shell/process as a parent loop — which is the normal case in `run-loop.sh` — the **outer loop's current iteration counter clobbered the actual iteration count of the run being evaluated**.

## Why it has to be fixed

The mis-scoring is not cosmetic; it corrupts the inputs to self-learning:

1. **Wrong success/failure labels.** A run that finished in 2 iterations could be scored against the outer loop's counter (e.g. 10), flipping `success` from `true` to `false` and dropping `score` from ~0.67 to `0.0`. The reverse is also possible — a run that took 10 iterations can be marked successful if the env var happens to read low.
2. **Polluted learning store.** `success` and `score` are the labels that drive `self-learning`'s reinforcement and pruning of `org-patterns.toon`. Wrong labels mean **good patterns get penalized and bad patterns can get reinforced** — the exact opposite of what the system is supposed to do, and the kind of bug that compounds silently over many runs.
3. **Per-run scoring stops being per-run.** With env-var override, every evaluation in a given process collapses to the same iteration count, defeating the purpose of keying `runs.log` by `run_id`.
4. **Hard to detect from the outside.** The function still returns a well-formed `GoalOutcome`, so nothing crashes; the only symptom is gradually drifting learning quality.

## The fix

Consult `CLOSEDLOOP_ITERATION` only when `run_id` is **not** found in `runs.log` (e.g. ad-hoc evaluations without a logged run). When `runs.log` has an entry for the run, that value wins.

## Context

Split out of #72 (PLN-502). The PLN-502 PR is about false-positive terminal-failure detection from benign rate-limit heartbeats; this iteration-precedence fix was found during the same investigation but is an independent correctness bug in a different plugin. Keeping them in separate PRs so review and revert scopes match.

## Test plan
- [x] `pytest plugins/self-learning/tools/python/` — 75 passed
- [x] Existing tests cover `runs.log`-driven scoring when env var is unset; new behavior is that the env var is ignored when `runs.log` has the row.

### DIFF
--- a/plugins/self-learning/tools/python/evaluate_goal.py
+++ b/plugins/self-learning/tools/python/evaluate_goal.py
@@ -25,7 +25,9 @@ except ImportError:
 # runs.log is append-only:
 # run_id|timestamp|goal_name|iteration|status[|command|last_session_id]
 # reduce-failures only needs run_id and iteration, so legacy 4+ field rows and
-# newer session-correlated rows are both accepted.
+# newer session-correlated rows are both accepted. Because the loop reuses a
+# single run_id across iterations, multiple rows can share the same run_id;
+# reduce-failures uses the maximum iteration across all matching rows.
 RUNS_LOG_MIN_FIELDS = 4
 
 
@@ -54,15 +56,22 @@ def evaluate_reduce_failures(config: GoalConfig, run_id: str, workdir: Path) -> 
     target = config.success_criteria.get('target', 3)
     found_in_log = False
 
-    # Try to read actual iteration count from runs.log
+    # Try to read actual iteration count from runs.log. The loop reuses run_id
+    # across iterations and post_iteration_processing runs before the current
+    # iteration row is appended, so multiple rows can share the same run_id;
+    # take the maximum iteration across all matching valid rows.
     if runs_log.exists():
+        max_iteration = -1
         with open(runs_log, 'r') as f:
             for line in f:
                 parts = line.strip().split('|')
-                if len(parts) >= RUNS_LOG_MIN_FIELDS and parts[0] == run_id:
-                    iterations = int(parts[3]) if parts[3].isdigit() else 10
+                if len(parts) >= RUNS_LOG_MIN_FIELDS and parts[0] == run_id and parts[3].isdigit():
+                    row_iteration = int(parts[3])
+                    if row_iteration > max_iteration:
+                        max_iteration = row_iteration
                     found_in_log = True
-                    break
+        if found_in_log:
+            iterations = max_iteration
 
     # Fall back to environment variable only when run_id was not found in runs.log
     if not found_in_log:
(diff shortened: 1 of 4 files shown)


plugins/self-learning/tools/python/evaluate_goal.py
```diff
@@ -60,12 +61,14 @@ def evaluate_reduce_failures(config: GoalConfig, run_id: str, workdir: Path) ->
...
                     iterations = int(parts[3]) if parts[3].isdigit() else 10
+                    found_in_log = True
                     break
 
-    # Also check environment variable
-    env_iterations = os.environ.get('CLOSEDLOOP_ITERATION')
-    if env_iterations and env_iterations.isdigit():
-        iterations = int(env_iterations)
+    # Fall back to environment variable only when run_id was not found in runs.log
+    if not found_in_log:
```
shafty023, line 77: **[P2]** Keep the current iteration for repeated run rows

When self-learning evaluates any iteration after the first, `runs.log` can already contain an earlier row for the same `run_id` because `post_iteration_processing` runs before the loop appends the current `in_progress`/`completed` row, and the loop reuses the same run id. With this guard, the evaluator stops at that older row and refuses to use `CLOSEDLOOP_ITERATION`, so multi-iteration runs are scored using the first logged iteration (often `1`) instead of the current one. This will over-score `reduce-failures` goals for normal runs that take more than one iteration.

aponamarev: Fixed at commit b683e70. Replaced the first-match `break` scan with a max(iteration) accumulator across all matching rows in runs.log (lines 64-74). This ensures multi-iteration runs are scored against the latest iteration, not the earliest. Four regression tests added covering the multi-row case, env-var fallback, env-var suppression when log has run_id, and malformed-row skipping.

PR #74, opened 2026-05-07
PLN-511: fix(code): fail loop when max iterations reached with no progress

aponamarev:
## Summary

Fix a silent failure mode in `run-loop.sh` where reaching `max_iterations` with zero forward progress was reported as a successful completion. The loop now tracks per-run progress and surfaces a user-visible failure when the iteration budget is exhausted without producing any successful iterations.

## Changes

- **`plugins/code/scripts/run-loop.sh`** — track a `successful_iterations` counter inside `main()`, incrementing when an iteration produces a non-empty `result` or when a `COMPLETE` promise is detected in the stream. On `max_iterations` exit, emit `RUNNER_ERROR` / `MAX_ITERATIONS_NO_PROGRESS` and exit 4 if `successful_iterations == 0`.
- **`runs.log` format** — extend `write_runs_log_entry` with an optional 8th field (`successful_iterations`), appended only when provided. Existing 7-field readers continue to parse the line correctly.
- **`plugins/code/tools/python/test_run_loop_failure_marker.py`** — new tests covering the no-progress failure path and runs.log field emission.
- **`plugins/self-learning/tools/python/test_evaluate_goal.py`** — clear ambient `CLOSEDLOOP_ITERATION` in `test_reduce_failures_reads_runs_log_from_workdir_root` so the test is hermetic.
- **`plugins/code/.claude-plugin/plugin.json`** — bump `code` plugin to `1.11.12`.

## Testing

- `pytest plugins/code/tools/python/test_run_loop_failure_marker.py`
- `pytest plugins/self-learning/tools/python/test_evaluate_goal.py`

## Risks

None identified. The runs.log change is append-only and backwards-compatible. The new exit code 4 path only fires in a previously silent failure mode (max iterations with zero progress), so existing successful runs are unaffected.

---
Loop ID: 019e038b-14f8-726f-89e8-505d1b3da137
Artifact: https://app.closedloop.ai/implementation-plans/PLN-511

### DIFF
--- a/plugins/code/scripts/run-loop.sh
+++ b/plugins/code/scripts/run-loop.sh
@@ -614,9 +614,10 @@ record_claude_session_id() {
 # Write runs.log entry for goal evaluation and session correlation.
 #
 # Format:
-#   run_id|timestamp|goal|iteration|status|command|last_session_id
+#   run_id|timestamp|goal|iteration|status|command|last_session_id[|successful_iterations]
 # The first five fields are the legacy contract; append-only fields keep older
 # self-learning readers compatible while allowing command-scoped session lookup.
+# The optional 8th field (successful_iterations) is appended only when provided.
 write_runs_log_entry() {
   local workdir="$1"
   local iteration="$2"
@@ -635,6 +636,10 @@ write_runs_log_entry() {
   else
     session_id="${LAST_CLAUDE_SESSION_ID:-}"
   fi
+  local success_count=""
+  if [[ $# -ge 6 ]]; then
+    success_count="$6"
+  fi
   local runs_log="$workdir/runs.log"
   local timestamp
   timestamp=$(date -u +"%Y-%m-%dT%H:%M:%SZ")
@@ -646,7 +651,11 @@ write_runs_log_entry() {
   command=$(sanitize_runs_log_field "$command")
   session_id=$(sanitize_runs_log_field "$session_id")
   mkdir -p "$(dirname "$runs_log")"
-  echo "$RUN_ID|$timestamp|${CLOSEDLOOP_ACTIVE_GOAL:-reduce-failures}|$iteration|$status|$command|$session_id" >> "$runs_log"
+  local entry="$RUN_ID|$timestamp|${CLOSEDLOOP_ACTIVE_GOAL:-reduce-failures}|$iteration|$status|$command|$session_id"
+  if [[ -n "$success_count" ]]; then
+    entry="$entry|$success_count"
+  fi
+  echo "$entry" >> "$runs_log"
 }
 
 # Post-iteration processing: enrichment pipeline, learning capture, citation verification, success rates
@@ -1534,6 +1543,7 @@ main() {
   local final_result='select(.type == "result").result // empty'
 
   local consecutive_empty=0
+  local successful_iterations=0
 
   while true; do
     # Check max iterations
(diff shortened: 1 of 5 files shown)


plugins/code/scripts/run-loop.sh
```diff
@@ -1534,6 +1543,7 @@ main() {
   local final_result='select(.type == "result").result // empty'
 
   local consecutive_empty=0
+  local successful_iterations=0
```
shafty023: **[P2]** Persist successful iteration count across resumes

Because this counter is initialized from scratch inside `main()` and only the iteration number is persisted, a documented `run-loop.sh` resume loses successes from earlier invocations. If an earlier invocation produced a non-empty result, the process is restarted, and the resumed invocation then reaches the max-iterations boundary with only empty/error iterations, this branch will emit `MAX_ITERATIONS_NO_PROGRESS` and exit 4 even though the run already made forward progress. Persist or reconstruct the count when loading existing state.

plugins/code/scripts/run-loop.sh
```diff
@@ -1554,10 +1564,10 @@ main() {
       )"
 
       echo -e "\n${GREEN}Max iterations ($max_iterations) reached. Loop complete.${NC}"
-      log_progress "Loop ended - max iterations reached"
+      log_progress "Loop ended - max iterations reached (successful_iterations=$successful_iterations)"
 
-      # Write runs.log entry
-      write_runs_log_entry "$effective_workdir" "$iteration" "max_iterations" "plan_execute"
+      # Write runs.log entry (includes successful_iterations count)
+      write_runs_log_entry "$effective_workdir" "$iteration" "max_iterations" "plan_execute" "" "$successful_iterations"
```
shafty023: **[P2]** Preserve session id on max-iteration log rows

Passing an explicit empty fifth argument disables `write_runs_log_entry`'s fallback to `LAST_CLAUDE_SESSION_ID` or `session-id.txt`, so every max-iterations row that includes the new 8th field writes an empty `last_session_id` even after `plan_execute` produced a session. This breaks the existing runs.log session-correlation contract specifically for loops that stop by max iterations; resolve the session id before adding the success count or let the helper append the count without suppressing fallback.

plugins/code/scripts/run-loop.sh
aponamarev: Fixed in commit e635e03. `successful_iterations` is now read from the state file at line 1567-1571 (`get_field "successful_iterations"`, defaulting to 0 for legacy state files). The state file is initialized with `successful_iterations: 0` (line 1406) and a new `update_successful_iterations()` helper (lines 1345-1359) durably persists the counter after each increment (lines 1738-1739, 1765-1766). A resumed run that follows prior forward progress will no longer falsely trigger MAX_ITERATIONS_NO_PROGRESS.

plugins/code/scripts/run-loop.sh
aponamarev: Fixed in commit e635e03. The session id is now resolved at the call site into `session_id_for_log` (lines 1598-1601) before being passed as arg 5 to `write_runs_log_entry` (line 1604). The resolution mirrors the helper's own fallback chain: `LAST_CLAUDE_SESSION_ID`, then `session-id.txt`, then empty. The empty-string arg that was suppressing all fallbacks is gone.

PR #75, opened 2026-05-07
feat: add scripts/test-local-plugins.sh smoke-test harness

peterulsteen:
## Summary

- New `scripts/test-local-plugins.sh` smoke-test harness for plugin changes: auto-detects plugins changed vs `origin/main`, redirects the marketplace cache to local source via symlink (with sidecar-tracked revert), runs the plugin's own test suites, smoke-tests slash commands, and runs synthetic payloads against any pre/post-tool-use hooks following the closedloop perf-event contract.
- End-of-run banner lists every marketplace path still redirected plus the exact `--revert` command — the symlink intentionally persists past script exit so the user can restart Claude Code with local hooks loaded.
- `scripts/README.md` documents purpose, usage, exit codes, and out-of-scope (PRD/plan parsing belongs in a separate spec tool; multi-iteration loops are driven externally by `run-loop.sh`). Manually maintained — explicitly outside `/update-documentation`'s plugin-scoped scope. No per-script CHANGELOG; git history is the source of truth.

## Test plan

- [x] End-to-end run against an active branch with local plugin changes (`symphony/pln-466`, code v1.11.10/11): **17 pass / 0 fail / 0 skip**.
- [x] `--revert` restores caches cleanly; sidecars and backups removed.
- [x] `--plugin <typo>` exits 1 (hard fail), not 0 silently.
- [x] `--help` prints full docstring (awk-based extraction, resilient to docstring growth).
- [x] Synthetic payloads exercise all four hook code paths: `tool` event, `skill` via `tool_input.skill`, `skill` via `tool_input.command` fallback, `spawn` event for `Agent`, and `ok:false` for a failed `Read`.
- [x] Sentinels in `.tool-calls/` cleaned up after post-hook fires; assertion fails loudly if pre-hook never created the directory.
- [x] EXIT trap preserves `$TMP_BASE` on failure so `fail()`-referenced log paths remain inspectable.
- [x] `revert_cache` refuses to remove a symlink unless it knows what to restore — sidecars with unknown kind warn loudly instead of orphaning the cache.
- [x] Independent sub-agent verification confirmed detection, symlink resolution, sidecar tracking, hook emission, and revert.
- [x] Multi-agent code review (Bug Hunter A×2, Bug Hunter B, Auditor, Premise): 7 findings — 6 addressed, 1 false-positive (`CLOSEDLOOP_PERF_V2`) preempted with an inline comment explaining the gate was removed in `be6e758` so future readers don't re-add it.

### DIFF
--- a/scripts/test-local-plugins.sh
+++ b/scripts/test-local-plugins.sh
@@ -35,7 +35,7 @@
 #   1 — at least one check failed
 #   2 — environment problem (missing jq, python3, no installed plugins, etc.)
 
-set -uo pipefail
+set -euo pipefail
 
 # ──────────────────────────────────────────────────────────────────────────────
 # Config
@@ -92,13 +92,19 @@ usage() {
   exit 0
 }
 
+need_value() {
+  local flag="$1"
+  local value="${2:-}"
+  [ -n "$value" ] || die "$flag requires a value"
+}
+
 while [ $# -gt 0 ]; do
   case "$1" in
-    --plugin)  ONLY_PLUGIN="$2"; shift 2 ;;
-    --base)    BASE_BRANCH="$2"; shift 2 ;;
-    --owner)   OWNER="$2"; shift 2 ;;
+    --plugin)  need_value "$1" "${2:-}"; ONLY_PLUGIN="$2"; shift 2 ;;
+    --base)    need_value "$1" "${2:-}"; BASE_BRANCH="$2"; shift 2 ;;
+    --owner)   need_value "$1" "${2:-}"; OWNER="$2"; shift 2 ;;
     --revert)  REVERT_MODE=1; shift ;;
-    --repo-root) REPO_ROOT_OVERRIDE="$2"; shift 2 ;;
+    --repo-root) need_value "$1" "${2:-}"; REPO_ROOT_OVERRIDE="$2"; shift 2 ;;
     --verbose|-v) VERBOSE=1; shift ;;
     -h|--help) usage ;;
     *)         echo "Unknown arg: $1" >&2; exit 2 ;;
@@ -127,12 +133,15 @@ require_cmd python3
 require_cmd git
 
 [ -f "$INSTALLED_JSON" ] || die "No installed plugins file at $INSTALLED_JSON. Is the marketplace installed?"
+jq -e '.plugins and (.plugins | type == "object")' "$INSTALLED_JSON" >/dev/null \
+  || die "Installed plugins file is malformed: $INSTALLED_JSON"
 [ -d "$REPO_ROOT/plugins" ] || die "No plugins/ directory at $REPO_ROOT — run from the closedloop-ai monorepo root or pass through scripts/"
 
 mkdir -p "$TMP_BASE"
 # Preserve TMP_BASE on failure: fail() messages reference log files inside it,
 # and removing the dir before the user reads them defeats the report. Clean
 # up only on a fully green run.
+# shellcheck disable=SC2329 # Invoked indirectly by the EXIT trap.
 cleanup_tmp_base() {
   if [ "$FAIL_COUNT" -eq 0 ]; then
     rm -rf "$TMP_BASE"
@@ -141,6 +150,7 @@ cleanup_tmp_base() {
   fi
 }
 trap cleanup_tmp_base EXIT
+trap 'FAIL_COUNT=$((FAIL_COUNT+1)); exit 130' INT TERM
 
 # ──────────────────────────────────────────────────────────────────────────────
 # Core helpers
@@ -237,10 +247,11 @@ revert_cache() {
     if [ -L "$install_path" ]; then
       warn "[$plugin] symlink at $install_path has no sidecar — refusing to remove (would orphan the cache)"
       warn "[$plugin] inspect manually: ls -la $install_path"
+      return 1
     else
       info "[$plugin] no sidecar and no symlink — nothing to revert"
+      return 0
     fi
-    return 1
   fi
 
   local kind value
@@ -290,9 +301,17 @@ verify_symlink_target() {
   local plugin="$1" install_path="$2"
   local local_path="$REPO_ROOT/plugins/$plugin"
   local actual_target
-  actual_target=$(readlink "$install_path" 2>/dev/null || echo "")
+  if [ ! -e "$install_path" ] && [ ! -L "$install_path" ]; then
+    fail "[$plugin] symlink path $install_path does not exist"
+    return 1
+  fi
+  if [ ! -L "$install_path" ]; then
+    fail "[$plugin] expected $install_path to be a symlink"
+    return 1
+  fi
+  actual_target=$(readlink "$install_path")
   if [ "$actual_target" != "$local_path" ]; then
-    fail "[$plugin] symlink at $install_path does not point to $local_path (points to: ${actual_target:-<not a symlink>})"
+    fail "[$plugin] symlink at $install_path does not point to $local_path (points to: ${actual_target:-<empty target>})"
     return 1
   fi
   local local_pj="$local_path/.claude-plugin/plugin.json"
@@ -333,6 +352,10 @@ run_plugin_tests() {
       pys+=("$f")
     done < <(find "$pdir/tools/python" -maxdepth 1 -name 'test_*.py' -print0 2>/dev/null)
     if [ "${#pys[@]}" -gt 0 ]; then
+      if ! python3 -m pytest --version >/dev/null 2>&1; then
+        skip "[$plugin] pytest tools/python (${#pys[@]} file(s)) — pytest is not installed"
+        return
+      fi
       if (cd "$REPO_ROOT" && python3 -m pytest -q "$pdir/tools/python" \
             >"$TMP_BASE/$plugin-pytest.log" 2>&1); then
         pass "[$plugin] pytest tools/python (${#pys[@]} file(s))"
@@ -360,7 +383,7 @@ classify_command() {
   fi
   # Orchestrator: explicit orchestrator role markers, agent/subagent delegation,
   # or Task( invocations in the command body.
-  if grep -qiE '\*\*you are the orchestrator\*\*|^You orchestrate|^# .*[Oo]rchestrator|delegate to|subagent|Task\(|SendMessage' "$cmd_file" 2>/dev/null; then
+  if grep -qE '\*\*[Yy]ou are the orchestrator\*\*|^[Yy]ou orchestrate|^# .*[Oo]rchestrator|[Dd]elegate to|Task\(|SendMessage' "$cmd_file" 2>/dev/null; then
     echo orchestrator
     return
   fi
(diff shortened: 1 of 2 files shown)


scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,682 @@
...
+    bootstrap)
+      local bootstrap_line
+      # Extract the inside of !`...` from the first matching line.
+      bootstrap_line=$(grep -E '^\s*!`\s*bash' "$cmd_file" | head -1 | sed 's/^[^`]*`[[:space:]]*//; s/`.*$//')
+      if [ -z "$bootstrap_line" ]; then
+        info "[$plugin] /$plugin:$cmd_name — bootstrap line not parseable, skipping run"
+        return
+      fi
+      local resolved
+      resolved=$(echo "$bootstrap_line" | sed "s|\${CLAUDE_PLUGIN_ROOT}|$install_path|g; s|\$CLAUDE_PLUGIN_ROOT|$install_path|g")
```
thadeusb: The sed here expands `$CLAUDE_PLUGIN_ROOT` into a literal path string, but then line 405 passes the result to `eval` with `CLAUDE_PLUGIN_ROOT="$install_path"` already in the environment. So eval would expand the variable naturally — the sed is doing work that eval is about to redo. The problem is the sed bakes `$install_path` into the string as a literal, and then eval re-parses it. If `install_path` ever has a space (weird home dir, whatever), the literal substitution breaks at the eval layer because it's no longer quoted. Drop the sed, keep the `$ARGUMENTS` strip, and let the env var on line 405 do its job.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,682 @@
...
+    done <"$hwd/work/perf.jsonl"
+
+    # Verify the fallback skill path actually fires (skill_name extracted from
+    # tool_input.command when tool_input.skill is absent).
+    if [ -f "$post" ]; then
+      if jq -e 'select(.event=="skill" and .skill_name=="plugin:fallback-skill")' \
+            "$hwd/work/perf.jsonl" >/dev/null 2>&1; then
+        pass "[$plugin] post-hook skill_name fallback (tool_input.command) works"
+      else
+        warn "[$plugin] post-hook skill_name fallback (tool_input.command) not exercised in output"
```
thadeusb: This uses `warn` (no FAIL_COUNT increment) but the README and PR description both call this an assertion ("Asserts the fallback skill_name path fired"). If the `tool_input.command` fallback is broken, the run exits 0. Should this be `fail` instead of `warn`?

mikeangstadt (commented):
## Review: `scripts/test-local-plugins.sh`

This is a well-structured, substantial addition (682 lines of bash + README). The script solves a real problem — testing local plugin changes against marketplace installations — and handles the tricky state management (symlink redirect + revert) carefully.

### Strengths

- **Sidecar-based revert** (`dir|` / `link|` + unknown-kind guard) is resilient and handles the prior-symlink edge case properly.
- **Synthetic hook payloads** cover all four event types (`tool`, `skill`/`skill_field`, `skill`/`command_field` fallback, `spawn`, failed `Read`) — good threat model.
- **`$TMP_BASE` preserved on failure** so log paths referenced in `fail()` messages remain inspectable.
- `detect_changed_plugins` uses `git merge-base` (three-dot diff) which is correct, and prefers `origin/<base>` over local.

### Issues worth addressing

1. **No `set -e`**: `set -uo pipefail` without `errexit` means a silent failure in an intermediate pipeline stage won't stop the script. Some calls (e.g., `jq` in `lookup_installed` with a malformed `installed_plugins.json`) could produce empty results that cascade into confusing `skip`/`fail` outcomes instead of a clear `die`. Consider adding `set -o errexit` or explicitly `|| die` on critical pipeline calls.

2. **Potential `readlink` issues on non-macOS**: `verify_symlink_target` uses `readlink "$install_path"`. On Linux, `readlink` without `-f` returns the raw target of a symlink (which is correct here), but on some platforms `readlink` on a non-symlink exits with a non-zero code silently. The fallback `echo ""` covers the error case, but the error message would say `<not a symlink>` even when the path doesn't exist — could be confusing during debugging.

3. **`${ARGUMENTS}` brace-form not stripped**: In `smoke_test_command`, only `$ARGUMENTS` (bare) is stripped via `"${resolved//\$ARGUMENTS/}"`. If a bootstrap command uses `${ARGUMENTS}` (braced form), it will be left in the resolved string and likely produce a shell error when evaluated. This should either strip both forms or note the limitation in the skippable-bootstrap info message.

4. **`cleanup_tmp_base` on EXIT during Ctrl+C**: If the user presses Ctrl+C mid-run, `EXIT` trap fires with `FAIL_COUNT` possibly still 0 (no explicit failures yet), and the trap would `rm -rf "$TMP_BASE"` — deleting logs that might be useful for understanding what was interrupted. Consider using `SIGINT` trap to force `FAIL_COUNT=1` before exit, or just never auto-delete the tmp dir (leave it to `--revert` or a `--clean` flag).

5. **`python3 -m pytest` may not be installed**: `run_plugin_tests` calls pytest without checking. If a plugin has Python tests but pytest isn't in the venv, the error message in the log file will say `ModuleNotFoundError`, which might not be obvious to the user. Consider a `python3 -m pytest --version` guard before running tests, with a `skip` instead of `fail` if pytest is unavailable.

6. **No pip install step before pytest**: The function runs `python3 -m pytest` against `tools/python/` without installing `requirements.txt` first. Tests may hard-fail on missing dependencies even when the code is correct. Consider doing `pip install -r "$pdir/tools/python/requirements.txt"` (if it exists) as a precondition, or document that the caller should have deps installed.

7. **`CLOSEDLOOP_PERF_V2` not unset defensively**: The comment says "Do NOT export" but doesn't `unset CLOSEDLOOP_PERF_V2`. If the caller's environment has this set (e.g., from a prior session), it would be inherited by the synthetic hook subprocesses. Add `unset CLOSEDLOOP_PERF_V2` after the comment block to be safe.

### Nits

- `classify_command` uses case-insensitive grep for orchestrator markers like `"orchestrator"` — could match comments that happen to contain the word. Consider anchoring to YAML frontmatter keys or markdown headings.
- The `# NOTE:` comment re: `be6e758` is helpful but will go stale as the commit scrolls off `HEAD~N`. Consider referencing a permanent file path or a date-anchored note instead of a commit hash that future readers may not have in their local history.

> Auto-reviewed by opencode headless

scripts/test-local-plugins.sh
peterulsteen: Addressed in 9403e90: the bootstrap smoke test no longer substitutes CLAUDE_PLUGIN_ROOT into the command string before eval. It now leaves the plugin-root variable for the eval environment and only strips $ARGUMENTS / ${ARGUMENTS} placeholders.

scripts/test-local-plugins.sh
peterulsteen: Addressed in 9403e90: this is now a fail instead of a warn, so a broken tool_input.command fallback increments FAIL_COUNT and exits non-zero. The temp-HOME harness run also exercised this path successfully.

peterulsteen:
Addressed the summary review in 9403e90:

- enabled `errexit` and added explicit installed-plugin JSON validation
- tightened symlink/readlink diagnostics
- stripped both `$ARGUMENTS` and `${ARGUMENTS}` without pre-expanding `CLAUDE_PLUGIN_ROOT`
- preserved temp logs on interrupt
- added a pytest availability guard and README dependency note
- unset `CLOSEDLOOP_PERF_V2` defensively
- made the fallback skill-name assertion a hard failure
- narrowed command classification and removed the stale commit-hash note

I also rebased the branch onto current `origin/main` and ran `shellcheck`, `bash -n`, `git diff --check`, and a temp-HOME harness pass for `code` (17 pass, 0 fail, 1 expected pytest skip because pytest is not installed locally), plus `--revert`, missing-plugin, and malformed-JSON checks.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,705 @@
...
+  local entry; entry=$(lookup_installed "$plugin")
+  [ -n "$entry" ] || { skip "[$plugin] not installed; nothing to revert"; return; }
+  revert_cache "$plugin" "${entry%%|*}" || fail "[$plugin] revert did not complete"
+}
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Entry point
+# ──────────────────────────────────────────────────────────────────────────────
+
+# Pick plugin list
```
shafty023, line 636: `--revert` is using the same changed-plugin detection as test mode, so a bare `scripts/test-local-plugins.sh --revert` only restores plugins currently changed vs the selected base. If the prior run used `--plugin code`, the user rebases/switches branches, or there is no current `plugins/*` diff, lines 646-648 exit 0 while sidecar-marked cache redirects remain active. Revert mode with no `--plugin` should enumerate the `.test-local-plugins.prev` sidecars under the marketplace cache and restore those install paths directly.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,705 @@
...
+  git diff --name-only "$merge_base"...HEAD -- 'plugins/*' 2>/dev/null \
+    | awk -F/ 'NF>=2 && $1=="plugins" {print $2}' \
+    | sort -u
+}
+
+# Look up the installed cache directory for a plugin from installed_plugins.json.
+# Echoes "<installPath>|<version>" or empty if not installed.
+lookup_installed() {
+  local plugin="$1"
+  jq -r --arg key "${plugin}@${OWNER}" '
```
shafty023, line 187: `installed_plugins.json` can contain multiple entries per plugin, and the repo's installer treats `scope == "user"` as the canonical marketplace install. Taking `[0]` can pick a project-scoped or stale entry and redirect a cache path that a fresh Claude Code session will not use. Before mutating any returned `installPath`, also validate it resolves under the expected `~/.claude/plugins/cache/<owner>/<plugin>/...` root. Select a user-scoped entry with an existing install path, or fail clearly.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,705 @@
...
+      post_payload=$(echo "$payload" | jq -c '. + {tool_response:{error:"boom",success:false}}')
+    else
+      post_payload=$(echo "$payload" | jq -c '. + {tool_response:{success:true}}')
+    fi
+    if [ -f "$post" ]; then
+      echo "$post_payload" | bash "$post" || true
+    fi
+  done
+
+  if [ -s "$hwd/work/perf.jsonl" ]; then
```
shafty023, line 538: This pass condition is too weak for the synthetic hook contract. Any non-empty `perf.jsonl` plus one fallback skill row passes, even if the hook stopped emitting `tool` events, missed required fields, skipped `spawn`, or reported the failed `Read` as `ok: true`. Assert the expected taxonomy for these synthetic cases: 5 `tool` events, 2 `skill` events, 1 `spawn` event, required fields per event type, the `tool_input.command` fallback, `Agent.subagent_type`, and `ok:false` for the failed `Read`.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,705 @@
...
+    else
+      skip "[$plugin] not installed via marketplace ($plugin@$OWNER) — symlink not applicable"
+    fi
+    return
+  fi
+  local install_path="${entry%%|*}"
+  local installed_ver="${entry##*|}"
+  info "[$plugin] installed v$installed_ver at $install_path"
+
+  redirect_cache_to_local "$plugin" "$install_path" || return 0
```
shafty023, line 604: Returning success here hides redirect failures that did not call `fail()`. For example, `redirect_cache_to_local` only warns when a backup already exists without a sidecar, then returns 1; this `|| return 0` skips the plugin while leaving `FAIL_COUNT` unchanged, so the final run can still be green after refusing to redirect. Treat redirect/verification failures as failed checks instead of returning success from `test_plugin`.

scripts/test-local-plugins.sh
```diff
@@ -0,0 +1,705 @@
...
+      local wd="$TMP_BASE/wd-$plugin-$cmd_name"
+      mkdir -p "$wd"
+      local rc=0
+      (cd "$wd" && CLAUDE_PLUGIN_ROOT="$install_path" eval "$resolved" \
+            >"$TMP_BASE/$plugin-$cmd_name.log" 2>&1) || rc=$?
+      if [ "$rc" = "0" ]; then
+        pass "[$plugin] /$plugin:$cmd_name — bootstrap script ran cleanly"
+        debug "log: $TMP_BASE/$plugin-$cmd_name.log"
+      else
+        # Some bootstraps require args; treat exit as informational unless --verbose.
```
shafty023, line 436: Every bootstrap nonzero exit is currently downgraded to an informational 'may need args' case. In this repo the only bootstrap command is `/code:code`, and `setup-closedloop.sh` accepts a default workdir, so a real setup regression would exit nonzero and the harness would still finish green. Fail bootstrap exits unless the command is explicitly known to require unsupplied arguments.

PR #78, opened 2026-05-08
fix(code): add untrusted artifact prompt boundaries

mikeangstadt:
## Summary
- teach the code orchestrator, amend-plan, key planning agents, and judge context preparation to treat PRDs, plans, notes, and logs as untrusted evidence rather than instructions
- add the same adversarial-artifact rule to the shared judge input preamble so all judge runs inherit it
- add a regression test that guards the required trust-boundary language in those prompt files

## Testing
- `pytest plugins/code/tools/python/test_untrusted_artifact_boundary.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### code v1.5.10
+
+#### Security
+- Added explicit untrusted-artifact boundaries to the main code orchestrator prompt, `amend-plan`, and key planning agents (`pre-explorer`, `plan-draft-writer`, `plan-evaluator`, `plan-validator`) so PRDs, plans, notes, and attachments are treated as evidence rather than instructions
+
+### judges v1.4.1
+
+#### Security
+- Added shared untrusted-artifact handling guidance to judge context preparation and the common judge input preamble so PRDs, plans, diffs, and logs are evaluated as evidence instead of prompt authority
+
 ### code v1.5.9
 
 #### Fixed
--- a/plugins/code/tools/python/test_untrusted_artifact_boundary.py
+++ b/plugins/code/tools/python/test_untrusted_artifact_boundary.py
@@ -1,0 +1,33 @@
+"""Regression tests for untrusted-artifact prompt boundaries."""
+
+from pathlib import Path
+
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
```
thadeusb, line 7: CHANGELOG says code v1.5.10 and judges v1.4.1, but neither `plugins/code/.claude-plugin/plugin.json` nor `plugins/judges/.claude-plugin/plugin.json` were modified in this PR. The versions on the branch are still 1.5.9 and 1.4.0 respectively.

plugins/code/tools/python/test_untrusted_artifact_boundary.py
```diff
@@ -0,0 +1,33 @@
+"""Regression tests for untrusted-artifact prompt boundaries."""
+
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parents[4]
+
+FILES_REQUIRING_GUARDRAIL = [
```
thadeusb, line 8: `implementation-subagent` reads plan.json and has `Write, Edit, Bash` — it's the agent that actually executes instructions from the plan. Same story with `plan-writer` and `plan-importer` (which ingests external markdown). These feel like higher-risk injection surfaces than the read-only evaluators you did cover. Was the scope here intentionally limited to planning/evaluation, with implementation agents coming later?

plugins/code/tools/python/test_untrusted_artifact_boundary.py
```diff
@@ -0,0 +1,33 @@
...
+    / "skills"
+    / "artifact-type-tailored-context"
+    / "preambles"
+    / "common_input_preamble.md",
+]
+
+
+def test_prompt_boundaries_treat_artifacts_as_untrusted_data() -> None:
+    for path in FILES_REQUIRING_GUARDRAIL:
+        content = path.read_text()
```
thadeusb, line 28: This guards the files you know about today but won't catch a new agent added next month that reads artifacts without the boundary. Any appetite for a second test that globs `plugins/*/agents/*.md`, checks which ones have tool access to `Read` or `Bash`, and asserts they're either in this list or explicitly excluded?

PR #79, opened 2026-05-08
PLN-530: Restrict rate-limit detection to "rejected" status

aponamarev:
## Summary

Narrows the `rate_limit_signal` predicate in `plugins/code/scripts/run-loop.sh`'s `detect_claude_terminal_failure` from a denylist (`status != "allowed"`) to a single explicit fatal status (`status == "rejected"`). Benign Claude `rate_limit_event` heartbeats with `allowed_warning`, `paused`, `throttled`, and informational `exceeded` statuses no longer abort the loop with a false `CLAUDE_RATE_LIMIT` marker. The overage branch is gated identically: `overageStatus == "rejected"` with `isUsingOverage == true`. Status_429 and error-string match paths are unchanged.

## Changes

- `plugins/code/scripts/run-loop.sh`: `rate_limit_signal` jq predicate inverted from `(status != "allowed") or (overage path with overageStatus != "allowed")` to `(status == "rejected") or (overage path with overageStatus == "rejected")`.
- `plugins/code/.claude-plugin/plugin.json`: bump `code` to `v1.11.14`.
- `plugins/code/tools/python/test_run_loop_failure_marker.py`:
  - Update existing `test_rate_limit_event_predicate` rows so previously-fatal benign statuses (`allowed_warning` and friends) now expect `None` instead of `"CLAUDE_RATE_LIMIT"`.
  - Add RL-18..RL-31: `allowed_warning` with/without `warning_pct_int`, the `rejected`-only fatal path, overage-branch regression guards (`allowed`/`exceeded` with `isUsingOverage=true`, `rejected` with `isUsingOverage=false`), and cross-branch interactions where one of `status`/`overageStatus` is `rejected` and the other is not.
  - Add Group E `test_rate_limit_info_malformed` parametrized over RL-32..RL-35 (empty object, integer status, boolean status, string-instead-of-object) — exercises jq type guards.
  - Add Group G `test_rl_group_g_allowed_warning_e2e`: feeds a complete realistic Claude API JSONL stream (system init, benign heartbeat, `allowed_warning` with `warning_pct_int=80`, assistant tool call, tool result, second heartbeat, completion-promise turn, success result) through the full pipeline and asserts no signal fires.

## Why

PLN-530 tracks loop runs that aborted prematurely on benign rate-limit advisories. Anthropic emits `rate_limit_event` heartbeats throughout normal API usage to surface usage telemetry (`allowed_warning` with `warning_pct_int`, `paused` during transient throttles, etc.). Treating any non-`allowed` status as fatal turned every advisory into a terminal failure. Switching to a positive equality on `"rejected"` aligns the predicate with the actual API contract (rejection is the only status that means the request was not served) and is robust against new informational statuses that Anthropic may introduce.

## Testing

- `pytest plugins/code/tools/python/test_run_loop_failure_marker.py` — parametrized rate-limit predicate, malformed-payload, and end-to-end JSONL stream tests pass against the new predicate.

## Risks

Low. Tightens a fatal-failure predicate; net effect is fewer false positives. Genuine `rejected` events continue to produce `CLAUDE_RATE_LIMIT` markers exactly as before. Status_429 and error-string match branches are untouched, preserving fallback detection when `rate_limit_event` rows are absent.

---
Loop ID: 019e08de-7506-75d9-bcfb-2a6b38afa9a9
Artifact: https://app.closedloop.ai/implementation-plans/PLN-530

### DIFF
--- a/plugins/code/scripts/run-loop.sh
+++ b/plugins/code/scripts/run-loop.sh
@@ -259,8 +259,8 @@ detect_claude_terminal_failure() {
         ((.type? == "rate_limit_event") and (
           (.rate_limit_info? | type) == "object"
           and (
-            ((.rate_limit_info.status? // null) != null and (.rate_limit_info.status? != "allowed"))
-            or ((.rate_limit_info.isUsingOverage? == true) and ((.rate_limit_info.overageStatus? // null) != null and (.rate_limit_info.overageStatus? != "allowed")))
+            (.rate_limit_info.status? == "rejected")
+            or ((.rate_limit_info.isUsingOverage? == true) and (.rate_limit_info.overageStatus? == "rejected"))
           )
         ))
         or (error_string | ascii_downcase | test("^rate_limit(_error)?$"))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code v1.11.14
+
+#### Fixed
+- `rate_limit_signal` in `run-loop.sh`'s `detect_claude_terminal_failure` now fires only when `rate_limit_info.status == "rejected"` (or `overageStatus == "rejected"` with `isUsingOverage == true`), replacing the prior "any non-`allowed` value" denylist. Benign heartbeats with `status` of `allowed_warning`, `paused`, `throttled`, or informational `exceeded` no longer abort the loop. The `status_429` and error-string match paths remain unchanged, so genuine rate-limit failures continue to be marked. (PLN-530)
+
+#### Changed
+- Expanded `test_rate_limit_event_predicate` parametrization with RL-18..RL-31 covering `allowed_warning`, the rejected-only fatal path, overage-branch regression guards, and cross-branch interactions. Pre-existing rows for `paused`, `throttled`, `exceeded` (with overage on), and bare `rejected` (with `isUsingOverage` false) flip from `CLAUDE_RATE_LIMIT` to no-signal to encode the new gating. Adds Group E (RL-32..RL-35) malformed-payload coverage exercising jq's string-equality and object type guards, plus a Group G end-to-end test feeding a realistic Claude JSONL stream with `allowed_warning` heartbeats and asserting no signal fires.
+
 ### code v1.11.13
 
 #### Fixed
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.15",
+  "version": "1.11.14",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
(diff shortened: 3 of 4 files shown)


plugins/code/scripts/run-loop.sh
```diff
@@ -259,8 +259,8 @@ detect_claude_terminal_failure() {
         ((.type? == "rate_limit_event") and (
           (.rate_limit_info? | type) == "object"
           and (
-            ((.rate_limit_info.status? // null) != null and (.rate_limit_info.status? != "allowed"))
-            or ((.rate_limit_info.isUsingOverage? == true) and ((.rate_limit_info.overageStatus? // null) != null and (.rate_limit_info.overageStatus? != "allowed")))
+            (.rate_limit_info.status? == "rejected")
+            or ((.rate_limit_info.isUsingOverage? == true) and (.rate_limit_info.overageStatus? == "rejected"))
```
peterulsteen, line 263: - [P2] Preserve detection of terminal non-rejected rate states — claude-plugins/plugins/code/scripts/run-loop.sh:262-263
    When Claude emits a rate_limit_event with status values such as exceeded, paused, or throttled, or overageStatus: "exceeded" while isUsingOverage is true, this predicate now ignores the event because it only matches "rejected". These states were previously treated as terminal rate-limit signals, so affected runs fall through to the generic nonzero-exit path and can continue retrying without writing the user-visible rate-limit marker.

peterulsteen (changes requested):

plugins/code/scripts/run-loop.sh
aponamarev: Thanks for raising this — I dug in but I think the premise doesn't match the documented contract, so I'd push back on changing the predicate.

**The cited statuses aren't in the documented enum.** Anthropic's `rate_limit_info.status` is documented as `allowed | allowed_warning | rejected` (Claude Agent SDK schema mirrors the stream-json contract). `exceeded`, `paused`, and `throttled` aren't documented `status` values — they were only being treated as "terminal" as a side effect of the prior `status != "allowed"` denylist, not because the API contract says they are.

**That denylist behavior is exactly the bug PLN-530 fixes.** Anthropic emits `rate_limit_event` heartbeats throughout normal API usage to surface telemetry (`allowed_warning` with `warning_pct_int`, etc.). Under the old predicate every benign heartbeat aborted the loop with a false `CLAUDE_RATE_LIMIT` marker, which is what PLN-530 was opened to address.

**The scenarios you raised are explicitly tested as non-terminal** in `plugins/code/tools/python/test_run_loop_failure_marker.py`:

| Test ID | status | overageStatus | isUsingOverage | Expected |
|---|---|---|---|---|
| `RL-05-status-exceeded` | `exceeded` | `allowed` | False | None |
| `RL-06-status-paused` | `paused` | `allowed` | False | None |
| `RL-07-status-throttled` | `throttled` | `allowed` | False | None |
| `RL-26-overage-exceeded-on` | `allowed` | `exceeded` | **True** | **None** |

`RL-26` is the exact scenario in your comment (`overageStatus: "exceeded"` with `isUsingOverage: true`) and the test asserts it must not fire. The genuine fatal path is covered by `RL-20/23/28-30` (`rejected` with/without overage), which still produce `CLAUDE_RATE_LIMIT`.

**Fallback detection still holds for any unknown terminal state.** `status_429` (HTTP 429 from `api_error_status` / `apiErrorStatus`) and the `envelope_text_match` regex on `is_error` / `isApiErrorMessage` envelopes (`"you've hit your limit"`, `"usage limit"`, `"rate limit reached"`, etc.) are unchanged. A genuine rate-limit rejection would carry one of those signals, so even if Anthropic introduces a new terminal `status` outside the documented enum, the loop would still mark and exit.

Recommendation: leave the predicate as-is. Happy to add a code comment pointing to the documented enum if that helps future readers.

PR #82, opened 2026-05-11
PLN-539: Add slash command telemetry instrumentation

peterulsteen:
## Summary

Adds matched start/complete telemetry instrumentation to every public slash command across `bootstrap`, `code`, `code-review`, and `self-learning`, so each command run emits structured events to `perf.jsonl` with `run_id`, `command`, `duration`, and `exit_status` — enabling Datadog rollups per slash command.

## Changes

- **New per-plugin scripts** (in `plugins/<plugin>/scripts/`):
  - `command-telemetry-init.sh` — sourceable initialiser that exports `CLOSEDLOOP_RUN_ID`, `CLOSEDLOOP_COMMAND`, `CLOSEDLOOP_WORKDIR`, `CLOSEDLOOP_START_TIME`, persists a `.cmd-state.env` so a later Bash invocation can recover context, and triggers `record_run.sh`.
  - `command-telemetry-complete.sh` — finaliser that emits a `command_complete` JSONL event and cleans up the recovered state.
  - `record_run.sh` — added to `bootstrap`, `code-review`, and `self-learning` (already present in `code`).
- **`plugins/code/scripts/telemetry-helpers.sh`** — new sourceable module that lifts `emit_perf_event`, `run_timed_step`, and `emit_skipped_step` out of `run-loop.sh` (net `-77 / +2` in `run-loop.sh`). The helpers now resolve `CLOSEDLOOP_RUN_ID` first, falling back to `RUN_ID` for backward-compat when sourced inside the loop.
- **Slash command instrumentation** — added init/complete bookends to:
  - `bootstrap`: `agent-bootstrap`
  - `code`: `amend-plan`, `cancel-code`, `plan-with-codex` (plus `allowed-tools` updates where required)
  - `code-review`: `start`
  - `self-learning`: `export-closedloop-learnings`, `goal-stats`, `process-learnings`, `prune-learnings`, `pull-learnings`, `push-learnings`
- **Tests** under `plugins/code/scripts/tests/`:
  - `test_telemetry_helpers.sh` — covers the extracted helpers with an isolated env-var-only context.
  - `test_command_telemetry_init.sh` and `test_command_telemetry_complete.sh` — cover argument precedence, run-id format, fail-open behaviour, and state-file recovery across separate shells.
  - `plugins/code/tools/python/test_run_loop_failure_marker.py` — regression guard that `unset`s `CLOSEDLOOP_COMMAND` before exercising `write_runs_log_entry`.
- **Version bumps**:
  - `bootstrap` 1.2.0 → 1.2.2
  - `code` 1.11.16 → 1.14.5
  - `code-review` 1.5.5 → 1.5.7
  - `self-learning` 1.2.5 → 1.2.7

## Test plan

- [x] `bash plugins/code/scripts/tests/test_telemetry_helpers.sh`
- [x] `bash plugins/code/scripts/tests/test_command_telemetry_init.sh`
- [x] `bash plugins/code/scripts/tests/test_command_telemetry_complete.sh`
- [x] `pytest plugins/code/tools/python/test_run_loop_failure_marker.py`
- [ ] Run a slash command end-to-end and confirm a matched `command_complete` event lands in `perf.jsonl`.

## Risks

None identified. Both telemetry scripts fail open (init wraps everything in a function that always returns 0; complete uses `trap 'exit 0' ERR`), so a broken telemetry path cannot abort the calling command. `run-loop.sh` behaviour is preserved because the extracted helpers keep the `RUN_ID` fallback.

---
Loop ID: 019e18b5-49c4-75bd-88fe-47daf14fd36c
Artifact: https://app.closedloop.ai/implementation-plans/PLN-539

### DIFF
--- a/plugins/bootstrap/.claude-plugin/plugin.json
+++ b/plugins/bootstrap/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "bootstrap",
   "description": "Bootstrap plugin for ClosedLoop",
-  "version": "1.2.0",
+  "version": "1.2.2",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/bootstrap/scripts/command-telemetry-complete.sh
+++ b/plugins/bootstrap/scripts/command-telemetry-complete.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+# command-telemetry-complete.sh — Telemetry finaliser for slash commands.
+#
+# Designed to be called (not sourced) by command markdown files at the end of
+# a command run to emit a `command_complete` event to perf.jsonl.
+#
+# Usage:
+#   bash "${CLAUDE_PLUGIN_ROOT}/scripts/command-telemetry-complete.sh" [exit_status]
+#
+# Arguments:
+#   $1  EXIT_STATUS (optional, default: 0) — the exit status of the command
+#
+# Environment variables consumed (set by command-telemetry-init.sh):
+#   CLOSEDLOOP_START_TIME  — unix epoch when the command started (required for duration)
+#   CLOSEDLOOP_RUN_ID      — unique run identifier
+#   CLOSEDLOOP_WORKDIR     — working directory for perf.jsonl output
+#   CLOSEDLOOP_COMMAND     — slash command name
+#
+# Fail-open guarantee: the entire finalisation is wrapped in a trap so any
+# error is suppressed and the calling command continues unaffected.
+
+# Fail open: any unexpected error exits 0 so the calling command is unaffected.
+trap 'exit 0' ERR
+
+_closedloop_telemetry_complete() {
+  local exit_status="${1:-0}"
+
+  # Resolve working directory. The matching command-telemetry-init.sh exports
+  # CLOSEDLOOP_WORKDIR, but those exports are scoped to the shell that ran
+  # `source` and do not survive across separate Bash tool invocations. When
+  # the env var is unset, fall back to the default workdir init.sh uses
+  # (.closedloop-ai/telemetry) only if a state file exists there — otherwise
+  # the command was never initialised and there is nothing to finalise.
+  local workdir="${CLOSEDLOOP_WORKDIR:-}"
+  if [[ -z "$workdir" ]]; then
+    if [[ -f ".closedloop-ai/telemetry/.cmd-state.env" ]]; then
+      workdir=".closedloop-ai/telemetry"
+    else
+      return 0
+    fi
+  fi
+
+  # If RUN_ID is missing, recover the run context that init.sh persisted to
+  # disk. This bridges the gap between separate Bash tool invocations: init
+  # exported env vars but they died with its source shell.
+  local state_file="$workdir/.cmd-state.env"
+  if [[ -z "${CLOSEDLOOP_RUN_ID:-}" && -f "$state_file" ]]; then
+    # shellcheck source=/dev/null
+    source "$state_file" 2>/dev/null || true
+    workdir="${CLOSEDLOOP_WORKDIR:-$workdir}"
+  fi
+
+  local perf_file="$workdir/perf.jsonl"
+
+  # Capture end time
+  local end_epoch
+  end_epoch=$(date +%s 2>/dev/null) || end_epoch=""
+  local ended_at
+  ended_at=$(date -u +%Y-%m-%dT%H:%M:%SZ 2>/dev/null) || ended_at=""
+
+  # Resolve start time and calculate duration
+  local start_epoch="${CLOSEDLOOP_START_TIME:-}"
+  local started_at=""
+  local duration_s=0
+
+  if [[ -n "$start_epoch" && -n "$end_epoch" ]]; then
+    duration_s=$(( end_epoch - start_epoch ))
+    # Convert start epoch to ISO-8601 timestamp
+    started_at=$(date -u -r "$start_epoch" +%Y-%m-%dT%H:%M:%SZ 2>/dev/null) \
+      || started_at=$(date -u -d "@$start_epoch" +%Y-%m-%dT%H:%M:%SZ 2>/dev/null) \
+      || started_at=""
+  fi
+
+  local run_id="${CLOSEDLOOP_RUN_ID:-unknown}"
+  local command="${CLOSEDLOOP_COMMAND:-interactive}"
+
+  # Ensure perf.jsonl directory exists
+  mkdir -p "$(dirname "$perf_file")" 2>/dev/null || true
+
+  # Build and emit the command_complete JSONL event, injecting command: field
+  # to match the pattern used by emit_perf_event in telemetry-helpers.sh
+  jq -n -c \
+    --arg event "command_complete" \
+    --arg run_id "$run_id" \
+    --arg command "$command" \
+    --arg started_at "$started_at" \
+    --arg ended_at "$ended_at" \
+    --argjson duration_s "$duration_s" \
+    --argjson exit_status "$exit_status" \
+    '{event:$event,run_id:$run_id,command:$command,started_at:$started_at,ended_at:$ended_at,duration_s:$duration_s,exit_status:$exit_status}' \
+    >> "$perf_file" 2>/dev/null || true
+
+  # Drop the recovered state file so stale state cannot leak into a later
+  # command that fails to call init.sh.
+  rm -f "$state_file" 2>/dev/null || true
+}
+
+_closedloop_telemetry_complete "$@"
+exit 0
--- a/plugins/bootstrap/scripts/record_run.sh
+++ b/plugins/bootstrap/scripts/record_run.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+# record_run.sh - Append a run event to perf.jsonl once per Loop.
+#
+# Emits exactly one `run` event containing command, repo, branch, and start
+# time so every perf.jsonl record can be attributed to the slash-command that
+# launched the Loop. Fails open (exits 0) on any error so the caller loop is
+# unaffected.
+#
+# Usage:
+#   bash record_run.sh [WORKDIR]
+# WORKDIR defaults to $CLOSEDLOOP_WORKDIR.
+
+# Fail open: any unexpected error exits 0 so the caller loop is unaffected.
+trap 'exit 0' ERR
+
+WORKDIR="${1:-${CLOSEDLOOP_WORKDIR:-}}"
+if [[ -z "$WORKDIR" ]]; then
+  exit 0
+fi
+
+PERF_FILE="$WORKDIR/perf.jsonl"
+
+RUN_ID="${CLOSEDLOOP_RUN_ID:-unknown}"
+COMMAND="${CLOSEDLOOP_COMMAND:-interactive}"
+TIMESTAMP=$(date -u +%Y-%m-%dT%H:%M:%SZ 2>/dev/null || echo "")
+
+# Capture repo and branch via git -C. Prefer GNU `timeout` as a hang guard, but
+# fall back to bare `git` when timeout isn't installed (default macOS lacks it),
+# so this never silently drops repo/branch attribution on dev machines.
+if command -v timeout >/dev/null 2>&1; then
+  REPO=$(timeout 5 git -C "$WORKDIR" remote get-url origin 2>/dev/null || echo "")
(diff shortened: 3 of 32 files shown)


plugins/bootstrap/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "bootstrap",
   "description": "Bootstrap plugin for ClosedLoop",
-  "version": "1.2.0",
+  "version": "1.2.2",
```
mikeangstadt, line 4: **[MEDIUM]** [P3] Version bumped by 2 patch levels (1.2.0 → 1.2.2), skipping 1.2.1

CLAUDE.md requires 'the version must be updated in the same commit as the code change' with standard semver rules. This PR introduces a single cohesive change (telemetry init/complete scripts + command sections) to the bootstrap plugin and should receive one version increment. Jumping from 1.2.0 to 1.2.2 bypasses 1.2.1, implying an intermediate release that never existed. Semver version numbers are intended to be sequential within a release stream, so a single PR should advance by exactly one version.

**Recommendation:** Change version to "1.2.1" — the correct next PATCH version for this single PR's changes.

**Also affects:** `plugins/code-review/.claude-plugin/plugin.json:4`

mikeangstadt (commented):
Finding 1/8

plugins/code/scripts/tests/test_command_telemetry_init.sh
```diff
@@ -0,0 +1,409 @@
...
+run_init_and_dump() {
+    # $1 = command_name arg (may be empty string)
+    # $2 = workdir arg     (may be empty string)
+    # $3 = CLOSEDLOOP_WORKDIR env override (may be empty string)
+    local cmd_arg="${1:-}"
+    local workdir_arg="${2:-}"
+    local env_workdir="${3:-}"
+
+    bash -c "
+        $([ -n '$env_workdir' ] && echo "export CLOSEDLOOP_WORKDIR='$env_workdir'" || echo "unset CLOSEDLOOP_WORKDIR")
```
mikeangstadt, line 84: **[MEDIUM]** [P2] Single quotes prevent variable expansion in run_init_and_dump, making the unset branch dead code

On line 84, `[ -n '$env_workdir' ]` uses single quotes around `$env_workdir` inside the `$()` command substitution. Single quotes prevent variable expansion, so the test always evaluates the literal 12-character string `$env_workdir`, which is always non-empty. The `|| echo "unset CLOSEDLOOP_WORKDIR"` branch can never execute. When `env_workdir` is empty the helper produces `export CLOSEDLOOP_WORKDIR=''` instead of `unset CLOSEDLOOP_WORKDIR`. While this does not cause test failures today because the init script checks with `${CLOSEDLOOP_WORKDIR:-}` (empty string and unset behave the same), the helper cannot distinguish set-to-empty from truly-unset, and future tests relying on `${VAR+x}` semantics would silently pass incorrectly.

**Recommendation:** Replace the single-quoted test with a properly escaped double-quoted expansion: `$([ -n "$env_workdir" ] && echo "export CLOSEDLOOP_WORKDIR='$env_workdir'" || echo "unset CLOSEDLOOP_WORKDIR")`

mikeangstadt (commented):
Finding 2/8

plugins/code/scripts/command-telemetry-complete.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
mikeangstadt, line 1: **[MEDIUM]** [P2] command-telemetry-complete.sh is duplicated verbatim across three plugins (bootstrap, code-review, code)

All three files -- plugins/bootstrap/scripts/command-telemetry-complete.sh, plugins/code-review/scripts/command-telemetry-complete.sh, and plugins/code/scripts/command-telemetry-complete.sh -- are byte-for-byte identical at 99 lines each. Any bug fix or enhancement (e.g., adding new telemetry fields, fixing edge cases in duration calculation) must be manually replicated across all copies. This is the same pattern as the also-duplicated command-telemetry-init.sh files. A single shared copy sourced or executed from a common location would eliminate this maintenance risk.

**Recommendation:** Consider placing a single canonical copy (e.g., in plugins/code/scripts/) and having other plugins reference it via a symlink or a thin wrapper that delegates: `bash "${CLOSEDLOOP_SHARED_SCRIPTS:-$CLAUDE_PLUGIN_ROOT/scripts}/command-telemetry-complete.sh" "$@"`.

mikeangstadt (commented):
Finding 3/8

plugins/self-learning/scripts/command-telemetry-complete.sh
```diff
@@ -0,0 +1,99 @@
...
+    workdir="${CLOSEDLOOP_WORKDIR:-$workdir}"
+  fi
+
+  local perf_file="$workdir/perf.jsonl"
+
+  # Capture end time
+  local end_epoch
+  end_epoch=$(date +%s 2>/dev/null) || end_epoch=""
+  local ended_at
+  ended_at=$(date -u +%Y-%m-%dT%H:%M:%SZ 2>/dev/null) || ended_at=""
```
mikeangstadt, line 59: **[MEDIUM]** [P2] Sourcing unquoted .cmd-state.env silently truncates values containing spaces

This is the consumer side of the unquoted state file written by command-telemetry-init.sh. When CLOSEDLOOP_RUN_ID is missing and the fallback source of .cmd-state.env runs at line 59, unquoted values like `CLOSEDLOOP_WORKDIR=/path/with spaces` are parsed by bash as an assignment of the truncated value `/path/with` followed by an attempt to execute `spaces` as a command. The `2>/dev/null || true` suppresses the error, leaving CLOSEDLOOP_WORKDIR silently set to a wrong value. Subsequent operations (perf_file path, rm of state_file) then use the corrupted path. All four plugin copies of command-telemetry-complete.sh share this issue.

**Recommendation:** This is fixed on the writer side (command-telemetry-init.sh should use printf %q). On this side, no additional changes are needed once values are written correctly. However, adding a validation check that the sourced CLOSEDLOOP_WORKDIR is a valid directory would add defense-in-depth.

mikeangstadt (commented):
Finding 5/8

mikeangstadt (commented):
Finding 6/8

plugins/code/commands/cancel-code.md
```diff
@@ -1,18 +1,28 @@
 ---
```
mikeangstadt, line 1: **[HIGH]** [P2] allowed-tools Bash patterns use unexpanded $CLAUDE_PLUGIN_ROOT env var

The allowed-tools frontmatter adds patterns 'Bash(source $CLAUDE_PLUGIN_ROOT/scripts/command-telemetry-init.sh:*)' and 'Bash(bash $CLAUDE_PLUGIN_ROOT/scripts/command-telemetry-complete.sh:*)'. No other allowed-tools declaration in the entire codebase uses environment variables -- they all use literal paths or general tool names (e.g., 'Bash'). At runtime the agent will execute a bash command with the expanded path (e.g., 'source /home/user/.claude/plugins/.../scripts/command-telemetry-init.sh "cancel-code"'), and if the allowed-tools pattern matching does not expand $CLAUDE_PLUGIN_ROOT, the literal '$CLAUDE_PLUGIN_ROOT' in the pattern will not match the resolved path, causing the telemetry bash commands to be rejected.

**Recommendation:** Verify that Claude Code expands environment variables in allowed-tools Bash patterns. If it does not, either switch to a general 'Bash' permission or use a glob pattern that does not rely on env var expansion. Alternatively, use '${CLAUDE_PLUGIN_ROOT}' syntax consistent with hooks.json if that form is supported in allowed-tools.

plugins/bootstrap/scripts/command-telemetry-complete.sh
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env bash
```
mikeangstadt, line 1: **[MEDIUM]** [P2] command-telemetry-complete.sh duplicated verbatim across 4 plugins

The patch introduces 4 identical copies of command-telemetry-complete.sh: plugins/bootstrap/scripts/, plugins/code-review/scripts/, plugins/code/scripts/, and plugins/self-learning/scripts/. All 4 files are byte-for-byte identical (99 lines each). The same duplication applies to command-telemetry-init.sh (4 copies). Any future bug fix or behavioural change must be applied in 4 separate files. This also means that the 3 new plugin copies could drift from the code/ canonical version in future PRs without anyone noticing.

**Recommendation:** Place the canonical copies in a shared location (e.g., a new `plugins/shared/scripts/` directory, or symlink to `plugins/code/scripts/`). The existing plugin infrastructure uses `CLAUDE_PLUGIN_ROOT` per plugin, but the scripts themselves only rely on env vars (`CLOSEDLOOP_WORKDIR`, `CLOSEDLOOP_RUN_ID`, etc.) and have no plugin-specific logic — making them ideal candidates for a shared location.

**Also affects:** `plugins/bootstrap/scripts/command-telemetry-init.sh:1`

mikeangstadt (commented):
Finding 7/8

plugins/bootstrap/scripts/record_run.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
mikeangstadt, line 1: **[MEDIUM]** [P2] record_run.sh copied 3 times; identical to existing plugins/code/scripts/record_run.sh

The patch adds plugins/bootstrap/scripts/record_run.sh, plugins/code-review/scripts/record_run.sh, and plugins/self-learning/scripts/record_run.sh. All three are identical to the pre-existing plugins/code/scripts/record_run.sh (50 lines). The only reason these copies are needed is that command-telemetry-init.sh resolves record_run.sh relative to its own location via `${BASH_SOURCE[0]}`. There are now 4 copies that must be kept in sync. A future fix to the git timeout logic or jq invocation would require 4 edits.

**Recommendation:** Reference the canonical copy from plugins/code/scripts/record_run.sh, or create a shared scripts location. If the co-location requirement (same directory as command-telemetry-init.sh) is hard, command-telemetry-init.sh could be modified to look for record_run.sh at a well-known absolute location (e.g., via a CLOSEDLOOP_PLUGIN_SCRIPTS env var or by walking up from BASH_SOURCE[0] to find the code plugin).

mikeangstadt (commented):
Finding 8/8

mikeangstadt:
**[MEDIUM] plugins/code/scripts/command-telemetry-init.sh:81** [P2] .cmd-state.env values written unquoted cause silent data corruption when paths contain spaces

The printf statements write variable values without shell quoting: `printf 'CLOSEDLOOP_WORKDIR=%s\n' "$workdir"` produces `CLOSEDLOOP_WORKDIR=/path/with spaces`. When command-telemetry-complete.sh later sources this file, bash parses `CLOSEDLOOP_WORKDIR=/path/with` as an assignment and `spaces` as a command. The variable gets silently truncated to the portion before the first space, and the `2>/dev/null || true` on the source call suppresses the error. The result is that perf.jsonl is written to the wrong (truncated) path. This triggers when CLOSEDLOOP_WORKDIR contains spaces, which is realistic on macOS where usernames with spaces (e.g., /Users/John Smith/...) create such paths. The same pattern appears identically in all four plugin copies of command-telemetry-init.sh (bootstrap, code-review, code, self-learning).

**Recommendation:** Use printf with %q format specifier to produce properly shell-escaped values: `printf 'CLOSEDLOOP_WORKDIR=%q\n' "$workdir"`. This ensures the sourced file handles spaces, quotes, and other metacharacters correctly.

plugins/code/scripts/run-loop.sh
```diff
@@ -1246,83 +1246,8 @@ log_progress() {
...
-    --arg started_at "$now" \
-    --arg ended_at "$now" \
-    --argjson duration_s 0 \
-    --argjson exit_code 0 \
-    --argjson skipped true \
-    '{event:$event,run_id:$run_id,iteration:$iteration,step:$step,step_name:$step_name,started_at:$started_at,ended_at:$ended_at,duration_s:$duration_s,exit_code:$exit_code,skipped:$skipped}'
-  )"
-}
+# shellcheck source=telemetry-helpers.sh
+source "${SCRIPTS_DIR}/telemetry-helpers.sh"
```
thadeusb, line 1250: > Note: This comment targets `run-loop.sh:1144` (MAX_ITERATIONS default) which is outside this diff hunk. Posting here as the nearest changed line in this file.

MAX_ITERATIONS went from 50 to 5. That's a 10x reduction in the default iteration budget — every loop that doesn't pass `--max-iterations` explicitly now caps at 5. Is this intentional or left over from testing? The PR description only mentions telemetry instrumentation, doesn't say anything about changing core loop behavior.

PR #86, opened 2026-05-12
FEA-860: Verify user-scoped ClosedLoop plugins

shafty023:
## Summary
- Repairs the installer to verify all ClosedLoop plugins are user-scoped, enabled, and backed by existing install paths.
- Re-enables disabled user-scoped plugins once and reports actionable repair commands for project-scoped duplicates.
- Adds installer tests and updates the plugin changelog and installation docs.

## Links
- Work item: https://app.closedloop.ai/features/FEA-860
- Plan: https://app.closedloop.ai/implementation-plans/PLN-552

## Validation
- `bash -n install.sh` - passed
- `pytest plugins/code/tools/python/test_install_script.py` - passed, 7 tests
- `ruff check plugins/code/tools/python/test_install_script.py` - passed
- `git diff --check` - passed

## Known Blockers
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,19 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code v1.11.20
+
+#### Changed
+- README installation guidance now states the installer installs and verifies the five Symphony runtime plugins at user scope, with `bootstrap` excluded from the default runtime install.
+
+#### Fixed
+- `install.sh` now refreshes the configured `closedloop-ai` marketplace, installs the five Symphony runtime plugins at user scope, then verifies those runtime plugins have existing install paths and enabled user-scoped `claude plugin list --json` entries. Disabled user-scoped runtime plugins are re-enabled once and re-read before the installer reports success.
+- Project-scoped ClosedLoop plugin duplicates are repaired before user-scope install/update when Claude reports a usable `projectPath`; entries without a usable project path now produce a manual project-directory uninstall command while user-scope repair continues.
+
 ### code v1.11.19
 
 #### Changed
 - `test_write_runs_log_entry_uses_workdir_root` in `test_run_loop_failure_marker.py` now `unset`s `CLOSEDLOOP_COMMAND` and `LAST_CLAUDE_COMMAND` inside the bash heredoc before invoking `write_runs_log_entry`, so the default-command path is exercised deterministically regardless of the caller's ambient environment. Test-only change isolating the existing behavior — no production code paths altered.
-- README installation guidance now states the user-scope verification, disabled-plugin re-enable, and project-scope repair behavior performed by the installer.
-
-#### Fixed
-- `install.sh` now installs all six ClosedLoop plugins at user scope, then verifies that the five Symphony runtime plugins have existing install paths and enabled user-scoped `claude plugin list --json` entries. Disabled user-scoped runtime plugins are re-enabled once and re-read before the installer reports success.
-- Project-scoped ClosedLoop plugin duplicates are repaired before user-scope install/update when Claude reports a usable `projectPath`; entries without a usable project path now produce a manual project-directory uninstall command while user-scope repair continues.
 
 ### judges v1.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 #### Changed
 - `test_write_runs_log_entry_uses_workdir_root` in `test_run_loop_failure_marker.py` now `unset`s `CLOSEDLOOP_COMMAND` and `LAST_CLAUDE_COMMAND` inside the bash heredoc before invoking `write_runs_log_entry`, so the default-command path is exercised deterministically regardless of the caller's ambient environment. Test-only change isolating the existing behavior — no production code paths altered.
+- README installation guidance now states the user-scope verification, disabled-plugin re-enable, and project-scope repair behavior performed by the installer.
+
+#### Fixed
+- `install.sh` now installs all six ClosedLoop plugins at user scope, then verifies that the five Symphony runtime plugins have existing install paths and enabled user-scoped `claude plugin list --json` entries. Disabled user-scoped runtime plugins are re-enabled once and re-read before the installer reports success.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code v1.11.16
```
mikeangstadt: **[MEDIUM]** [P2] CHANGELOG entry labels new features as v1.11.16 but plugin.json ships them as v1.11.18

The patch adds `### code v1.11.16` to CHANGELOG.md while the same changeset bumps `plugins/code/.claude-plugin/plugin.json` from `1.11.17` to `1.11.18`. The features described in the entry — install.sh user-scope verification, disabled-plugin re-enable, project-scope repair, and the README installation guidance update — are all introduced by this PR. The install.sh on disk before this patch contains none of the new functions (`read_plugin_list`, `plugin_enabled_state`, `repair_project_scoped_plugins`, `ensure_user_plugin_enabled`, `registry_has_user_install`, `mark_plugin_failed`, `plugin_failed`), proving these are new code in this PR rather than retroactive documentation of a prior release. CLAUDE.md states 'Any change to a plugin's files MUST include a version bump in that plugin's `plugin.json`' — the bump is present (1.11.17→1.11.18), but the CHANGELOG label `v1.11.16` does not match the released version `v1.11.18`. A reader correlating changelog notes to code would find the install.sh verification behavior documented under v1.11.16 while the code carries version 1.11.18. CLAUDE.md also states 'Do NOT manually edit `CHANGELOG.md` or `README.md`' — the version mismatch is evidence these files were edited manually rather than generated by `/update-documentation`, which would use the current plugin version number.

**Recommendation:** Change `### code v1.11.16` to `### code v1.11.18` so the CHANGELOG entry matches the version in `plugin.json`. If v1.11.16 and v1.11.17 also need backfilled entries, add them as separate sections above v1.11.15. Prefer running `/update-documentation` to generate CHANGELOG and README updates rather than editing them manually.

install.sh
```diff
@@ -129,29 +251,65 @@ SNAPSHOT_POST="$WORK_DIR/snapshot_post"
...
+fi
+
+for plugin in "${PLUGINS[@]}"; do
+    plugin_ref="${plugin}@${MARKETPLACE_NAME}"
+    if ! registry_has_user_install "$plugin_ref"; then
+        warn "Missing user-scoped registry entry with existing installPath: $plugin_ref"
+        mark_plugin_failed "$plugin_ref"
+        continue
+    fi
+    if [[ ! -s "$FINAL_LIST" ]] || [[ "$(plugin_enabled_state "$FINAL_LIST" "$plugin_ref" 2>/dev/null || echo missing)" != "enabled" ]]; then
```
mikeangstadt: **[HIGH]** [P1] Final verification loop marks all plugins as failed when FINAL_LIST read fails

If `read_plugin_list "$FINAL_LIST"` fails (e.g., transient `claude plugin list --json` error), the script only warns at line ~292, but the verification loop at line ~303 still runs. The condition `[[ ! -s "$FINAL_LIST" ]] || ...` short-circuits to true for every plugin when FINAL_LIST is empty/missing, causing `mark_plugin_failed` to be called for all 6 plugins. This overrides the earlier per-plugin success from `ensure_user_plugin_enabled`, making the summary report all plugins as failed and triggering `exit 1` with misleading repair instructions -- even though install + enable all succeeded.

**Recommendation:** Skip the FINAL_LIST enabled-state check when the list file is unavailable. Either wrap the entire verification loop in `if [[ -s "$FINAL_LIST" ]]; then ... fi`, or replace `[[ ! -s "$FINAL_LIST" ]]` with a `continue` that skips the enabled check for that plugin when the file is invalid.

CHANGELOG.md
mikeangstadt: **[MEDIUM]** [P2] CHANGELOG version label v1.11.16 does not match plugin.json version v1.11.18

The new CHANGELOG entry is titled `### code v1.11.16` but the same PR bumps `plugins/code/.claude-plugin/plugin.json` from `1.11.17` to `1.11.18`. The CHANGELOG should document the version being released (v1.11.18), not v1.11.16. The previous CHANGELOG entry is `### code v1.11.15`, confirming v1.11.16 and v1.11.17 had no CHANGELOG entries, but the fix should label this entry as v1.11.18 to match the actual version.

**Recommendation:** Change `### code v1.11.16` to `### code v1.11.18` in the CHANGELOG.

install.sh
```diff
@@ -55,6 +55,128 @@ chmod 700 "$WORK_DIR"
...
+  claude plugin uninstall "$p@closedloop-ai" --scope project
+  claude plugin install "$p@closedloop-ai" --scope user
+  claude plugin enable "$p@closedloop-ai" --scope user
+done
+claude plugin list --json
+REPAIR
+}
+
+FAILED_PLUGINS=()
+mark_plugin_failed() {
```
mikeangstadt, line 186: **[MEDIUM]** [P2] mark_plugin_failed duplicates the array-membership loop already in plugin_failed

Both `mark_plugin_failed` (lines 161-169) and `plugin_failed` (lines 171-178) contain the identical three-line loop that scans `FAILED_PLUGINS` for a matching entry and returns 0 if found. Because `mark_plugin_failed` only appends when the item is not already present, it can simply delegate the existence check to `plugin_failed` instead of repeating the loop. Any future change to the membership-check logic (e.g. case-insensitive match) must be applied in both places or bugs are introduced silently.

**Recommendation:** Replace the duplicated loop inside `mark_plugin_failed` with a call to `plugin_failed`: `plugin_failed "$plugin_ref" && return 0`. This eliminates the duplication and makes the deduplication logic canonical.

mikeangstadt (commented):
Code review by Claude Code — 4 findings (1 HIGH, 3 MEDIUM). See inline comments.

CHANGELOG.md
shafty023: Handled. After rebasing onto current main, code v1.11.18 already exists as a shipped changelog section, so I followed the /update-documentation rule for that case: bumped plugins/code/.claude-plugin/plugin.json to 1.11.19 and added this installer/README entry as the top CHANGELOG.md section for code v1.11.19.

install.sh
shafty023: Handled. read_plugin_list "$FINAL_LIST" now records FINAL_LIST_AVAILABLE=0 on failure, and the final verification loop continues past the enabled-state check instead of marking every plugin failed. Registry/installPath verification still runs. Added a regression test covering a final list failure with all installs otherwise successful.

PR #88, opened 2026-05-12
feat(code): add per-user security override mechanism

mikeangstadt:
## Summary
- Adds a granular override system to `pretooluse-hook.sh` so users can selectively unblock security-restricted categories via `~/.claude/security-overrides.json`
- Splits the monolithic cloud credentials category into independent `cloud-credentials-cmd` (gcloud auth commands) and `cloud-credentials-files` (~/.aws/credentials, ~/.config/gcloud/*) categories
- Supports override via `CLAUDE_SECURITY_OVERRIDES` env var for testing

### Override categories
| Category | What it unblocks |
|---|---|
| `process-kill` | `pkill`, `killall` |
| `keychain` | macOS `security find-*-password`, `dump-keychain` |
| `browser-profiles` | Chrome/Firefox/Safari profile dirs and sqlite DBs |
| `ssh-keys` | `~/.ssh/id_*` |
| `cloud-credentials-cmd` | `gcloud auth print-access-token`, `gcloud auth application-default` |
| `cloud-credentials-files` | `~/.aws/credentials`, `~/.config/gcloud/*` |

### Example `~/.claude/security-overrides.json`
```json
{
  "overrides": {
    "cloud-credentials-cmd": true,
    "cloud-credentials-files": true,
    "process-kill": true
  }
}
```

## Test plan
- [x] 33 new tests in `test_security_overrides.sh` — all passing
- [ ] Verify default deny still works with no override file present
- [ ] Verify partial overrides only unblock the specified category
- [ ] Verify malformed JSON override file degrades safely to deny-all
- [ ] Verify cloud-credentials-cmd and cloud-credentials-files are independent

### DIFF
--- a/plugins/code/hooks/pretooluse-hook.sh
+++ b/plugins/code/hooks/pretooluse-hook.sh
@@ -30,6 +30,35 @@ CWD=$(echo "$INPUT" | jq -r '.cwd // empty')
 # rather than auto-allowed by the workspace rule below.
 _SEC_DENY='{"hookSpecificOutput":{"hookEventName":"PreToolUse","permissionDecision":"deny","permissionDecisionReason":"Blocked: credential access denied by security policy"}}'
 
+# ── Security overrides ─────────────────────────────────────────────────────
+# Per-user overrides loaded from ~/.claude/security-overrides.json.
+# Only categories explicitly set to true are allowed through.
+# Categories: process-kill, keychain, browser-profiles, ssh-keys,
+#             cloud-credentials-cmd, cloud-credentials-files
+_OVERRIDES_FILE="${CLAUDE_SECURITY_OVERRIDES:-$HOME/.claude/security-overrides.json}"
+
+_override_enabled() {
+    local category="$1"
+    if [[ -f "$_OVERRIDES_FILE" ]]; then
+        local val
+        val=$(jq -r --arg cat "$category" '.overrides[$cat] // false' "$_OVERRIDES_FILE" 2>/dev/null)
+        if [[ "$val" == "true" ]]; then
+            echo "$(date): SECURITY OVERRIDE: allowed category=$category" >> "${DEBUG_LOG:-/dev/null}"
+            return 0
+        fi
+    fi
+    return 1
+}
+
+_sec_deny_unless_override() {
+    local category="$1"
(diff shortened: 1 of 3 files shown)


plugins/code/hooks/pretooluse-hook.sh
```diff
@@ -39,39 +68,37 @@ case "$TOOL_NAME" in
             # (e.g. a running desktop-dev in the main tree killed by a worktree agent).
             # Claude should never need these; use process.kill(pid) for specific PIDs instead.
             *pkill*|*killall*)
-                echo "$_SEC_DENY"
-                exit 0
+                _sec_deny_unless_override "process-kill"
```
mikeangstadt, line 98: Because Bash routing still uses a first-match `case`, an override on an earlier category can mask a later disallowed one in the same command string. For example, enabling `process-kill` currently allows `pkill node && cat ~/.aws/credentials` because the `pkill` arm returns before the cloud-credentials-file check ever runs. This needs per-category scanning instead of first-match routing.

plugins/code/hooks/pretooluse-hook.sh
```diff
@@ -30,6 +30,35 @@ CWD=$(echo "$INPUT" | jq -r '.cwd // empty')
...
+#             cloud-credentials-cmd, cloud-credentials-files
+_OVERRIDES_FILE="${CLAUDE_SECURITY_OVERRIDES:-$HOME/.claude/security-overrides.json}"
+
+_override_enabled() {
+    local category="$1"
+    if [[ -f "$_OVERRIDES_FILE" ]]; then
+        local val
+        val=$(jq -r --arg cat "$category" '.overrides[$cat] // false' "$_OVERRIDES_FILE" 2>/dev/null)
+        if [[ "$val" == "true" ]]; then
+            echo "$(date): SECURITY OVERRIDE: allowed category=$category" >> "${DEBUG_LOG:-/dev/null}"
```
mikeangstadt: This audit log never reaches the per-run debug log in a normal ClosedLoop session because `DEBUG_LOG` is still `/dev/null` here and only gets rebound to the workdir log later in the script. As written, allowed overrides leave no persistent audit trail despite the new logging.

PR #94, opened 2026-05-18
Change model from 'sonnet' to 'opus'

mikeangstadt:
Update to latest model for performance.

### DIFF
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.21",
+  "version": "1.12.0",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.20",
+  "version": "1.11.21",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/code/agents/implementation-subagent.md
+++ b/plugins/code/agents/implementation-subagent.md
@@ -1,7 +1,7 @@
 ---
 name: implementation-subagent
 description: Implements missing requirements for a task from the implementation plan.
-model: sonnet
+model: opus
 tools: Read, Write, Edit, Glob, Grep, Bash
 ---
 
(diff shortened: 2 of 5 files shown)


plugins/code/agents/implementation-subagent.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: implementation-subagent
 description: Implements missing requirements for a task from the implementation plan.
-model: sonnet
+model: opus
```
shafty023, line 4: The changed model conflicts with the documented repository model contract. `CLAUDE.md` and `CONTRIBUTING.md` both define `opus` for creative/planning work and `sonnet` for implementation, while this agent is the implementation worker invoked for normal write/fix loops. Changing it to `opus` silently changes the runtime model, cost/latency, and availability assumptions for implementation work.

Required fix: keep this as `model: sonnet`, or first update the model-selection policy, generated docs, and validation so implementation agents are intentionally allowed to use `opus`.

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.11.20",
+  "version": "1.11.21",
```
shafty023: If the Opus change is intentional, this patch version understates the compatibility impact. The repo semver guidance reserves PATCH for bug fixes or prompt wording, but this PR changes runtime-visible agent metadata for the implementation worker. Consumers updating from 1.11.20 to 1.11.21 would get a different model without a minor-version signal.

Required fix: either revert the model change and keep a patch-level bump only for a true patch, or bump at least MINOR and document the compatibility impact with the generated docs/changelog update.

plugins/code/agents/implementation-subagent.md
mikeangstadt: Addressed in 9303a80. Rather than revert, I made the opus switch intentional: documented it as a deliberate exception to the "sonnet for implementation" convention in both `CLAUDE.md` and `CONTRIBUTING.md`, so the model-selection policy now sanctions opus for the implementation worker. No validation enforces a model whitelist for this agent, so no validator change was needed. The runtime/cost/latency impact is now called out explicitly in the docs and CHANGELOG.

plugins/code/.claude-plugin/plugin.json
mikeangstadt: Addressed in 9303a80. Since the model change is intentional, I bumped the compatibility signal accordingly: `plugins/code/.claude-plugin/plugin.json` now moves 1.11.20 -> 1.12.0 (MINOR) instead of a PATCH, matching the runtime-visible agent-metadata change. The generated `CHANGELOG.md` documents the model/cost/latency impact under `### code v1.12.0`, and the model-selection policy in `CLAUDE.md`/`CONTRIBUTING.md` now records the opus exception.

PR #96, opened 2026-05-20
feat(code): dynamic specialist agent selection in Phase 3

mikeangstadt:
## Summary

- Orchestrator Phase 3 now **dynamically discovers** all available agents (repo-level + plugin) at runtime and selects the best-fit specialist per implementation task via LLM reasoning
- 7 repo-level agents gain **dual-mode** (critic + implementation) with write-capable tools and the same self-verification gate pattern as `implementation-subagent`
- Phase 5.5 `code_drift` routing also uses dynamic selection instead of hardcoding `implementation-subagent`
- Works automatically with any agent — gStack, custom repo agents, installed plugins — no config changes needed

## Test plan

- [ ] Run `/code:code` on a feature touching Python files — verify orchestrator discovers and selects `python-script-reviewer` over generalist
- [ ] Run `/code:code` on a feature with no specialist match — verify fallback to `implementation-subagent`
- [ ] Install an additional plugin (e.g., gStack) and verify it appears in agent discovery output
- [ ] Verify critic mode still works normally for all 7 updated agents (no regression from added tools)
- [ ] Verify Phase 5.5 code_drift routes to specialists when applicable

Implements [PLN-626](https://app.closedloop.ai/implementation-plans/PLN-626)

### DIFF
--- a/.claude/agents/agent-definition-expert.md
+++ b/.claude/agents/agent-definition-expert.md
@@ -4,10 +4,63 @@ description: Reviews Claude Code agents for YAML frontmatter, writing style, str
 model: sonnet
 color: cyan
 skills: platform:claude-code-expert
-tools: Read, Grep, Glob, Skill
+tools: Read, Write, Edit, Grep, Glob, Bash, Skill
 ---
 
-## Your Role
+## Execution Modes
+
+This agent supports two execution modes:
+
+1. **Critic Mode (default)**: Validate Claude Code agent definition files against platform requirements
+2. **Implementation Mode**: Create or modify agent definition files using platform expertise
+
+### Mode Detection
+
+**Implementation mode** if prompt contains: `WORKDIR=`, `Implement task`, `Missing requirements:`
+**Critic mode** otherwise (default)
+
+---
+
+## Implementation Mode
+
+When activated in implementation mode, combine your agent platform expertise with write-capable tools to implement agent definitions.
+
+### Instructions
+
+1. Read existing agent files related to the task
+2. Before creating new agents, read existing agents to follow established patterns
+3. Ensure YAML frontmatter is valid: `name`, `description`, `model`, `color` required
+4. Use comma-separated inline format for `tools` and `skills` (never block arrays)
+5. Follow model selection convention: opus for creative/planning, sonnet for implementation, haiku for coordination
+6. Implement ONLY the missing requirements provided
+7. After implementing, proceed to the Self-Verification Gate below
+
+### Self-Verification Gate
+
+After implementing, you MUST pass all four gates before emitting the completion promise.
+
+**Gate 1: Re-read Modified Files** — For every file you created or modified, use Read to re-read it in full. Verify correctness.
+
+**Gate 2: Requirement Verification** — For each item in the NOT_IMPLEMENTED list, locate specific `file:line` evidence:
+```
+VERIFICATION:
+- "requirement description" → PASS (agents/name.md:1 - valid frontmatter)
+- "another requirement" → FAIL (not found)
+```
+If any requirement has FAIL status, go back and implement it.
+
+**Gate 3: Integration Check** — For each new agent created, verify the name matches the filename and skills/tools referenced exist.
+
+**Gate 4: Static Analysis** — Verify YAML frontmatter parses correctly. Check description length < 1024 chars.
+
+### Return Format
+
+**Success:** Output `IMPLEMENTATION_VERIFIED:` with file changes, then `<promise>IMPLEMENTATION_VERIFIED</promise>`
+**Blocked:** Output `BLOCKED:` with details, then `<promise>IMPLEMENTATION_VERIFIED</promise>`
+
+---
+
+## Critic Mode
 
 Validate Claude Code agent definition files against platform requirements and project conventions. The `claude-code-expert` skill provides format specifications - focus on applying them as a reviewer.
 
--- a/.claude/agents/agent-definition-expert.md
+++ b/.claude/agents/agent-definition-expert.md
@@ -3,7 +3,7 @@ name: agent-definition-expert
 description: Reviews Claude Code agents for YAML frontmatter, writing style, structure, and format compliance.
 model: sonnet
 color: cyan
-skills: platform:claude-code-expert
+skills: platform:claude-code-expert, implementation-self-check
 tools: Read, Write, Edit, Grep, Glob, Bash, Skill
 ---
 
@@ -33,30 +33,16 @@ When activated in implementation mode, combine your agent platform expertise wit
 4. Use comma-separated inline format for `tools` and `skills` (never block arrays)
 5. Follow model selection convention: opus for creative/planning, sonnet for implementation, haiku for coordination
 6. Implement ONLY the missing requirements provided
-7. After implementing, proceed to the Self-Verification Gate below
+7. Activate skill `implementation-self-check` and follow its shared four-gate protocol using the agent-specific checks below.
 
-### Self-Verification Gate
+### Agent-Specific Verification Checks
 
-After implementing, you MUST pass all four gates before emitting the completion promise.
+When the `implementation-self-check` skill reaches Gates 3 and 4, apply these checks:
 
-**Gate 1: Re-read Modified Files** — For every file you created or modified, use Read to re-read it in full. Verify correctness.
+- **Gate 3: Integration Check** — For each new agent created, verify the name matches the filename and skills/tools referenced exist.
+- **Gate 4: Static Analysis** — Verify YAML frontmatter parses correctly. Check description length < 1024 chars.
 
-**Gate 2: Requirement Verification** — For each item in the NOT_IMPLEMENTED list, locate specific `file:line` evidence:
-```
-VERIFICATION:
-- "requirement description" → PASS (agents/name.md:1 - valid frontmatter)
-- "another requirement" → FAIL (not found)
-```
-If any requirement has FAIL status, go back and implement it.
-
-**Gate 3: Integration Check** — For each new agent created, verify the name matches the filename and skills/tools referenced exist.
-
-**Gate 4: Static Analysis** — Verify YAML frontmatter parses correctly. Check description length < 1024 chars.
-
-### Return Format
-
-**Success:** Output `IMPLEMENTATION_VERIFIED:` with file changes, then `<promise>IMPLEMENTATION_VERIFIED</promise>`
-**Blocked:** Output `BLOCKED:` with details, then `<promise>IMPLEMENTATION_VERIFIED</promise>`
+Use the skill's standard `IMPLEMENTATION_VERIFIED` / `BLOCKED` return format.
 
 ---
 
--- a/.claude/agents/json-schema-architect.md
+++ b/.claude/agents/json-schema-architect.md
@@ -3,7 +3,8 @@ name: json-schema-architect
 description: Reviews JSON Schema draft-07 design patterns, validation rules, schema composition, and contract adherence for agent definitions, plugin manifests, and workflow artifacts. Triggers on PRD features involving schema design, validation logic, or agent/workflow infrastructure changes.
 model: sonnet
 color: green
-tools: Read, Write, Edit, Grep, Glob, Bash
+skills: implementation-self-check
+tools: Read, Write, Edit, Grep, Glob, Bash, Skill
 ---
 
 ## Execution Modes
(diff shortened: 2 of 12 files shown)


.claude/agents/test-engineer.md
```diff
@@ -1,19 +1,73 @@
 ---
 name: test-engineer
-description: Specialized in Python testing with pytest. Expert in running tests, fixing failures, and ensuring test coverage. Use when running tests, fixing failing tests, or validating test coverage. Does NOT write new tests - focuses on running and fixing.
+description: Specialized in Python testing with pytest. Expert in running tests, fixing failures, and ensuring test coverage. Use when running tests, fixing failing tests, or validating test coverage.
 model: sonnet
 color: yellow
+tools: Read, Write, Edit, Grep, Glob, Bash
```
thadeusb: You added an explicit tools list but left out Skill. Before this PR the agent had no tools field, which means it got all tools. Now it's locked to these 6. Line 66 still says `activate skill python-patterns` in critic mode, which requires the Skill tool. That call is going to fail at runtime.

.claude/agents/json-schema-architect.md
```diff
@@ -3,11 +3,59 @@ name: json-schema-architect
 description: Reviews JSON Schema draft-07 design patterns, validation rules, schema composition, and contract adherence for agent definitions, plugin manifests, and workflow artifacts. Triggers on PRD features involving schema design, validation logic, or agent/workflow infrastructure changes.
 model: sonnet
 color: green
+tools: Read, Write, Edit, Grep, Glob, Bash
```
thadeusb: Same issue as test-engineer. This agent previously had no tools field (= all tools). Now you're pinning it to 6 tools without Skill. Less urgent since the body doesn't explicitly call activate skill, but you're also stripping Agent, which means it can't spawn subagents if it ever needed to. Was the intent to restrict capabilities or just to add Write/Edit? If the latter, start from what each agent actually needs, don't just paste the same list everywhere.

plugins/code/prompts/prompt.md
```diff
@@ -205,16 +205,46 @@ Here are the key phases you must complete:
 
 - Activate `code:plan-validate` skill (runs Python script against $CLOSEDLOOP_WORKDIR) — semantic check is unnecessary here since the plan hasn't changed since Phase 2.7
 - If `pending_tasks` is empty, all tasks are done → proceed to Phase 4
+
+**Dynamic agent discovery (once at phase start):**
+
+Discover all available specialist agents that could handle implementation tasks. Run this Bash command and store the output in working memory as `available_agents`:
+
+```bash
+echo "=== Repo-level agents ===" && for f in .claude/agents/*.md; do name=$(head -10 "$f" 2>/dev/null | grep '^name:' | head -1 | sed 's/^name: *//'); desc=$(head -10 "$f" 2>/dev/null | grep '^description:' | head -1 | sed 's/^description: *//'); tools=$(head -10 "$f" 2>/dev/null | grep '^tools:' | head -1 | sed 's/^tools: *//'); [ -n "$name" ] && echo "  @$name | $desc | tools: $tools"; done 2>/dev/null; echo "=== Plugin agents ===" && echo "  @code:implementation-subagent | Generalist implementation agent (fallback)"
```
thadeusb: The "Plugin agents" section of this discovery command just hardcodes `@code:implementation-subagent`. That's not discovery, that's a constant. The PR description says "Works automatically with any agent — gStack, custom repo agents, installed plugins — no config changes needed" but installed plugin agents won't show up here at all. Either actually discover plugin agents or drop that claim from the PR description.

.claude/agents/agent-definition-expert.md
```diff
@@ -4,10 +4,63 @@ description: Reviews Claude Code agents for YAML frontmatter, writing style, str
...
+
+1. Read existing agent files related to the task
+2. Before creating new agents, read existing agents to follow established patterns
+3. Ensure YAML frontmatter is valid: `name`, `description`, `model`, `color` required
+4. Use comma-separated inline format for `tools` and `skills` (never block arrays)
+5. Follow model selection convention: opus for creative/planning, sonnet for implementation, haiku for coordination
+6. Implement ONLY the missing requirements provided
+7. After implementing, proceed to the Self-Verification Gate below
+
+### Self-Verification Gate
```
thadeusb: This 4-gate self-verification block is copy-pasted almost identically across all 7 agent files in this PR. Only the Gate 3/4 details differ slightly. CLAUDE.md [mistake] rule says to extract shared helpers when used by 2+ files. When the gate protocol changes you'll need to update 7 files in lockstep. This should be a skill that each agent activates, not inline copy-paste.

.claude/agents/test-engineer.md
mikeangstadt: Added Skill support here via the shared implementation-self-check skill, and moved test-engineer onto explicit skills frontmatter so implementation mode can invoke both that shared verifier and python-patterns without losing capability.

.claude/agents/agent-definition-expert.md
mikeangstadt: Extracted the shared four-gate implementation verifier into .claude/skills/implementation-self-check/SKILL.md. The individual agent files now keep only their domain-specific Gate 3 and Gate 4 checks and delegate the common protocol and return format to the shared skill.

plugins/code/prompts/prompt.md
mikeangstadt: Replaced the hardcoded plugin fallback with discover_available_agents.py. It now enumerates repo agents, local monorepo plugin agents, and installed plugin-cache agents, and returns namespaced invocations like @plugin:agent for the orchestrator to choose from.

.claude/agents/json-schema-architect.md
mikeangstadt: Adjusted this agent to use a purpose-built tool set instead of the pasted six-tool list: it now adds Skill because implementation mode explicitly activates the shared verifier skill, while the rest stays scoped to the tools the agent actually uses.

thadeusb (approved):

plugins/code/prompts/prompt.md
```diff
@@ -205,16 +205,52 @@ Here are the key phases you must complete:
...
+
+The discovery script scans repo agents from `.claude/agents/`, local plugin agents from `plugins/*/agents/` when working in this monorepo, and installed plugin agents from `~/.claude/plugins/cache/`.
+
+**Agent selection (per-task, LLM-decided):**
+
+For each task, after verification identifies NOT_IMPLEMENTED requirements, choose the best agent:
+
+1. Consider the task description, the `files:` list, and the `missing:` requirements from verification output
+2. Review `available_agents` in working memory — pick the agent whose description is the **best domain match** for this task
+3. An agent is eligible for implementation only if its tools include **Write or Edit** (read-only agents cannot implement)
```
shafty023: This selection contract is too loose for the new routing. Discovery only emits freeform `invocation | description | tools`, then the prompt asks the LLM to choose any agent whose tools mention `Write` or `Edit`. That can select write-capable plan/build/review agents that do not implement the `IMPLEMENTATION_VERIFIED` contract, excludes agents rendered as `tools: inherited` even when inherited tools are write-capable, and does not satisfy PLN-626's file/domain-pattern mapping. Please make discovery emit a structured implementation-capability record (domain/file patterns, implementation-capable flag, fallback rank, trust source) and have the prompt consume that instead of freeform descriptions/tool strings.

plugins/code/prompts/prompt.md
```diff
@@ -205,16 +205,52 @@ Here are the key phases you must complete:
...
 - Activate `code:plan-validate` skill (runs Python script against $CLOSEDLOOP_WORKDIR) — semantic check is unnecessary here since the plan hasn't changed since Phase 2.7
 - If `pending_tasks` is empty, all tasks are done → proceed to Phase 4
+
+**Dynamic agent discovery (once at phase start):**
+
+Discover all available specialist agents that could handle implementation tasks. Run this Bash command and store the output in working memory as `available_agents`:
+
+```bash
+python3 "${CLAUDE_PLUGIN_ROOT}/tools/python/discover_available_agents.py" \
+  --workspace "$PWD" \
```
shafty023, line 215: Using `$PWD` scans the command launch directory, not necessarily the target repo. `/code:code` resolves the target into `CLOSEDLOOP_WORKDIR`, so invoking the command from elsewhere can silently miss the repo's `.claude/agents/` and `plugins/*/agents/`. Also define a hard fallback to `@code:implementation-subagent` when discovery fails, returns empty output, or is unparseable. Please pass the resolved workdir/repo root here and add a subprocess regression test with `PWD != CLOSEDLOOP_WORKDIR` and `--plugin-root` outside the workspace.

plugins/code/prompts/prompt.md
```diff
@@ -205,16 +205,52 @@ Here are the key phases you must complete:
...
+5. Prefer specificity: a domain specialist that fits the task well will outperform the generalist
+
 - For each task in `pending_tasks`:
   1. **Update state.json** with task-level tracking (see State Tracking section above)
   2. Launch @code:verification-subagent with prompt: "WORKDIR=$CLOSEDLOOP_WORKDIR. Verify task T-X.Y: {task description}"
   3. Process based on result:
      - **VERIFIED**: Proceed to step 4
-     - **NOT_IMPLEMENTED**: Parse the `missing:` and `files:` sections from the verification output. Launch @code:implementation-subagent with prompt: "WORKDIR=$CLOSEDLOOP_WORKDIR. Implement task T-X.Y: {task description}. Missing requirements: {missing list}. Relevant source files already identified: {files list}"
-       - After implementation-subagent returns, check its output:
+     - **NOT_IMPLEMENTED**: Parse the `missing:` and `files:` sections from the verification output. Select the best implementation agent using the **agent selection** logic above. Launch the selected agent with prompt: "WORKDIR=$CLOSEDLOOP_WORKDIR. Implement task T-X.Y: {task description}. Missing requirements: {missing list}. Relevant source files already identified: {files list}"
```
shafty023, line 266: Launching arbitrary selected agents bypasses the existing loop-agent contract. `plugins/code/scripts/loop-agents.json` only registers `code:implementation-subagent`, `loop-stop-hook.sh` skips unregistered agent types, and `merge_build_result.py` still attributes build outcomes only to `implementation-subagent`. Specialist runs therefore miss the old 4-iteration promise enforcement and build-result learning attribution. Please update the hook/config/attribution path for dynamically selected implementation agents, or route specialists through a registered wrapper before replacing the hardcoded implementation agent.

.claude/skills/implementation-self-check/SKILL.md
```diff
@@ -0,0 +1,80 @@
...
+```text
+BLOCKED:
+- [describe what is blocked]
+- [what information, dependency, or decision is needed]
+```
+
+Then emit:
+
+```text
+<promise>IMPLEMENTATION_VERIFIED</promise>
```
shafty023, line 79: The blocked path still emits the verified promise. Combined with Phase 3 accepting either `IMPLEMENTATION_VERIFIED` or `BLOCKED` before marking the task complete, a blocked specialist can cause an unfinished task to be checked off. Blocked output should not include `IMPLEMENTATION_VERIFIED`, and the Phase 3 prompt should leave the task pending or hard-stop on `BLOCKED` instead of proceeding to completion.

.claude/agents/python-script-reviewer.md
```diff
@@ -3,11 +3,50 @@ name: python-script-reviewer
...
+## Execution Modes
+
+This agent supports two execution modes:
+
+1. **Critic Mode (default)**: Review Python scripts for type safety, PEP-8 compliance, error handling, and security
+2. **Implementation Mode**: Implement Python scripts, tools, and utilities using domain expertise
+
+### Mode Detection
+
+**Implementation mode** if prompt contains: `WORKDIR=`, `Implement task`, `Missing requirements:`
```
shafty023, line 19: This mode detection is ambiguous across the new dual-mode agents. Critic/test prompts commonly include `WORKDIR=`, so critic-mode invocations can match implementation mode; meanwhile the Phase 5.5 drift prompt uses `Implement missing behavioral requirement` and `Missing:` rather than the exact `Implement task` / `Missing requirements:` shape. Please use an explicit `MODE=critic|implementation` marker, or one shared prompt envelope, across the prompt routes and all dual-mode agents.

plugins/code/tools/python/discover_available_agents.py
```diff
@@ -0,0 +1,211 @@
...
+        if (plugin_root / ".claude-plugin" / "plugin.json").is_file():
+            yield plugin_root
+
+
+def _iter_cached_plugin_roots(plugin_cache_root: Path) -> Iterable[Path]:
+    if not plugin_cache_root.is_dir():
+        return
+
+    seen: set[Path] = set()
+    for manifest_path in sorted(plugin_cache_root.rglob("plugin.json")):
```
shafty023: Walking the whole plugin cache with `rglob()` and then keeping the first duplicate invocation can surface stale cached plugin versions and makes discovery sensitive to oversized or deeply nested cache trees. The repo already has cache-resolution code that chooses the latest semver plugin version. Please limit traversal to the expected cache layout, choose the active/latest version before deduping agents, and add bounds/oversized-cache coverage.

mikeangstadt: Done in 5d02a2c. Replaced the `rglob("plugin.json")` whole-tree walk with a bounded `<owner>/<plugin>/<version>/.claude-plugin/plugin.json` descent. `_latest_version_root()` now picks the highest semver version dir (reusing the same digit-prefixed parse the cache resolver uses) *before* deduping, so stale cached versions are never surfaced, and the fixed three-level traversal bounds it regardless of how deep/oversized the cache tree is. Added `test_selects_latest_cached_version_and_ignores_stale_versions` and `test_ignores_non_version_and_manifestless_cache_entries` for version selection + junk/manifestless bounds.

plugins/code/prompts/prompt.md
mikeangstadt: Done in dae1af3. Discovery now emits a structured JSON capability record per agent — `implementation_capable`, `file_patterns`, `domains`, `trust_source`, `fallback_rank` — instead of the freeform `invocation | description | tools` line, and Phase 3 consumes the typed fields rather than pattern-matching the tools string.

Addressing each point:
- **Selects non-contract write-capable agents:** `implementation_capable` is derived from the shared `implementation-self-check` skill marker (the agent that honors the `IMPLEMENTATION_VERIFIED` contract + four gates), not from "tools mention Write/Edit". Selection step 1 now considers *only* `implementation_capable: true` agents, so write-capable plan/build/review agents are excluded.
- **Excludes `tools: inherited`:** because eligibility keys on the skill marker, not the tool string, contract-honoring agents rendered as `tools: inherited` are now correctly included (covered by `test_emits_structured_implementation_capability_record`).
- **PLN-626 file/domain mapping:** records carry `file_patterns` and `domains`; selection step 2 globs the task `files:` against `file_patterns` then matches `domains`/`description`, with `fallback_rank`/`trust_source` driving generalist fallback.

PR #100, opened 2026-05-27
feat(code-review): foundation — canonical schema + envelope + budget arbitration + run-plan (PLN-719)

wongk:
## Summary

Lands **PLN-719 Foundation Phase A** (and most of B/C/D) — the shared infrastructure the seven follow-on plans (01–07) depend on. After this merges, plans 01–07 become content additions to a stable system instead of each one re-spec'ing it.

**What ships:**
- Canonical `Finding` + `ResultEnvelope` schema (`schema_version: 1`) with three finding scopes (`diff` / `system` / `pr_metadata`) and the canonical `system_marker` enum.
- New helper subcommands: `finalize-result`, `arbitrate-budget`, `prepare-run`.
- Deterministic finding ids (`<reviewer>_f<index>`) + provenance + structured evidence.
- Verdict precedence rewrite (`APPROVED` | `NEEDS_ATTENTION` | `CHANGES_REQUESTED`); legacy `<pr_verdict>` tag preserved for `run-loop.sh` consumers.
- Canonical `prompt_hash` folds `schema_version` so a MAJOR schema bump invalidates every cache namespace.
- 30-stage declarative `run_plan.json` (`prepare-run`), with stages from plans 01/03/05/06 reserved as `enabled: false` until those plans land.
- Determinism tier vocabulary (`deterministic` / `reproducible_via_cache` / `llm_driven`) + stage tier mapping.
- `/fix` skill now prefers `review_result.json` and surfaces system-scoped findings as manual-surface items.
- `SCHEMA.md` canonical reference + README updates.
- Plugin version `1.5.5 → 2.0.0` (MAJOR — schema-version bump + finding-contract break).

**What's intentionally deferred** (follow-up PRs, per PLN-719's rollout):
- **Phase 4b:** `start.md` orchestrator rewrite (1,264 → ~200 lines walking `run_plan.json`). Largest risk surface — isolated from this PR. The new subcommands are present but **not yet called by the live pipeline**.
- **Phase 5:** Production move of `extract-patches` after `parse-diff` — encoded in `run_plan.json` today; orchestrator move ships with 4b.
- **Phase 7:** Refactor existing BHA cache to the canonical `prompt_hash` recipe + per-namespace TTLs. (The canonical helper landed; the refactor is the follow-up.)
- **Phase 8:** Golden fixture harness + 9 required fixtures.
- **Phase 9:** Per-stage telemetry wiring (depends on 4b).
- **Phase B:** Delete legacy `validate_output.json` / `verified_output.json` emission (coordinated cross-plugin release with `run-loop.sh` + `/fix`).

**Cross-plugin compatibility (Phase A scope):**
- `code` plugin's `run-loop.sh`: no change required. `cmd_verdict` still emits the legacy `verdict` field (`approve|needs_attention|decline`) alongside `canonical_verdict`.
- symphony-alpha's `claude-code-review.yml`: audited — fully compatible. The three handoff files (`code-review-findings.json`, `-threads.json`, `-summary.md`) still get written by the existing `start.md`.

## Test plan

- [x] Unit tests: 369 code-review tests pass (47 new schema tests, 31 new integration tests, 282 pre-existing untouched, 9 review-feedback regression tests).
- [x] Global pytest sweep: 966 plugin tests pass.
- [x] `ruff check .` clean.
- [x] `pyright` clean on production code (0 errors).
- [x] Hygiene check clean on changed files.
- [ ] **Reviewer to run:** `/start --hygiene-only` against any local diff (exercises `setup → parse-diff → hygiene → collect-findings → validate → verdict → present` end-to-end without spawning agents).
- [ ] **Reviewer to run:** `/code-review:start ultra` on this PR (dogfood v2.0.0 helpers against a real diff before merge).
- [ ] **After merge:** confirm `code-review v2.0.0` publishes to the closedloop-ai marketplace; first review on any consumer (e.g. symphony-alpha) will be a one-time cache miss (intentional — MAJOR bump).

## Review feedback addressed

Prior review pass on this branch (4 findings → all fixed in commit `6802f39`):
- **HIGH** — `cmd_collect_findings` no longer drops the whole agent file when an LLM emits a non-canonical reviewer string. Sanitization helper + per-finding try/except + regression test.
- **HIGH** — Test factories extracted to `plugins/code-review/tools/python/conftest.py`. Both test files import from it.
- **MEDIUM** — Removed unreachable second BLOCKING loop in `_compute_canonical_verdict`.
- **MEDIUM** — CLAUDE.md updated to document the shared-library exception for `code_review_schema.py`.
- **Hygiene** — `/home/runner` test fixture now built at runtime so the literal pattern doesn't appear in source.

## Refs

- ClosedLoop: [PRD-409](https://app.closedloop.ai/closedloop-ai/prds/PRD-409) → [FEA-1395](https://app.closedloop.ai/closedloop-ai/features/FEA-1395) → [PLN-719](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-719)
- Loop: `019e6abf-439a-702e-937d-9910e6bdde84`
- Schema reference: [plugins/code-review/SCHEMA.md](plugins/code-review/SCHEMA.md)


plugins/code-review/tools/python/code_review_schema.py
```diff
@@ -0,0 +1,732 @@
...
+# ---------------------------------------------------------------------------
+
+
+def normalize_legacy_finding(
+    raw: dict[str, Any],
+    *,
+    reviewer: str = "unknown",
+    source: str = "agent",
+    index: int = 0,
+    emitted_at: str = "",
```
mikeangstadt, line 585: Nice backward-compat pattern using `setdefault` — fills canonical fields without overwriting existing values. The shallow copy + `setdefault` cascade means agents already emitting schema_version, id, or reviewer just pass through untouched. This will make the Phase A → B migration smooth when legacy producers are retired.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,49 +3402,942 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+    required: list[dict[str, Any]] = list(coverage_plan_in.get("required", []) or [])
+    best_effort: list[dict[str, Any]] = list(coverage_plan_in.get("best_effort", []) or [])
+    deprecation_warnings: list[str] = list(coverage_plan_in.get("deprecation_warnings", []) or [])
+
+    bha_floor = 0 if _is_docs_only(diff_data) else BUDGET_BHA_FLOOR_DEFAULT
+    max_bha = _max_bha_partitions_by_loc(diff_data)
+
+    now_iso = datetime.now(timezone.utc).isoformat()
+    coverage_gaps: list[dict[str, Any]] = []
+    dropped_required: list[dict[str, Any]] = []
```
mikeangstadt, line 4221: The `f not in coverage_gaps` filter relies on dict equality, which works here since both lists reference objects from the same `validated` list. If this legacy fallback path ever merges findings from multiple sources, dict equality could match incorrectly. Using `gap_ids = {f.get("id") for f in coverage_gaps}` and `f.get("id") not in gap_ids` would be a more robust guard for future refactors.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,49 +3402,942 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+        "telemetry": {
+            "expected_total_duration_ms": 0,
+            "estimated_cost_usd": 0.0,
+        },
+    }
+
+    output_path = Path(args.output) if args.output else Path(cr_dir) / "run_plan.json"
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    with open(output_path, "w") as f:
+        json.dump(run_plan, f, indent=2)
```
mikeangstadt, line 4103: Clean forward-compat design — Rules 4–6 are documented as inert placeholders for plans 02/03/06, so those plans only need to update `_compute_canonical_verdict` without touching the verdict subcommand wiring. The docstring makes the intent explicit, which is exactly the kind of contract you want in a foundation PR.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,49 +3402,942 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+                "input_uncached": 0,
+                "input_cached": 0,
+                "output": 0,
+                "by_model": {},
+            },
+            "schema_versions_seen": {"finding": SCHEMA_VERSION, "result": SCHEMA_VERSION},
+        },
+    }
+
+    errors = validate_result_envelope(envelope)
```
thadeusb, line 4499: You validate the envelope, capture errors, then write it to disk and return 0 regardless. When the orchestrator starts walking run_plan.json (Phase 4b), stage_25 has on_failure: "abort", but this function never returns non-zero, so a malformed envelope will silently flow into verdict and present. At minimum return 1 when errors is non-empty, or at least log the errors to stderr so operators have a signal.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,49 +3402,942 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+            "id": "stage_16_arbitrate_budget",
+            "kind": "helper",
+            "subcommand": "arbitrate-budget",
+            "args": [
+                "--coverage-plan", f"{cr_dir}/coverage_plan_initial.json",
+                "--diff-data", f"{cr_dir}/diff_data.json",
+                "--output", f"{cr_dir}/coverage_plan.json",
+            ],
+            "expected_outputs": [f"{cr_dir}/coverage_plan.json", f"{cr_dir}/coverage_gaps.json"],
+            "depends_on": ["stage_15_coverage_critic"],
```
thadeusb, line 3804: stage_16_arbitrate_budget is enabled: True but depends_on stage_15_coverage_critic which is enabled: False. Its --coverage-plan arg points at coverage_plan_initial.json, which is the output of disabled stage_14. When the orchestrator actually walks this plan, arbitrate-budget will immediately error because its input file won't exist. Should this depend on parse-diff (stage_05) instead, with the coverage_plan input falling back to a default empty plan?

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,49 +3402,942 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+        },
+    ]
+
+
+def cmd_prepare_run(args: argparse.Namespace) -> int:
+    """Emit ``run_plan.json`` describing the full review pipeline.
+
+    PLN-719 Section 6. The output is consumed by the orchestrator (a future
+    rewrite of start.md will walk this declarative plan). Determinism: same
+    inputs produce byte-identical output.
```
thadeusb: Docstring says "same inputs produce byte-identical output" but the function calls uuid.uuid4() on line 3932. The CHANGELOG correctly qualifies this with "modulo the review_id uuid" but the docstring doesn't. Small thing but someone will read this and assume determinism for diffing or caching, then waste time debugging why outputs differ.

thadeusb (approved):
Solid foundation work. One thing to watch: finalize-result silently writes invalid envelopes, left a couple comments.

plugins/code-review/tools/python/code_review_schema.py
wongk: Thanks! Agreed — the `setdefault` cascade was deliberate so canonical-emitting agents land untouched. The Phase B retirement of legacy producers is now just "delete the legacy emission path" with no normalization gymnastics.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Thanks. The verdict subcommand intentionally stays dumb: it reads canonical `verdict` + `verdict_reason` from the envelope and maps to the legacy tag. All precedence logic lives in `_compute_canonical_verdict` so plans 02/03/06 just extend the function — no wiring changes here.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good call. Fixed in 80a51ab — both `cmd_verdict`'s legacy fallback and `cmd_finalize_result` now partition the validated list by enumerate-index instead of dict-equality membership. Robust against future multi-source merges.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Caught it — fixed in 80a51ab. `cmd_finalize_result` now prints validation errors to stderr and returns exit code 1 when the envelope fails schema validation. The file is still written for operator inspection; the non-zero exit gives stage_25's `on_failure: "abort"` the signal it needs when the orchestrator walks the plan in Phase 4b.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Right — `coverage_plan_initial.json` is the output of stage_14_resolve_coverage which is plan-05-gated. Fixed in 80a51ab by flipping stage_16_arbitrate_budget to `enabled: false` with an inline comment explaining it ungates together with stages 11/14/15 when plan 05 ships. The subcommand itself remains callable today — only the orchestrated stage is gated. Added `test_arbitrate_budget_gated_on_plan_05` to lock this in.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fair point — fixed in 80a51ab. Docstring now reads: "same inputs produce byte-identical output **except for the `review_id` field**, which is a fresh `uuid.uuid4()` per invocation. Compare via a JSON pop of `review_id` before diffing or hashing". Matches the CHANGELOG wording.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,56 +3402,1118 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+
+    # Promote findings to canonical schema (defensive — collect-findings
+    # should already have done this, but legacy producers may bypass it).
+    now_iso = datetime.now(timezone.utc).isoformat()
+    canonical_findings: list[dict[str, Any]] = []
+    for idx, raw in enumerate(validated):
+        if not isinstance(raw, dict):
+            continue
+        canonical_findings.append(
+            normalize_legacy_finding(
```
shafty023: **[HIGH] [P1] cmd_finalize_result crashes on non-canonical reviewer strings with no pre-assigned id**

**Issue:** cmd_finalize_result calls normalize_legacy_finding(raw, reviewer=raw.get('reviewer', 'unknown'), ...) for every finding in validated[]. If a finding has no 'id' field and its 'reviewer' field is a non-canonical string such as 'Bug Hunter A' (spaces or uppercase), normalize_legacy_finding will call make_finding_id(out['reviewer'], index), which raises ValueError because 'Bug Hunter A' does not match ^[a-z][a-z0-9_-]*$. This exception is not caught anywhere in cmd_finalize_result, causing an unhandled traceback and a non-zero exit. cmd_collect_findings guards against this via _coerce_reviewer_id, but cmd_finalize_result's defensive normalization path (the comment explicitly states 'legacy producers may bypass [collect-findings]') has no equivalent guard, so the exact failure mode the guard was written to prevent can still occur here.

**Recommendation:** Apply the same _coerce_reviewer_id guard in cmd_finalize_result before calling normalize_legacy_finding, or wrap the normalize_legacy_finding call in a try/except (ValueError, TypeError) block that logs a warning and skips the malformed finding rather than crashing the whole command.

<sub>Category: Correctness · Confidence: 0.78</sub>

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3208,56 +3402,1118 @@ def cmd_collect_findings(args: argparse.Namespace) -> int:
...
+            "enabled": False,
+        },
+        {
+            "id": "stage_17_partition",
+            "kind": "helper",
+            "subcommand": "partition",
+            "args": ["--diff-data", f"{cr_dir}/diff_data.json"],
+            "stdout": f"{cr_dir}/partitions.json",
+            "expected_outputs": [f"{cr_dir}/partitions.json"],
+            "depends_on": ["stage_16_arbitrate_budget"],
```
shafty023: **[HIGH] [P1] stage_17_partition depends on disabled stage_16_arbitrate_budget in run_plan.json**

**Issue:** In `_build_run_plan_stages`, `stage_17_partition` has `depends_on: ["stage_16_arbitrate_budget"]` while `stage_16_arbitrate_budget` is `enabled: False`. The test `test_foundation_stages_enabled` asserts that `stage_17_partition` is `enabled: True` and must run in Phase A. But a dependency-aware orchestrator that walks `run_plan.json` would find that stage_17's sole declared prerequisite is disabled (never completed), and would either skip partition or block indefinitely. No test verifies that the dependency chain from stage_16 through stage_17 is executable when stage_16 is disabled. The comment in the stage definition says stage_16 flips to True together with plan 05 stages, but partition is a foundation stage that must run unconditionally in Phase A. The produced `run_plan.json` therefore contains a structurally unsatisfiable dependency for any orchestrator that enforces the `depends_on` graph.

**Recommendation:** Change `stage_17_partition` to depend on `stage_15_coverage_critic` instead of `stage_16_arbitrate_budget` (the same approach plan 05 uses for the disabled chain), or add a bypass rule: when a dependency stage is disabled, skip it in the dependency check so execution proceeds to the next enabled stage. Either way, the test `test_arbitrate_budget_gated_on_plan_05` should also assert that `stage_17_partition`'s `depends_on` does not reference any disabled stage.

<sub>Category: Correctness · Confidence: 0.90</sub>

shafty023 (dismissed):
done with review

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good catch — fixed in f3ecd9e. `cmd_finalize_result` now mirrors `cmd_collect_findings`: `_coerce_reviewer_id` overwrites the reviewer field with the `unknown` fallback when the LLM emits a non-canonical string like "Bug Hunter A", and the `normalize_legacy_finding` call is wrapped in `try/except (ValueError, TypeError)` so a single malformed finding logs a warning and is skipped rather than aborting the whole envelope. Added `TestFinalizeResult.test_non_canonical_reviewer_string_does_not_crash` to cover the exact regression you described.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Valid — fixed in f3ecd9e. `stage_17_partition` now declares `depends_on: ["stage_05_parse_diff"]`, which is its actual data dependency (it only reads `diff_data.json`). The earlier `stage_16_arbitrate_budget` dep was a sequencing placeholder that became unsatisfiable when 80a51ab flipped stage_16 to `enabled: false`. I also added a universal regression test — `test_enabled_stages_do_not_depend_on_disabled_stages` — that walks every stage in the produced plan and asserts no enabled stage names a disabled stage in its `depends_on`. That'll catch this class of mistake on any future stage, not just stage_17.

PR #103, opened 2026-05-28
fix(code-review): add "Code Quality" to CATEGORIES enum

wongk:
## Summary

Hotfix for a production warning surfaced during a code review:

> ⚠️ Note: finalize-result rejected the canonical envelope because the existing finding uses category: \"Code Quality\", which is not in the schema's allowed enum (Correctness | Coverage | Hygiene | … | TestQuality). Verdict fell back to validate_output.json as designed.

### Root cause

The shared reviewer prompt at [shared_prompt.txt:74](plugins/code-review/tools/prompts/shared_prompt.txt#L74) documents \`\"category\": \"Code Quality\"\` as the canonical example for MEDIUM-tier DRY/maintainability findings — but the schema enum at [code_review_schema.py:49](plugins/code-review/tools/python/code_review_schema.py#L49) omitted it. Both are foundation-owned and they disagreed.

When a reviewer emitted a Code Quality finding:
1. \`validate_finding\` flagged the category as unknown.
2. \`finalize-result\` recorded the validation error but still wrote the envelope (for operator inspection) and returned exit code 1.
3. \`cmd_verdict\` fell back to \`validate_output.json\`, which doesn't apply the canonical category gate — so the verdict computed correctly, but the canonical envelope path silently dropped the finding.

### Fix

Smallest correct change: add \`\"Code Quality\"\` to \`CATEGORIES\`. The reviewer prompt's choice was semantically right — DRY/maintainability isn't a Correctness bug — and the enum was the thing that drifted. \`SCHEMA.md\` line 40 (which mirrors the enum verbatim) is updated to match.

Considered + rejected: mapping Code Quality → Correctness in \`normalize_legacy_finding\`. That would have collapsed two legitimately distinct semantic categories and hidden the contract violation behind a normalizer.

### Tests (+3)

- \`test_categories_include_code_quality\`
- \`test_code_quality_finding_passes_validation\`
- \`test_code_quality_finding_in_envelope_passes_validation\`

### Version

2.1.0 → **2.1.1** (PATCH — additive enum entry, fixes a real production warning).

383 tests pass (was 380). ruff + uv pyright clean.

## Test plan
- [ ] CI green
- [ ] Spot-check: run \`/start\` on a small diff where a reviewer emits a Code Quality finding; verify no warning surfaces and the envelope path retains the finding

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #102, opened 2026-05-28
feat(code-review): Phase 9 — canonical telemetry schema + finalize-result wiring (PLN-719)

wongk:
## Summary

PLN-719 Phase 9 (Telemetry). Defines the canonical per-run metrics surface on \`review_result.json.telemetry\`, tightens the v2.0.0 open-object stub into a typed contract, and wires \`finalize-result\` to merge any optional \`<cr_dir>/telemetry.json\` produced upstream.

The actual end-to-end wiring (per-stage timestamps + cache hit/miss plumbing) lands in Phase 4b/7. This PR ships the **contract** so producers have a stable target — same Foundation pattern as PR #100.

### Schema (\`code_review_schema.py\`)
- \`empty_telemetry()\` returns the zero-valued block.
- \`validate_telemetry(t)\` enforces required keys, non-negative numerics, \`cache_hit_rate\` in \`[0, 1]\`, typed tokens, \`schema_versions_seen\` as int map. Forward-compat: unknown keys permitted.
- \`merge_telemetry(base, overlay)\` deep-merges one level for known nested objects (e.g. \`tokens\`), so callers can populate \`tokens.input_uncached\` without overriding the whole block.
- \`result_envelope_json_schema()\` declares \`telemetry\` as a typed object with required keys (was: open \`{type: \"object\"}\`).
- \`validate_result_envelope\` recurses into telemetry when present.
- Canonical cache namespace constants — \`CACHE_NAMESPACES = {bha, signals, coverage_critic, verifications, overrides}\` per PLN-719 §9.

### Helper (\`code_review_helpers.py\`)
- \`_build_telemetry_block(cr_dir)\` reads optional \`<cr_dir>/telemetry.json\`, deep-merges over \`empty_telemetry()\`, and always overwrites \`schema_versions_seen\` so upstream cannot spoof the version stamp.
- \`cmd_finalize_result\` uses the helper instead of an inline stub.
- \`validate_result_envelope\` was at SonarLint complexity 38; extracted per-section validators (\`_validate_envelope_scalars\`, \`_validate_envelope_buckets\`, \`_validate_coverage_plan\`, \`_validate_envelope_findings\`) bring it under the threshold. Same coverage; flatter call graph.

### Tests (+16)
Schema: empty/required/range/by_model/forward-compat/deep-merge/JSON Schema declaration. Finalize integration: defaults when no telemetry.json; deep-merge into envelope; \`schema_versions_seen\` cannot be spoofed; malformed telemetry.json degrades gracefully.

[...]

PR #105, opened 2026-05-28
feat(code-review): Phase 8 — golden fixture harness + 3 fixtures + prepare-run determinism (PLN-719)

wongk:
## Summary

PLN-719 Phase 8 (Golden Fixture Harness). Ships the harness infrastructure plus the 3 fixtures whose expected envelopes don't depend on unshipped plans. The other 6 fixtures (PLN-719 §10) get reserved directories with READMEs and skip until their dependent plans land.

### Harness (\`golden_fixture_harness.py\`)
- \`GoldenFixture.load(path)\` reads \`config.yaml\`.
- \`run_post_collection_pipeline(cr_dir, fixture)\` stages \`inputs/\`, runs \`collect-findings\` → \`validate\` → \`finalize-result\`, returns the envelope, and re-validates via \`validate_result_envelope\`. Every fixture doubles as a schema round-trip check (§10 acceptance criterion: \"every fixture round-trips emit → write → read → validate\").
- \`normalize_envelope(envelope)\` scrubs non-deterministic fields (\`review_id\` uuid, \`emitted_at\` timestamps, the wall-clock telemetry block) so expected/ artifacts stay stable.
- \`diff_envelope_against_expected\` produces structural diffs with dotted paths so failures read cleanly.
- \`update_expected\` writes through the same path the assertion uses, so \`--update-golden\` followed by no-flag produces byte-identical output.

### Pytest integration
- New \`--update-golden\` CLI option in \`conftest.py\` rewrites expected/ artifacts in place for human review.
- \`test_golden_fixture\` is parametrized over every \`golden_*\` directory under \`fixtures/\`; deferred fixtures skip with reasons citing the unshipped plan.
- \`test_prepare_run_produces_byte_identical_output_modulo_review_id\` pins §6 determinism — two \`prepare-run\` invocations differ only in \`review_id\`.

### Fixtures shipped end-to-end
- \`golden_minimal_correctness\` — single HIGH Correctness finding from a BHA agent, verdict NEEDS_ATTENTION.
- \`golden_all_categories\` — four findings spanning Correctness, Code Quality, Security, TestQuality (verifies the post-PR-#103 CATEGORIES enum flows through finalize's \`by_category\` stats).
- \`golden_schema_v1_round_trip\` — single Security finding with every optional schema field populated (\`evidence[]\`, \`reasoning_certificate\`, \`other_locations\`, \`subcategory\`) — the maximal v1 envelope.

### Deferred fixtures (reserved dirs + READMEs)
- \`golden_premise_justified\`, \`golden_premise_rejected\` → plan 02
- \`golden_impact_with_callsites\` → plan 06
- \`golden_coverage_gap\` → plans 03 + 05
- \`golden_injection_quarantine\` → plan 01
- \`golden_budget_exceeded\` → arbitrate-budget integration

### Docs
SCHEMA.md §12 documents the harness contract (layout, \`--update-golden\` workflow, Phase 8 vs deferred scope, Phase 4b extension path).

### Out of scope (Phase 4b follow-up)
Walking \`run_plan.json\` end-to-end through a declarative stage runner. The post-collection harness is the minimum that ships value without depending on the unshipped Phase 4b orchestrator rewrite.

### Version
2.3.0 → **2.4.0** (MINOR — additive test infrastructure + new fixtures + new pytest option; no runtime API change).

427 tests pass (was 421), 6 skipped (deferred fixtures). ruff + uv pyright clean.

## Test plan
- [ ] CI green
- [ ] Spot-check: run \`pytest plugins/code-review/tools/python/test_golden_fixtures.py -v\` locally — expect 3 PASSED + 6 SKIPPED + 3 sanity tests
- [ ] Spot-check: \`pytest --update-golden\` followed by \`pytest\` produces no diff


plugins/code-review/tools/python/golden_fixture_harness.py
```diff
@@ -0,0 +1,238 @@
...
+        cr_dir=str(cr_dir),
+        output="findings.json",
+        hygiene=str(hygiene_path) if hygiene_path.exists() else None,
+    )
+    _run(cmd_collect_findings, collect_ns)
+
+    validate_ns = argparse.Namespace(
+        findings=str(findings_path),
+        diff_data=str(diff_data_path),
+        scope_kind=fixture.config.get("scope_kind"),
```
thadeusb: cmd_validate never reads scope_kind, and none of the config.yaml files set it either, so this is always None feeding a param nothing consumes. Harmless today, but it reads like the harness threads scope through validate when it doesn't. Drop it, or it'll bite whoever adds a pr_metadata fixture and assumes this does something.

plugins/code-review/tools/python/golden_fixture_harness.py
```diff
@@ -0,0 +1,238 @@
...
+        )
+    return envelope
+
+
+def diff_envelope_against_expected(
+    fixture: GoldenFixture, actual: dict[str, Any],
+) -> list[str]:
+    """Return a list of mismatch lines comparing actual vs expected envelope.
+
+    Both sides are normalized first (review_id, emitted_at, telemetry).
```
thadeusb: docstring says both sides are normalized first, but you only normalize actual. expected is read straight off disk. It works because expected always comes from update_expected which runs normalize_envelope, but if someone hand-drops a real envelope into expected/ with a live emitted_at or telemetry block the diff will report spurious mismatches instead of scrubbing them. Either fix the comment or normalize expected too.

thadeusb (approved):
Harness is solid, ran it locally and all 6 active fixtures pass plus the update-golden round-trip is byte-stable. Couple of minor things on the validate wiring and a stale docstring.

wongk:
Pushed `7c816af` addressing all 7 review findings. Version stays at 2.4.0 (folded into the open PR, consistent with how PR #103 was folded into PR #102).

**Per-finding response:**

1. **(auditor) Dead `scope_kind` in `validate_ns`** — ✅ Fixed. Removed the dead arg from `golden_fixture_harness.py`. `cmd_validate` only reads `--findings` and `--diff-data`, so the Namespace attribute was silently ignored.

2. **(auditor) Unread `expected_verdict` / `expected_verified_count` / `expected_coverage_gap_count`** — ✅ Fixed. Wired all three into hard assertions in `test_golden_fixture()`. They run even in `--update-golden` mode so the rewriter cannot silently pin a verdict that contradicts config intent. Added the missing count keys to `golden_all_categories` and `golden_schema_v1_round_trip`. SCHEMA.md §12 updated. Verified by temporarily corrupting `expected_verdict: CHANGES_REQUESTED` and confirming the assertion fires before restoring.

3. **(auditor) `_invoke` duplicates `TestPrepareRun._run`** — ✅ Fixed. Added `invoke_prepare_run(cr_dir, *, output=None, ...)` to `conftest.py`. Both call sites now delegate.

4. **(bhb) Inline stdout-suppression duplicates harness `_run`** — ✅ Fixed. Hoisted `_run` to module-level `run_with_stdout_capture(fn, ns, *, stdout_to=None)` in `golden_fixture_harness.py`. The harness's three internal stage invocations and the conftest helper all delegate to it.

5. **(bhb) Asymmetric normalization docstring** — ✅ Fixed. Rewrote the docstring per the recommendation: only `actual` is normalized; the expected file is compared as-is (already written through `update_expected`'s normalization path).

6. **(bha_p2) "every CATEGORIES value" inaccurate** — ✅ Fixed. Rewrote the header comment + `description` in `golden_all_categories/config.yaml` to reflect actual coverage (4 of 11 canonical categories). Noted that the remaining categories are owned by the deferred fixtures.

7. **(bha_p2) `setup.current_branch` contradicts `scope.review_branch`** — ✅ Fixed. Set `current_branch: "feature/x"` in both `golden_minimal_correctness` and `golden_all_categories` setup.json files (matching the consistent shape in `golden_schema_v1_round_trip`). Regenerated `expected/review_result.json` for both fixtures to pin the corrected `review_branch`.

**Test stats:** 1024 passed, 6 skipped. ruff + pyright clean.

**Version note:** stays at 2.4.0 since these fold into the open Phase 8 PR. CHANGELOG entries merged into the existing v2.4.0 section under `#### Added` / `#### Changed` / `#### Fixed` subheadings.

plugins/code-review/tools/python/golden_fixture_harness.py
wongk: Resolved in 7c816af. Removed the dead `scope_kind=fixture.config.get("scope_kind")` from `validate_ns` construction. If scope-aware validation is added later, it'll be wired explicitly when `cmd_validate` gains that capability.

plugins/code-review/tools/python/golden_fixture_harness.py
wongk: Resolved in 7c816af. Updated the docstring to state that only `actual` is normalized and that the expected file is compared as-is (it's already written through `update_expected`'s normalization path, so hand-editing an expected file with a real UUID now surfaces as a diff rather than being silently scrubbed at read time).

PR #106, opened 2026-05-28
feat(code): harness adapter interface, dataclasses, and registry (PLN-751)

aponamarev:
## Summary

PLN-751 foundation feature. Establishes the harness-agnostic seam every later workstream plugs into: a stable Python contract that the bash skeleton in `plugins/code/scripts/run-loop.sh` can call instead of speaking to any specific harness (`claude`) directly.

No concrete adapter implementations land here — only the contract, registry, and CLI plumbing, exercised by a trivial in-test fake adapter to prove the wiring. New package: `plugins/code/tools/python/harness/`.

### Types (`harness/types.py`)
- **`Command` StrEnum** — five members (`PLAN_EXECUTE`, `PROCESS_LEARNINGS`, `EXPORT_LEARNINGS`, `CODE_REVIEW_START`, `CODE_REVIEW_FIX`). Each request types `command` as a distinct enum-literal so pyright narrows the concrete type from a `match request.command:` block — a new member without a `case` is type-checker-visible (exhaustiveness).
- **Discriminated union of requests** — five frozen, `kw_only=True` dataclasses (`PlanExecuteRequest`, `ProcessLearningsRequest`, `ExportLearningsRequest`, `CodeReviewStartRequest` with typed `base_sha`, `CodeReviewFixRequest` with typed `cr_dir`) sharing a `_BaseRequest` (`workdir`, `prompt`). `InvocationRequest` is the union alias.
- **Result types** — `TurnResult(result_text, is_error, session_id)` and `TerminalFailure(status, subcode, message)`, both frozen. `TerminalFailure.subcode` is validated at construction time against the regex enforced by `write_loop_user_visible_failure()` at `run-loop.sh:79` (`^[A-Z][A-Z0-9_]{2,63}$`) so a misbuilt failure fails fast instead of corrupting the user-visible marker downstream.
- **JSON boundary** — `request_from_json(payload)` deserializes by dispatching on `Command(payload["command"])` via a `_BY_COMMAND` table; symmetric `turn_result_{to,from}_json` / `terminal_failure_{to,from}_json` helpers round-trip cleanly so bash can shell out, pass JSON in, and parse JSON back.

### Adapter (`harness/adapter.py`)
- `HarnessAdapter` ABC with `name: ClassVar[str]` and six abstract methods: `supports`, `build_entry_prompt`, `build_argv`, `parse_session_id`, `parse_turn_result`, `classify_terminal_failure`. Subclassing without implementing all six raises `TypeError` at instantiation.

### Registry (`harness/registry.py`)
- `register(adapter_cls)` / `get_adapter(name)` keyed on the class-level `name`. Looking up an unregistered name raises a clear, actionable error naming the registered keys.

### Thin CLI (`harness/cli.py`)
- Reads a JSON request from stdin, takes `adapter_name` and `method_name` as positional args, dispatches to the named adapter method via `request_from_json`, and writes JSON to stdout — mirroring the existing `tools/python/` pipeline tools. Side-effect-free except for stdout.

### Non-functional
- Python standard library only (`json`, `re`, `dataclasses`, `enum`, `abc`, `typing`) — no third-party deps, runs without `pip install`. All dataclasses frozen; the data contract is immutable across the boundary.

### Tests (+82)
- `test_harness_types.py` — table-driven coverage of enum values, all request construction/field access, `TerminalFailure` subcode validation (valid accepted, invalid rejected), `TurnResult`, and JSON round-trips. Shared fixtures in `conftest.py`.
- `test_harness_registry.py` — register/get_adapter via a fake adapter; nonexistent lookup raises a clear error; ABC `TypeError` on incomplete subclass.
- `test_harness_cli.py` — full stdin→stdout subprocess round-trip against a registered fake adapter.

### Version
1.11.20 → **1.12.0** (MINOR — new module added; no existing behavior changed).

82 tests pass. ruff + pyright clean (0 errors).

### Out of scope (later PLN-751 workstreams)
- Concrete adapter implementations (Claude, other harnesses).
- `run-loop.sh` modifications, `--harness` flag / env-fallback wiring (PRD Open Question 3).
- `analyze-turn` consolidation (PRD Open Question 1).
- CHANGELOG.md / README.md — to be generated via `/update-documentation` before merge.

## Test plan
- [ ] CI: `Plugin Version Bump`, `Lint`, `Type Check`, `Tests` all green
- [ ] Manual (T-6.1): omit a `Command` member from a `match`/`case` block over `request.command` and confirm pyright reports the missing case
- [ ] Spot-check the CLI round-trip: `echo '{"command":"code_review_start","workdir":"/tmp","prompt":"p","base_sha":"abc"}' | python -m harness.cli fake parse_turn_result`

### DIFF
--- a/plugins/code/tools/python/harness/cli.py
+++ b/plugins/code/tools/python/harness/cli.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import json
 import sys
+from typing import NoReturn
 
 from harness.registry import get_adapter
 from harness.types import (
@@ -80,20 +81,45 @@ def _dispatch(adapter_name: str, method_name: str, payload: dict[str, object]) -
     )
 
 
+def _error_exit(msg: str) -> NoReturn:
+    """Print a one-line error to stderr and exit non-zero.
+
+    Mirrors the ``error_exit`` idiom used by sibling code-plugin tool scripts
+    (e.g. ``count_tokens.py``) so shell callers get a clean message instead of
+    a raw Python traceback. Defined locally because the no-cross-import rule for
+    tool scripts forbids importing that helper directly.
+    """
+    print(f"Error: {msg}", file=sys.stderr)
+    sys.exit(1)
+
+
 def main() -> None:
     """CLI entry point: reads stdin JSON, writes stdout JSON."""
     if len(sys.argv) != 3:
-        print(
-            "Usage: python -m harness.cli <adapter_name> <method_name>",
-            file=sys.stderr,
-        )
-        sys.exit(1)
+        _error_exit("Usage: python -m harness.cli <adapter_name> <method_name>")
 
     adapter_name = sys.argv[1]
     method_name = sys.argv[2]
 
-    payload = json.loads(sys.stdin.read())
-    result = _dispatch(adapter_name, method_name, payload)
+    raw_stdin = sys.stdin.read()
+    if not raw_stdin.strip():
+        _error_exit("No JSON request provided on stdin.")
+
+    try:
+        payload = json.loads(raw_stdin)
+    except json.JSONDecodeError as exc:
+        _error_exit(f"Invalid JSON request on stdin: {exc}")
+
+    if not isinstance(payload, dict):
+        _error_exit("JSON request must be a JSON object.")
+
+    try:
+        result = _dispatch(adapter_name, method_name, payload)
+    except (KeyError, ValueError, TypeError) as exc:
+        # KeyError stringifies with surrounding quotes; unwrap to the bare message.
+        message = exc.args[0] if exc.args else str(exc)
+        _error_exit(str(message))
+
     print(json.dumps(result))
 
 
--- a/plugins/code/tools/python/harness/cli.py
+++ b/plugins/code/tools/python/harness/cli.py
@@ -1,0 +1,101 @@
+"""
+Thin CLI entry point for the harness package.
+
+Usage::
+
+    python -m harness.cli <adapter_name> <method_name> < request.json
+
+Reads a JSON request from stdin, dispatches to the named method on the
+named adapter, and writes a JSON result to stdout. All harness-specific
+behavior is reached through the HarnessAdapter interface.
+
+Supported method names and their stdout JSON shape:
+
+- ``build_entry_prompt``  -> {"result": "<prompt_string>"}
+- ``build_argv``          -> {"argv": [...]}
+- ``parse_session_id``    -> {"session_id": "<id>" | null}
+- ``parse_turn_result``   -> TurnResult JSON (via turn_result_to_json)
+- ``classify_terminal_failure`` -> TerminalFailure JSON (via terminal_failure_to_json)
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+
+from harness.registry import get_adapter
+from harness.types import (
+    request_from_json,
+    terminal_failure_to_json,
+    turn_result_to_json,
+)
+
+
+def _dispatch(adapter_name: str, method_name: str, payload: dict[str, object]) -> dict[str, object]:
+    """Look up the adapter, deserialize the request, call the method.
+
+    Raises KeyError if the adapter or method is not found.
+    """
+    adapter_cls = get_adapter(adapter_name)
+    adapter = adapter_cls()
+
+    if method_name == "build_entry_prompt":
+        raw_add_dirs = payload.get("add_dirs", [])
+        add_dirs = list(raw_add_dirs) if isinstance(raw_add_dirs, list) else []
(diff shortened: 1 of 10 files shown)


plugins/code/tools/python/harness/types.py
```diff
@@ -0,0 +1,186 @@
...
+
+    ``subcode`` must match ``^[A-Z][A-Z0-9_]{2,63}$`` (the same constraint
+    enforced by ``write_loop_user_visible_failure()`` in run-loop.sh line 79).
+    """
+
+    status: str
+    subcode: str
+    message: str
+
+    def __post_init__(self) -> None:
```
thadeusb, line 136: The docstring sells this as validating against 'the same constraint enforced by run-loop.sh' so a misbuilt failure fails fast instead of corrupting the marker. But __post_init__ only checks subcode. run-loop.sh:70-87 also validates the code against a fixed allowlist (RUNNER_ERROR|PRE_RUN_VALIDATION_FAILED|PLAN_STATE_UNAVAILABLE) and the message length (1-1000). So a TerminalFailure(status='EXIT', subcode='NON_ZERO_EXIT', message='') constructs fine here, exactly like your fake adapters build it, then gets rejected when bash tries to write the marker. If status maps to that code field, validate it here too and check message length, otherwise the fail-fast covers one of the three things bash will reject and the other two still blow up downstream.

plugins/code/tools/python/harness/types.py
```diff
@@ -0,0 +1,186 @@
...
+_BY_COMMAND: dict[Command, type] = {
+    Command.PLAN_EXECUTE: PlanExecuteRequest,
+    Command.PROCESS_LEARNINGS: ProcessLearningsRequest,
+    Command.EXPORT_LEARNINGS: ExportLearningsRequest,
+    Command.CODE_REVIEW_START: CodeReviewStartRequest,
+    Command.CODE_REVIEW_FIX: CodeReviewFixRequest,
+}
+
+
+def request_from_json(payload: dict[str, object]) -> InvocationRequest:
```
thadeusb, line 168: cls(**kwargs) throws TypeError on any unexpected key in the payload, but every other branch in cli._dispatch pulls fields with .get() and ignores extras. So if the bash side sends one payload shape and just varies the method name, build_argv breaks on a stray key while parse_turn_result silently ignores it. Pick one boundary behavior, this asymmetry will be a confusing failure to debug.

plugins/code/tools/python/harness/cli.py
```diff
@@ -0,0 +1,101 @@
...
+- ``parse_turn_result``   -> TurnResult JSON (via turn_result_to_json)
+- ``classify_terminal_failure`` -> TerminalFailure JSON (via terminal_failure_to_json)
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+
+from harness.registry import get_adapter
```
thadeusb, line 27: These absolute harness.* imports only resolve when tools/python is on sys.path. run-loop.sh invokes the other python tools by direct file path ($tools_dir/foo.py), and that won't work here since `python .../harness/cli.py` leaves harness un-importable. Whoever wires this into the loop needs PYTHONPATH set or `-m` from the right cwd. Fine for a foundation PR, just flagging so it doesn't surprise the next workstream.

thadeusb (approved):
Clean seam and the tests cover it well. Main thing: TerminalFailure only validates subcode, but bash rejects bad status and message too, so the fail-fast doesn't actually cover what it claims. See comments.

plugins/code/tools/python/harness/types.py
aponamarev: Added full validation to `TerminalFailure.__post_init__` (lines 136-152): status against the allowlist {RUNNER_ERROR, PRE_RUN_VALIDATION_FAILED, PLAN_STATE_UNAVAILABLE}, subcode regex, and message length (1-1000 chars). Docstring updated to reflect all three constraints. Fixes committed in 5be8c0f.

plugins/code/tools/python/harness/types.py
aponamarev: The confusing-traceback harm is resolved: `cli.py:main()` now catches (KeyError, ValueError, TypeError) and routes through `_error_exit` (lines 116-121), so a TypeError from unexpected keys produces a clean error message on stderr with non-zero exit, not a raw traceback (commit 80bc1a9). The strict-vs-lenient asymmetry in request deserialization is by design.

plugins/code/tools/python/harness/cli.py
aponamarev: The module docstring documents the correct invocation as `python -m harness.cli` (line 6), which resolves the absolute import path constraint. This is deferred to the run-helper workstream that will wire this into run-loop.sh; no change needed in this PR.

PR #109, opened 2026-05-29
feat(code-review): PLN-720 — Prompt-Injection Defense (cmd_detect_injection + quarantine)

wongk:
## Summary

First feature plan downstream of the PLN-719 foundation (PR #107). Implements PLN-720 (Prompt-Injection Defense) by filling in the implementation behind the foundation's pre-stubbed `stage_09_detect_injection` and flipping it to enabled.

- **Refs:** [PLN-720](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-720), [FEA-1396](https://app.closedloop.ai/closedloop-ai/features/FEA-1396), [PRD-409](https://app.closedloop.ai/closedloop-ai/prds/PRD-409) row 2
- **Version:** code-review 2.6.4 → 2.7.0 (MINOR)
- **Tests:** 470 pass (27 new), 5 skipped. ruff + pyright clean.

## What's in the box

| Surface | Behavior |
|---|---|
| `cmd_detect_injection` helper | 9-class regex catalogue + score-based severity (none/low/medium/high). Position-aware: `>`-quoted lines halved; buried-past-500-chars × 0.75 |
| `stage_09_detect_injection` | Flipped to `enabled: True`; args contract pinned by foundation stub (`--cr-dir`, `--intent-context`) and guarded by `test_stage_09_detect_injection_enabled_with_pinned_args` |
| Quarantine (severity ≥ Medium) | Rewrites `intent_context.json` in place with `quarantine: true` + redacted fields using **real** field names (`title`/`body`/`commits`), not the v1-draft's `description`/`commits: []` |
| `cmd_classify_intent` guard | Short-circuits to `{intent: mixed, source: quarantine}` when upstream set `quarantine: true` — redacted body never reaches the classifier |
| Canonical finding (severity ≥ High) | Written to `agent_injection-detector.json` so `cmd_collect_findings` picks it up via the standard glob (no new merge wiring). Flows through `normalize_legacy_finding` → `validate_finding` → `verified[]` → canonical verdict precedence → `CHANGES_REQUESTED` |
| Audit log | `.closedloop-ai/injection-log.jsonl` — pattern class names only (**never** payload content; would re-amplify), 90-day TTL sweep on read |
| Reviewer prompt | `shared_prompt.txt` gains `<untrusted_content_policy>` block before `<constraints>` |
| Premise dispatch | Quarantine preamble prepended when `intent_context.json.quarantine == true` |
| Golden fixture | `golden_injection_quarantine` end-to-end (was deferred since Phase 8) |

## Foundation reconciliation

Before implementing, I re-verified PLN-720 against the actually-shipped foundation on `main` (which landed more than the v1 draft anticipated):

- Foundation **pre-stubbed** `stage_09_detect_injection` with `enabled: False` and a pinned args contract. Implementation became "flip the stage + match the contract", not "design and add a stage".
- Args are `--cr-dir` + `--intent-context` only (no `--diff-data` that the v1 draft proposed).
- `intent_context.json` fields are `title` / `body` / `commits` (string blob, not array) — the v1 draft's `description` / `commits: []` was wrong.
- Foundation already shipped `category: "InjectionAttempt"`, `source: "injection-detector"`, `finding_scope: "pr_metadata"`, `system_marker: "pr_description"`/`commit:{sha}`, and `is_valid_finding_id("injection-detector_f0")` validation. Zero schema changes needed.

[PLN-720 v2](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-720) captures this reconciliation in a "Foundation Reconciliation (verified 2026-05-29)" section before implementation started.

## Pattern catalogue calibration

Each pattern weighted per PLN-720 §Detection pattern catalogue. Severity thresholds:

| Score | Tier | Action |
|---|---|---|
| 0 | none | proceed normally |
| 1–29 | low | proceed, audit only |
| 30–69 | medium | quarantine Premise; rewrite `intent_context.json` |
| 70+ | high | quarantine + BLOCKING `InjectionAttempt` finding |

Single-pattern thresholds:
- `instruction_override`, `system_prompt_forgery`: 50 (single match → medium)
- `role_reversal`, `unicode_tag_chars`: 40 (single match → medium)
- `output_coercion`: 35 (single match → medium)
- `directive_injection`, `tool_coercion`: 30 (single match → medium)
- `encoded_payload`, `html_comment_exfil`: 25 (single match → low, paired with anything else → medium)

Two strong patterns (instruction_override + role_reversal = 90) → high. The fixture uses this exact pairing.

## Calibration is deferred

Per PLN-720 §Rollout, weight calibration + per-repo allowlist mechanism + 20–30 PR false-positive corpus all happen in Phase 2.5 (post-shadow) — not in this PR. The pattern-class unit tests assert *presence of match*, not specific score ranges, since calibration will change those.

## Test plan

- [x] Each of the 9 pattern classes triggers on a minimal positive payload (parametrized; one test per class)
- [x] Unicode tag chars (U+E0000..U+E007F) match `unicode_tag_chars`
- [x] Zero-width / BOM runs (U+200B..U+200F, U+FEFF, 3+ consecutive) match `encoded_payload`
- [x] Clean PR descriptions (legit fix/feat copy) score 0
- [x] Severity tier mapping (medium triggers quarantine without finding; high triggers both)
- [x] `>` quote-prefix downweights matches by 0.5×
- [x] Score accumulates across title + body + commits sections
- [x] Quarantine rewrite uses real field names (`title`, `body`, `commits`); preserves clean fields
- [x] Below-medium severity does NOT add a quarantine flag
- [x] End-to-end canonical-finding round-trip through `normalize_legacy_finding` → `validate_finding` clean
- [x] Literal `<system>`, `[INST]`, `<|im_start|>` stripped + reported in `redacted_excerpts`
- [x] Audit log appends one entry per run
- [x] Audit log sweeps entries older than 90 days on read
- [x] Missing intent_context.json → empty report + return 0 (on_failure: continue contract)
- [x] Malformed intent_context.json → empty report + return 0
- [x] `cmd_classify_intent` short-circuits on `quarantine: true` to `{intent: mixed, source: quarantine}`
- [x] `cmd_classify_intent` runs the normal classifier path when `quarantine` is absent (no regression)
- [x] `stage_09_detect_injection`: `enabled=True`, args contract pinned, `depends_on: [stage_08_fetch_intent]`, `stdout: <cr>/injection_report.json`, `on_failure: continue`
- [x] Golden fixture `golden_injection_quarantine` flows from pre-baked inputs through collect → validate → finalize-result with `verdict: CHANGES_REQUESTED`, `verified_count: 1`
- [x] `test_plan_dependent_stages_disabled` updated (stage_09 no longer asserted disabled; the new contract test takes over)


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3354,6 +3354,440 @@ def _classify_intent(
...
+        0,
+    ),
+    (
+        "unicode_tag_chars", 40,
+        r"[\U000E0000-\U000E007F]",        # Unicode tag character range
+        0,
+    ),
+    (
+        "html_comment_exfil", 25,
+        r"<!--[^>]{50,}-->",               # long HTML comment
```
thadeusb, line 3444: html_comment_exfil is `<!--[^>]{50,}-->` at weight 25, and _score_text_for_injection counts every match via finditer, so N comments is N times 25. The default GitHub PR template ships several instructional `<!-- ... -->` blocks and people routinely leave them in the body. Three of those past 50 chars is 75, which is High, which writes the BLOCKING finding and routes the whole PR to CHANGES_REQUESTED. That's a benign PR blocked on template boilerplate. Cap this class's contribution, require something actually adversarial inside the comment, or don't let it alone push past medium.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3854,7 +4298,7 @@ def _build_run_plan_stages(
             "expected_outputs": [f"{cr_dir}/injection_report.json"],
             "depends_on": ["stage_08_fetch_intent"],
             "on_failure": "continue",
-            "enabled": False,  # plan 01
+            "enabled": True,  # PLN-720
```
thadeusb, line 4354: This flips stage_09 to enabled and the high path emits a real BLOCKING finding, but the test docstring says weight calibration is deferred to the shadow rollout. So you're blocking PRs in prod on weights you haven't calibrated. Is there a shadow/observe toggle somewhere, or does merging this go straight to live-blocking?

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3354,6 +3354,440 @@ def _classify_intent(
...
+    # the trigger is a commit message we fall back to pr_description (which
+    # is a fixed canonical marker, not a templated one needing a suffix).
+    system_marker = "pr_description"
+
+    now_iso = datetime.now(timezone.utc).isoformat()
+
+    # Quarantine: rewrite intent_context.json on severity ≥ Medium so
+    # downstream readers (cmd_classify_intent, Premise prompt assembly) see
+    # the quarantine flag and redacted content. Preserve the original
+    # title/commits if they were clean — only redact what triggered.
```
thadeusb: The comment right above says 'only redact what triggered,' but body is redacted unconditionally while title and commits are gated on their score. A PR where only the title trips a pattern gets its clean body wiped, and Premise loses the actual intent. At medium severity that happens with no finding emitted, so it's silent. Intended?

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3354,6 +3354,440 @@ def _classify_intent(
...
+_INJECTION_PATTERN_DEFS: list[tuple[str, int, str, int]] = [
+    (
+        "instruction_override", 50,
+        r"\b(?:ignore|disregard|forget)\s+(?:all\s+)?(?:prior|previous|above|earlier)\s+"
+        r"(?:instructions|directives|prompts|messages|context|rules)\b",
+        re.IGNORECASE,
+    ),
+    (
+        "role_reversal", 40,
+        r"\b(?:you\s+are\s+now|act\s+as|pretend\s+to\s+be|roleplay\s+as|"
```
thadeusb: role_reversal's `act as \S` matches normal phrasing like 'act as a thin wrapper' or 'act as the source of truth'. 40 points toward quarantine on wording that has nothing to do with injection.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3354,6 +3354,440 @@ def _classify_intent(
...
+                    ts = datetime.fromisoformat(ts_str.replace("Z", "+00:00"))
+                    if ts >= cutoff:
+                        kept.append(stripped)
+                except (ValueError, KeyError, TypeError):
+                    continue
+        except OSError:
+            pass
+
+    kept.append(json.dumps(entry, separators=(",", ":")))
+    log_path.write_text("\n".join(kept) + "\n")
```
thadeusb, line 3673: This reads the whole log and rewrites it every call, so it's not append-only and two reviews in the same workdir clobber each other's entries. Fine if that never happens, but the docstring calling it append-only is a stretch.

thadeusb (commented):
Detection logic reads clean, but you've flipped the stage to enabled+BLOCKING on weights the tests admit aren't calibrated, and html_comment_exfil stacks on standard GitHub PR-template boilerplate. That combo can hand a benign PR a CHANGES_REQUESTED on merge day, so I'd want shadow-first before this lands.

wongk:
All 5 P2 findings addressed in 613f3f0:

| # | Reporter | Finding | Fix |
|---|---|---|---|
| 1 | bha_p1 | `_append_injection_audit_log` crashed `AttributeError` on non-dict JSON lines | `isinstance(obj, dict)` guard + regression test `test_sweep_handles_non_dict_json_lines` (4 pathological values: list, string, number, null) |
| 2 | auditor | Comment said "only redact what triggered" but body unconditionally redacted | Comment rewritten to make the asymmetry explicit — body always redacted (highest-risk surface, may carry sub-threshold signals); title/commits preserved when clean |
| 3 | bha_p0 | `test_malformed_intent_context_returns_empty_report` only asserted `rc == 0`, never the report shape | Added `report['score']` and `report['severity']` assertions mirroring the sibling test |
| 4 | bha_p2 | README listed `intent_context.json` as a pre-baked input but it wasn't in `inputs/` | Added the file as concrete documentation of the post-quarantine shape. Harness doesn't read it; expected envelope still byte-identical |
| 5 | bhb | Same malformed test duplicated the `_run_detect_injection` helper's boilerplate (cwd-swap + stdout-capture) | Replaced ~12 lines with the helper. CLAUDE.md "delegate instead of duplicating" |

**Stats**: 471 pass (1 new), 5 skipped. ruff + pyright clean. code-review 2.7.0 → 2.7.1.

**Notable**: finding #1 is the kind of bug that quietly degrades a feature — the pipeline's `on_failure: continue` absorbed the crash so external behavior looked fine, but every subsequent run silently lost its audit-log sweep until an operator manually noticed and removed the file. The regression test ensures that quiet-failure mode can't recur.

The other four are mid-tier hygiene (comment accuracy, test thoroughness, fixture documentation, DRY) but the cumulative effect is real: the audit-log feature would have shipped broken, the malformed-test would have rotted into a no-op assertion, and the fixture README would have misled the next reader. Each is small; together they're the kind of thing that defines whether a feature stays maintainable.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 550ec70 (v2.7.3). Added `_INJECTION_CLASS_MAX_MATCHES = {"html_comment_exfil": 1}` consulted in `_score_text_for_injection`, so the class contributes at most one matchs weight per scan. A GitHub-template body with 3+ instructional `<!-- ... -->` blocks now scores ≤ 25 (low) instead of 75 (high). Classes where repetition is genuinely more dangerous (`instruction_override`, `system_prompt_forgery`) are absent from the map and still accumulate unbounded. New regression `test_github_pr_template_does_not_quarantine` pins the default GitHub template at severity ≤ low; `test_html_comment_class_capped_at_single_match` asserts ten long HTML comments yield exactly one reported match.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Intentional — no shadow toggle, this goes live-blocking on merge. Calibration model is observation-driven: examine the audit log (`.closedloop-ai/injection-log.jsonl`) + score distributions on real PRs and tune weights / thresholds / pattern membership from there. The two FP fixes in v2.7.3 from this same review (comments #3325330078 and #3325332843) narrow the two noisiest false-positive sources — `html_comment_exfil` capped, `role_reversal` `act as <noun>` constrained — which makes the audit-log data meaningfully signal-bearing for that tuning. Operationally: false High calls → tune weights/thresholds; false negatives → extend patterns. Stage_09 has `on_failure: continue` so a detector bug never aborts review even if scoring goes sideways.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Intentional, and the comment was already fixed in 613f3f0 (v2.7.1) before this review round. Rewrote the comment to be explicit: title and commits are preserved when their per-section score is 0; body is *always* redacted on quarantine because its the highest-risk surface (longest free-form attacker-controlled text), and a clean-scoring body could still carry sub-threshold signals the catalogue missed. The asymmetry is now documented in the code so the next reader doesnt have to reconstruct the rationale.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 550ec70 (v2.7.3). Replaced `act\s+as\s+\S` with `\bact\s+as\s+(?:an?\s+|the\s+)?(?:AI|LLM|model|assistant|chatbot|agent|expert|admin|root|sysop|sudoer|developer|maintainer|reviewer|approver|owner|operator|moderator|user|hacker|attacker)\b`. The other branches of the alternation (`you are now`, `pretend to be`, `roleplay as`, `from now on you are`) were already specific enough and unchanged. Regression tests: `test_role_reversal_skips_benign_act_as_phrasing` covers "act as a thin wrapper", "act as the source of truth", and "act as the boundary between API and DB"; `test_role_reversal_still_matches_persona_injection` covers "act as an AI", "act as the admin", "act as a hacker".

plugins/code-review/tools/python/code_review_helpers.py
wongk: Docs fixed in 550ec70 (v2.7.3); implementation unchanged. Rewrote the `_append_injection_audit_log` docstring and the `_INJECTION_AUDIT_LOG` constant comment to spell out the actual semantics: read-modify-write (not atomic append), sweep-on-write (not sweep-on-read — theres no reader; operators inspect the file manually for triage), concurrent-clobber risk in the same workdir explicitly accepted because the log is observational (never read by the pipeline) and concurrent reviews against one workdir are rare. The docs lied; now they dont.

PR #110, opened 2026-05-29
fix(code-review): sharpen partitions.json access prose + add top-level-dict contract test

wongk:
## Summary

A real `/start` run had the walker model ignore the existing "do NOT run ad-hoc Python one-liners against `partitions.json`" directive at [start.md:328](plugins/code-review/commands/start.md#L328) — indexed the file as `data[0]`, got `KeyError: 0`, worked around it by cat'ing the file.

This is round 5 of the same systemic pattern that drove all the post-PLN-719 follow-ups: **prose alone doesn't beat the model's default behavior, structural constraints do**. Each prior round (argparse gaps, `on_failure: abort` vs prose, enum drift, stdout mismatch) was resolved with a test or a schema check — never with a louder prose directive. Round 5 is the same shape.

## The fix

Two-pronged. Doesn't claim to make the model stop ignoring directives — just removes the cliff at the end of the "wrong choice" path.

1. **Sharpen the prose so the next-best path is correct.** Inline the actual top-level shape of `partitions.json` (`{partitions: [...], test_file_paths: [...], force_merged_count, partition_patches?}`) right above the placeholder mapping. If a model reaches for Python anyway, it sees `data["partitions"][N]` literally adjacent to the dict layout. The wrong indexing becomes the harder choice; the right one is one line above.

2. **Add a producer-side contract test** (`test_partitions_json_is_top_level_dict_not_list` in `TestPartitionPostProcessing`). Asserts `cmd_partition`'s output is a top-level dict with `partitions` / `test_file_paths` / `force_merged_count` keys. If anyone ever restructures the producer, the prose at `start.md` becomes wrong — and this test fails BEFORE a real `/start` crash surfaces the docs gap.

## Three-tier proposal status

This is the cheap tier-1 fix the user accepted. The remaining tiers are queued for later:

- **Tier 1 (this PR)** — sharpen prose + contract test. ✓
- **Tier 2 (queued)** — `partition-summary` subcommand emitting a flat per-partition representation so Python indexing is safe by construction.
- **Tier 3 (queued)** — `cmd_partition` writes per-partition prompt fragments directly, so the walker doesn't inspect `partitions.json` at all (same principle as `cmd_fetch_intent` writing `intent_context.json`).

Tier 3 is the actually-correct long-term answer and belongs on the PRD-409 to-do list as Phase B; it's a bigger refactor than this hotfix should carry.

## Test plan

- [x] `test_partitions_json_is_top_level_dict_not_list` pins shape
- [x] Full code-review test suite: 443 pass, 6 skipped, ruff clean
- [x] Manually verified the sharpened prose at [start.md:327-345](plugins/code-review/commands/start.md#L327-L345) reads cleanly and the dict-shape example is correct (matches the actual output of `cmd_partition`)

## Version

code-review 2.6.4 → 2.6.5 (PATCH).

### DIFF
--- a/plugins/code-review/commands/start.md
+++ b/plugins/code-review/commands/start.md
@@ -324,7 +324,22 @@ The orchestrator assigns each agent a unique `AGENT_ID` (e.g., `bha_p0`, `bhb`, 
 
 **Important:** When constructing agent prompts, substitute the resolved `CR_DIR` path (e.g., `.closedloop-ai/code-review/cr-38291`) into `{CR_DIR}` — agents run in separate processes and do not have access to the orchestrator's shell variables.
 
-**Placeholder source mapping** (read from the partition entry — do NOT run ad-hoc Python one-liners against `partitions.json`):
+**Reading `partitions.json` (read the file once with `cat` or `Read`, then map keys; do NOT reach for `python -c "json.load(...)[0]"`).**
+
+The shape is a **top-level dict**, not a list:
+
+```
+{
+  "partitions": [ {id, files: [...], total_loc, is_test_only}, ... ],
+  "test_file_paths": ["test/foo.ts", ...],
+  "force_merged_count": 0,
+  "partition_patches": { "p0": "...patch text...", ... }   // optional
+}
+```
+
+So `data["partitions"]` is the list. `data[0]` is a `KeyError`. If you do reach for Python anyway, use `data["partitions"][N]["files"]` — never `data[N]`. (A regression test in `TestPartitionPostProcessing` pins this top-level shape so the prose above can't silently drift away from reality.)
+
+**Placeholder source mapping** (each key resolves from the partition entry):
 - `{filepath_N}` ← `partition["files"][N]["file"]` (key is `file`, NOT `path`)
 - `{loc_N}` ← `partition["files"][N]["loc"]`
 - `{status_N}` ← `diff_data["file_statuses"][filepath]` (added/modified/removed)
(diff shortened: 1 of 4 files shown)


plugins/code-review/commands/start.md
```diff
@@ -324,7 +324,22 @@ The orchestrator assigns each agent a unique `AGENT_ID` (e.g., `bha_p0`, `bhb`,
...
+
+The shape is a **top-level dict**, not a list:
+
+```
+{
+  "partitions": [ {id, files: [...], total_loc, is_test_only}, ... ],
+  "test_file_paths": ["test/foo.ts", ...],
+  "force_merged_count": 0,
+  "partition_patches": { "p0": "...patch text...", ... }   // optional
+}
```
thadeusb, line 337: partition_patches isn't a key in partitions.json. cmd_partition only writes partitions / test_file_paths / force_merged_count, which is exactly what your new contract test pins. partition_patches comes out of the extract-patches command into a separate manifest, and even there it's a list of filenames like ["patches_p0.txt"], not a dict mapping p0 to patch text. So the one shape you're inlining to keep the model from guessing wrong has a made-up fourth key, and a model that trusts it and does data["partition_patches"]["p0"] eats the same KeyError you're trying to kill. Drop the line.

plugins/code-review/tools/python/test_code_review_helpers.py
```diff
@@ -826,6 +826,44 @@ def _run_partition(
             _sys.stdin = old_stdin
             _sys.stdout = old_stdout
 
+    def test_partitions_json_is_top_level_dict_not_list(self) -> None:
```
thadeusb, line 829: This checks the three keys are present but not that they're the only ones, so it won't catch the partition_patches line you just added to the start.md shape. If the point of this test is to keep the inlined shape honest, assert the exact key set.

thadeusb (approved):
Right instinct, but the shape you inlined to stop the model guessing invents a partition_patches key the producer never writes, so the doc you're making accurate is itself wrong. One-line removal, see comments.

PR #113, opened 2026-05-29
feat(code-review): PLN-721 Premise Reviewer Hardening (v2.9.0)

wongk:
## Summary

PLN-721 — Premise Reviewer Hardening, Option B slice. Restructures the
Premise Reviewer around four subcategories with required reasoning
certificates, adds a MEDIUM allowance gated by a cumulative-3 verdict
rule, and ships a justification escape hatch + verifier audit so
authored defenses surface in `envelope.justified[]` instead of silently
suppressing findings.

- **Four subcategories.** `necessity` / `cohesion` / `workaround` /
  `complexity` — each with required `reasoning_certificate` fields the
  verifier re-executes independently. Certificate kind must match
  `subcategory`; malformed findings rejected.
- **New `premise_prompt.txt` per-run asset.** Same contract as
  `verifier_prompt.txt`. `cmd_prep_assets` copies it; `cmd_compute_hashes`
  optionally folds its bytes into `<PROMPT_HASH>` (back-compatible —
  omitting the flag matches v2.8.1 byte-identically).
- **MEDIUM + cumulative-3 gate (Rule 4).** Premise findings can be P2;
  ≥ 3 MEDIUM Premise findings on a PR → `NEEDS_ATTENTION`. Threshold
  tunable via `.closedloop-ai/settings/verdict-thresholds.json`.
- **Justification escape hatch.** `justified: true` + `justification`
  object. Verifier runs J1 (existence) + J2 (responsiveness) before the
  six checks; Premise findings get subcategory-specific strictness.
  New verdicts: `JUSTIFIED-VALID` (→ `justified[]`) and `JUSTIFIED-INVALID`
  (stays in `verified[]`). Sensitive-path tags outrank both.
- **Output surface.** `start.md` `## Justified Findings (PLN-721)`
  section + `github-review.md` Step 6d render the new bucket in both
  modes.

Phase 7 (telemetry) deferred to v2.9.1 per Option B; this slice ships
phases 1–6 + 8 + 9.

## Test plan

- [x] `pytest plugins/code-review/tools/python/` — 539 passed, 5 skipped
  (deferred fixtures awaiting other plans)
- [x] `pytest plugins/` (full repo) — 1225 passed, 5 skipped
- [x] `pyright` on touched files — 0 errors
- [x] `test_copies_files` extended to assert `premise_prompt.txt` is
  copied and listed in stdout
- [ ] CI green on push
- [ ] Manual smoke: run `/code-review ultra` against a small PR after
  merge to confirm the new presenter sections render correctly


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -4879,8 +4962,44 @@ def _short(text: str) -> str:
...
+            _VERDICT_PREMISE_MEDIUM_THRESHOLD_DEFAULT,
+        ),
+    )
+    premise_medium_count = 0
+    for finding in verified:
+        if str(finding.get("category", "")) != "Premise":
+            continue
+        if str(finding.get("severity", "")) != "MEDIUM":
+            continue
+        if finding.get("verifier_verdict") in (
```
thadeusb: JUSTIFIED-INVALID means the verifier looked at the author's defense and refused it, so the finding lands back in verified[] with the original concern standing. But here you skip it in the cumulative count right alongside JUSTIFIED-VALID, and those two aren't the same animal: VALID got dismissed, INVALID survived. So three MEDIUM Premise findings the author tried to wave off with comments the verifier then rejected won't trip the gate, where three plain MEDIUM findings would. I know terminal JUSTIFIED-INVALID is rare since the verifier usually falls through to a six-check verdict, but excluding it still reads backwards to me. The test pins it so I assume it's deliberate, what's the reasoning?

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5922,7 +6057,14 @@ def _bump(finding: dict[str, Any], bucket: str) -> None:
             "tentative_count": 0,
             "downgrade_count": 0,
             "justified_valid_count": len(justified),
-            "justified_invalid_count": 0,
+            # PLN-721: JUSTIFIED-INVALID lands in verified[] (the
+            # justification was audited and refuted, original finding
+            # stands). Count them here so telemetry surfaces the audit
+            # outcome without needing a separate JUSTIFIED-INVALID bucket.
+            "justified_invalid_count": sum(
```
thadeusb, line 6120: This counter includes JUSTIFIED-INVALID Premise MEDIUMs but Rule 4 excludes them, so premise_cumulative_medium_count can show 3 while the verdict stays APPROVED. However you land on the exclusion above, the stat and the gate should count the same set or the telemetry contradicts the verdict it's supposed to explain.

thadeusb (approved):
Plumbing's clean and the tests are genuinely thorough. One thing nags me: JUSTIFIED-INVALID gets skipped by the cumulative gate even though the concern survived verification, and a telemetry counter that falls out of the same decision. Both contained, see inline.

plugins/code-review/tools/python/code_review_helpers.py
wongk: You're right — that asymmetry was the actual bug, and the v2.9.1 helper baked it in.

JUSTIFIED-VALID is the verifier saying yes, the author's defense addresses the concern — finding goes to `justified[]`. JUSTIFIED-INVALID is the verifier refusing the defense — original concern stands. Excluding the second one from the gate did read backwards: it let a failed wave-off attempt prevent the cumulative signal that should have fired.

Fixed in 62060fa (v2.9.2): `_count_gateable_premise_medium` now excludes only JUSTIFIED-VALID. The docstring spells out the asymmetry, the old `test_justified_invalid_excluded_from_count` is renamed + inverted to assert NEEDS_ATTENTION on three Premise MEDIUMs where one is JUSTIFIED-INVALID, and a new `test_valid_vs_invalid_are_asymmetric` pins the per-helper delta directly. Shared-counter invariant still holds — the telemetry stat and the gate now agree on the corrected policy.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Agreed — your point on the gate is the root cause here too. Once the gate counts JUSTIFIED-INVALID (62060fa, v2.9.2), the telemetry stat counts it via the same `_count_gateable_premise_medium` helper, so the operator-facing number and the verdict it explains stay aligned by construction. The parametrized `test_telemetry_count_matches_rule_4_count` already covers the JUSTIFIED-INVALID shape and continues to pass under the new policy.

PR #114, opened 2026-05-29
feat(code-review): PLN-773 Verifier Override Flow + Premise/Verification Telemetry (v2.10.0)

wongk:
## Summary

PLN-773 — Consolidates the deferred scope from PLN-721 (Phase 7 — Premise telemetry + `--justified-only`) and PLN-722 (Phase 5 — Override Flow + Phase 8 — Verifier Telemetry). Closes the operator-facing surface gaps both prior plans left open. Without the override flow, every wrong verifier dismissal required a manual code edit + full re-run. Without the telemetry, the operator had no inverse health metric to watch the verifier or the justification hatch with.

**Override flow (PLN-722 Phase 5):**
- New `RE_ASSERTED` canonical verdict + `<CACHE_DIR>/overrides/<finding_id>.json` keyed on file-content hash (matching verifier's EXISTENCE-check window) so content drift auto-invalidates
- `re-assert` subcommand: promote findings from `rejected[]` / `pending_verification[]` to `verified[]` with audit metadata
- `review-dismissed-prepare` + `review-dismissed-consolidate` subcommands: two-phase haiku second-opinion fleet; auto-promote non-`REJECTED` verdicts via `REVIEW_DISMISSED` overrides; side-by-side diff at `<CR_DIR>/review_dismissed_diff.json`
- `--no-verify` emergency bypass with required `--no-verify-reason` so the bypass is never silent
- `cmd_verify_prepare` precedence: `overrides/` checked **before** `verifications/`

**Premise telemetry (PLN-721 Phase 7):**
- `stats.justification` — rate, rejection_rate, threshold_alert (NaN-safe)
- `stats.by_subcategory` — canonical-only buckets (necessity/cohesion/workaround/complexity)
- `stats.verification.by_reviewer` — per-reviewer FP rate + `RE_ASSERTED` counter (the inverse health metric)
- `justification_rate_alert` config knob in `verdict-thresholds.json` (default 0.30 per PLN-721 §Telemetry)
- Pending-learnings jsonl writers (`premise-justifications.jsonl`, `verifier-overrides.jsonl`) with `fcntl.flock` for concurrent-write safety
- JUSTIFIED-INVALID verdicts → `pending-learnings/premise-justifications.jsonl` so `self-learning:process-learnings` can tune the verifier's J2 (responsiveness) threshold

**Output surfaces:**
- Local-mode Verifier Stats footer in `commands/start.md`
- GitHub Step 6e writing `.closedloop-ai/code-review-verifier-stats.md` (collapsible `<details>` block; audit banner outside the details on `--no-verify` runs)
- README Configuration + Override Flow sections

**Test discipline:**
- Autouse `_isolate_pending_learnings` fixture redirects pending-learnings to `tmp_path` so tests cannot pollute the real `.closedloop-ai/pending-learnings/`
- 40+ new tests covering schema, telemetry math (NaN edges, threshold toggle), override cache (hash drift / valid / fallthrough), `--no-verify` (reason required), `re-assert` (promote / no-op / not-found / batch), `--review-dismissed` (promote / no_change / missing_output), pending-learnings (single + multi + concurrent safety with 10 `threading.Thread` writers + end-to-end JUSTIFIED-INVALID wiring)

## Test plan

- [x] `pytest plugins/` — 1301 passed, 5 skipped (deferred fixtures unchanged; PLN-771/772 not yet shipped)
- [x] Concurrent-write fcntl.flock test (10 threading.Thread writers each produce exactly one well-formed JSON line)
- [x] Override hash-drift invalidation pinned with separate `valid` and `fallthrough` tests
- [x] `--no-verify-reason` required gate exits with rc=2
- [x] Mutual-exclusion documented in `start.md` (orchestrator-level enforcement)
- [ ] CI green on push
- [ ] Manual smoke: run `/code-review` + override flow against a real PR after merge to confirm presenter rendering and override persistence


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1820,8 +2081,20 @@ def cmd_verify_prepare(args: argparse.Namespace) -> int:
         "skipped_no_verification": skipped,
         "deferred_budget": deferred,
         "cache_hits": cache_hits,
+        # PLN-773 Phase 3 — operator override telemetry. ``override_hits``
+        # are findings whose verification was short-circuited by a stored
+        # override; ``override_invalidated`` are findings whose override
+        # was rejected on file-content drift and ran normal verification.
+        "override_hits": override_hits,
```
thadeusb, line 2148: The synthesized RE_ASSERTED output never reaches the verdict. verify-prepare writes agent_verifier_<fid>.json and records the fid in override_hits, but cmd_verify_consolidate only reads agent_verifier_<fid>.json when the fid is in to_verify_ids or cache_hit_ids. It never looks at override_hits, so an override fid matches none of those sets and falls through the else branch as a tier-skip.

I ran prepare then consolidate against a valid override: the finding lands in verified[] with verifier_verdict=None, not RE_ASSERTED. So the finding does get promoted (fresh findings default to verified[]), but the RE_ASSERTED label, the operator-override reasoning, and every per-reviewer re_asserted counter are gone. stats.verification.by_reviewer[*].re_asserted counts verifier_verdict == 'RE_ASSERTED' in verified[], which now never happens, so it's always 0 even while the footer's override_hits line reports N honored. Those two numbers will openly disagree in the same report.

--review-dismissed has the same hole since REVIEW_DISMISSED overrides ride the exact same prepare->consolidate path. consolidate needs to pull override_hits from the manifest and merge the stub the same way it does cache_hits. The '_synthesize... so no branch in the consolidator needs to special-case the override path' comment is wrong, it does need one.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1769,11 +2010,31 @@ def cmd_verify_prepare(args: argparse.Namespace) -> int:
...
         finding = finding_by_id[fid]
         output_path = cr_dir / f"agent_verifier_{fid}.json"
 
+        # PLN-773 Phase 3 — operator override has precedence over the
+        # verifications/ cache. Check overrides FIRST; if one exists and
+        # the file-content hash still matches, synthesize a RE_ASSERTED
+        # verifier output and skip both the cache check and the agent
+        # spawn entirely. Hash drift invalidates the override silently
+        # (logged on the manifest); the verifier then runs normally.
+        override = _load_override(cache_dir, fid)
```
thadeusb, line 2086: Overrides are keyed on finding_id, and finding_id is reviewer_f<index> where index is just the position in the reviewer's emitted list (make_finding_id, no sort before the index is assigned). Across runs that's only stable if the LLM reviewer re-emits the same findings in the same order. Re-run, the reviewer reorders or finds one fewer, and bha_f3 is now a different finding or doesn't exist at all. The content-hash check stops you from promoting an unrelated finding at a different line, but the common case is the override silently matches nothing and the operator's re-assert quietly does nothing. At minimum document that re-assert is best-effort against id drift; ideally anchor the override on something more stable than emission order.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1548,6 +1549,223 @@ def _verifications_cache_path(cache_dir: Path, key: str) -> Path:
...
+    override: dict[str, Any], finding: dict[str, Any], cr_dir: Path,
+) -> bool:
+    """Override survives only while the cited file content matches.
+
+    Recomputes ``_file_content_hash`` against the file at HEAD and compares
+    against the hash stored at override-write time. Any drift → override
+    auto-invalidated (caller logs the event and falls through to the
+    standard verifier flow).
+    """
+    stored = str(override.get("file_content_hash", ""))
```
thadeusb, line 1742: overrides get a 90-day TTL in CACHE_TTL_DAYS but nothing enforces it. _load_override and _override_is_valid only check the content hash, never asserted_at against the TTL. Either sweep-on-read here like the other namespaces or drop the constant, right now it reads as enforced and isn't.

thadeusb (commented):
The override flow promotes the finding but loses the RE_ASSERTED verdict going through consolidate, so the per-reviewer re_asserted telemetry, which is the whole point of this PR, is dead on arrival and the footer ends up contradicting itself. The unit tests stay green because none of them run prepare and consolidate together. Wire override_hits through consolidate and add a real end-to-end test, then I'll take another pass.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 36df666. `cmd_verify_consolidate` now extracts `override_hit_ids` from the manifest (`code_review_helpers.py:2419-2422`) and folds them into the same read-back branch as `cache_hit_ids` (`code_review_helpers.py:2447`). No special-case branch — your point that the synthesized stub still needs the consolidator to actually read it was correct; the original comment claiming otherwise was wrong and has been removed. Cache write-back also skips `override_hit_ids` so synthesized stubs cannot corrupt the verifications/ cache.

The hole you flagged for `--review-dismissed` is fixed by the same change since `REVIEW_DISMISSED` overrides ride the same channel.

Added `TestPR114ReviewFixes::test_prepare_then_consolidate_routes_override_to_verified` (E2E prepare+consolidate, asserts `verifier_verdict == 'RE_ASSERTED'` in `verified[]`) and `test_prepare_then_consolidate_writes_re_asserted_to_stats` (asserts the per-reviewer counter actually increments through both phases). Bumped to v2.10.1.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Documented in README `Override Flow (PLN-773)` section (36df666). Calling out the best-effort caveat explicitly: IDs are `<reviewer>_f<index>` where `<index>` is emission position, across re-runs the LLM may reorder/drop findings so an override may map to a different finding or no finding at all. Content-hash check prevents promoting an unrelated finding; the common drift case becomes a silent no-op. README points operators at `override_hits` / `override_invalidated` in the verify-prepare manifest to confirm the override actually landed.

Not anchoring on a content-stable id in this PR — that's a wider schema change touching every reviewer's emission shape. Captured the structural fix as a follow-up; for now the docs are honest about the limitation.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 36df666. New `_override_is_expired` helper (`code_review_helpers.py`) sweeps on read against `cache_ttl_days(CACHE_NAMESPACE_OVERRIDES)` (90 days). Both `_override_is_valid` branches (content-hash and the new SYSTEM_SCOPE sentinel path) run through it.

Defensive: missing or unparseable `asserted_at` returns "not expired" so pre-fix overrides written before this enforcement do not silently drop after upgrade. Stricter contract would require rejecting at write time, not read time.

Tests: `TestPR114ReviewFixes::test_override_invalidated_when_ttl_expired` (120-day-old override → invalid) + `test_override_honored_when_ttl_within_bounds` (30-day-old → valid, negative control).

PR #115, opened 2026-05-30
refactor(code-review): extract local-mode presenter to skill (v2.11.0)

wongk:
## Summary

- Move local-mode presenter content (Gate A + stage_29_present) from `commands/start.md` into a new `code-review:present-local` skill. **start.md drops from 1014 → 754 lines (~26%).** No behavior change.
- Establishes the decomposition pattern for the rest of the start.md monolith. Operator-flag skills, fast-path skill, and agent-prompts skill follow in subsequent PRs if this pilot validates cleanly.

## Why

start.md grew from ~200 lines (pre-PLN-719) to 1014 today across three plans. Every future PRD-409 plan that touches the orchestrator was set to add another 50–100 lines to the monolith — by PLN-768 + PLN-774 we'd be past 1200, and the file becomes unreviewable. Skill decomposition gives us per-skill review, focused git diffs, and independent versioning of each presenter / flag flow.

The present-local skill is the **pilot** — biggest single chunk (~275 lines), cleanest boundary (no upstream dependencies), lowest interaction risk (presenter doesn't affect spawning or verification). If trigger reliability and behavior parity hold under real `/code-review` runs, the rest of the planned decomposition follows.

## What moved

- Gate A hygiene-only early-exit format
- Local Mode: Present Results pipeline (header, mode dispatch)
- BLOCKING / HIGH / MEDIUM section templates
- Justified Findings (PLN-721) presenter
- Dismissed Findings (PLN-722) presenter
- Verifier Stats (PLN-773) footer + audit banner for `--no-verify` runs
- Operator-flag descriptions + mutual-exclusion rules
- Override precedence rule for stage_22b (context for operators reading override-related output)
- Validation Summary + final Summary + Consolidated Finding Format

## What stays in start.md

The orchestration spine: stages, gates, walker contract, operator settings reference, reviewer fleet spawn/collect contract, verifier fleet contract, GitHub mode dispatch, footer, verdict.

## Trigger contract

The skill's `description` keys on two entry conditions:
1. **Gate A early-exit** when `flags.hygiene_only == true`
2. **Standard stage_29_present** when `MODE=local`

The orchestrator's slash-command playbook now points at the skill at the former presenter location instead of inlining 275 lines of presentation rules.

## Test plan

- [x] `pytest plugins/code-review/` — 621 passed, 5 skipped
- [ ] Manual smoke: run `/code-review` against a small local PR and confirm the local-mode presenter output is byte-identical (or only differs in inconsequential whitespace) to a v2.10.1 run. **This is the trigger-reliability validation gate** — if the skill doesn't auto-load at stage_29 or produces different output, revert.
- [ ] Manual smoke: run `/code-review` with `--justified-only` and confirm the flag-suppression rules from the skill still apply
- [ ] Manual smoke: hygiene-only Gate A path renders correctly via skill

## Risk + rollback

If the skill fails to trigger reliably in production, the rollback is a single revert commit — the presenter content is preserved 1:1 in the skill file, so reverting moves it back to start.md without content loss.


plugins/code-review/commands/start.md
```diff
@@ -709,280 +709,20 @@ Entries in `verify_manifest.json.cache_hits[]` are already on disk at `agent_ver
 
 ---
 
-## Hygiene Findings (Gate A presentation)
+## Local-Mode Presenter (Gate A + stage_29_present)
```
thadeusb: You moved the presenter out but didn't update the spots that actually point the walker at it. The walker contract (`kind: present` dispatch, ~line 183) still says dispatch to "Local Mode: Present Results", and the stage_29_present per-stage note (~line 272) says the same. Both name a heading that no longer exists in this file. Gate A step 3 (~line 206) has it too: "render using the 'Hygiene Findings' format below" points at a section you just deleted. This pointer block is nice but it's not what the orchestrator reads when it hits a present stage. If it follows the dispatch contract literally it never gets told to load present-local, and your "no behavior change" claim doesn't hold. CLAUDE.md flags exactly this under Code Quality [mistake] — audit adjacent references when you move content. Update those three to point at the skill.

plugins/code-review/skills/present-local/SKILL.md
```diff
@@ -0,0 +1,288 @@
+---
+name: present-local
+description: Render the operator-facing local-mode code review results at stage_29_present. Covers Gate A hygiene-only early-exit format, BLOCKING/HIGH/MEDIUM section templates, Justified Findings (PLN-721), Dismissed Findings (PLN-722), Verifier Stats footer (PLN-773), operator-flag descriptions, override precedence rule for stage_22b, Validation Summary, and final Summary. Invoke when MODE=local AND stage_29_present is reached, OR when Gate A early-exit (flags.hygiene_only==true) fires. Do NOT use for GitHub mode — see prompts/github-review.md.
```
thadeusb: Gate A fires in both modes, not just local. Stage 0c builds the hygiene todo list "either mode" and the hygiene section you moved here still carries the `If MODE=github, write the hygiene findings to .closedloop-ai/code-review-summary.md ...` branch. But this description says "Do NOT use for GitHub mode." So a `--github --hygiene-only` run is told to fire Gate A, which now lives in a skill it's told not to load, and github-review.md has no hygiene-only format of its own. Where does that path get its format and write step now? Either keep the github hygiene write in start.md's Gate A section or don't mode-exclude the skill for the Gate A entry point.

thadeusb (approved):
Clean extraction, content moved over intact. The one thing that has to get fixed before this is actually 'no behavior change': the walker contract and the stage_29 note still dispatch to the deleted section by name, so nothing in the spine tells the orchestrator to load the new skill. Couple notes inline.

plugins/code-review/commands/start.md
wongk: Fixed in 2b46630. All four stale references now point at the new skill or the exact renamed heading: L45 (Execution Model bullet), L183 (kind: present dispatch rule), L206 (Gate A step 3 — now references the exact heading 'Hygiene Findings Format (Gate A render target)'), L296 (stage_29_present per-stage note). You were right that the 'no behavior change' claim didn't hold under literal walker reading. It does now.

plugins/code-review/skills/present-local/SKILL.md
wongk: Already addressed in commit 9cf8f37 (pushed before this thread came in, so the version you reviewed was stale). Resolved as your option 1: kept the Gate A hygiene write in start.md and removed the entire Gate A hygiene section + MODE=github branch from the skill. The skill is now genuinely local-only — description, body, and frontmatter all consistent; grep for MODE=github in the skill returns nothing. start.md owns the mode-agnostic Gate A render path in its new 'Hygiene Findings Format (Gate A render target)' section.

PR #120, opened 2026-06-01
feat(code-review): PLN-727 Phase 1 — /fix category-dispatch (v2.13.0)

wongk:
## Summary

PLN-727 Phase 1 restructures `/code-review:fix` from a uniform verify-and-edit flow into a **category-dispatch system**. The motivating safety regression: shipping PLN-721 (Premise reviewer) introduced findings with categories `/fix` had no special handling for, so a `Premise/Cohesion` finding could in principle be applied as a code edit at the cited file:line — silently destructive. Every Premise verdict now routes to a category-specific **manual-action template** instead.

This is the first plan in PRD-409's queue that addresses a behavior already shipping today; it closes a hatch every other shipped reviewer (PLN-720/721/722/773/779) was widening.

## Dispatch buckets

| Bucket | What goes here | Handler |
|---|---|---|
| **auto-fix** | Correctness, Code Quality, Documentation, Security, safe Hygiene (`ci_artifacts` / `path_leakage` / `gitignore_drift`) | Direct code edit at anchor line, guarded by mandatory `code_snippet` drift check |
| **callsite-fix** | ImpactAnalysis | **Deferred** — routes to manual-surface today with `impact_semantic_change.md` (full `external_impact[]` list rendered). Activates when PLN-726 ships. |
| **specialized-fix** | TestQuality (`missing-coverage` / `weak-assertion` / `mock-faithfulness` / `missing-edge-case`) | **Deferred** — routes to manual-surface today. `bug-locking` and `test-deletion` permanently manual. Activates when PLN-723 ships. |
| **manual-surface** | Premise (4 subcategories), `Hygiene/sensitive_files`, `InjectionAttempt`, `CompanionChange`, Coverage gaps (re-run-or-acknowledge meta-action) | Category-specific template rendered into the final report |

## Behavioral changes

- **Source: `review_result.json` ONLY.** `validate_output.json` fallback removed per PRD-409 zero-release backward-compat policy.
- **Default in non-TTY: print plan + manual-surface report, exit without applying.** `--apply` required for code modification. Removes the prior auto-yes-after-5s timeout footgun. Interactive (TTY) prompts `Y/N` with no timeout.
- **Mandatory `code_snippet`-drift check before any code-modifying fix.** Re-greps the cited file at `line ±3` for the snippet's first line; miss → `STALE_FINDING` tag, skip, route to manual-surface with drift note. Stops `/fix` from clobbering code when the line has shifted, the bug was already fixed, or the line number is stale.
[...]

PR #121, opened 2026-06-02
feat(code-review): PLN-725 Phase 1 signal extraction foundation (v2.14.0)

wongk:
## Summary

Ships **Phase 1 of [PLN-725](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-725)** — the signal-extraction stage that fronts the deterministic coverage routing pipeline. Foundation only; no routing behavior change yet (matches the plan's Rollout Phase A: Shadow).

This is the first installment of a 10-phase rollout. Each phase will ship as an independent PR per PRD-409 cadence.

### Why this now

The PR #120 reviewer-recall gap (thadeusb's 4 round-2 findings) confirmed that the current single-reviewer architecture is brittle when specialized expertise is needed (auth, TypeScript types, hygiene-subcategory awareness, etc.). PLN-725 builds the framework that lets specialized reviewers be **dispatched deterministically** based on the contents of a diff rather than relying on one reviewer to recall every concern. A future grounding / spec-conformance reviewer will plug into this framework instead of standing alone.

## What's in this PR

| File | Type | Purpose |
|---|---|---|
| `plugins/code-review/tools/python/signal_taxonomy.json` | **NEW** | 48-signal fixed vocabulary across 4 categories (language ×14, framework ×12, concern ×18, quality ×4). Each entry carries `category`, `description`, `recommended_min_confidence`. |
| `plugins/code-review/tools/prompts/signal_extraction_prompt.txt` | **NEW** | Codifies the closed-vocabulary contract, evidence requirement (`<file>:<line> — <rationale>`), the `0.7` confidence floor, calibration ladder, and bias-toward-emission rule. |
| `plugins/code-review/tools/python/code_review_helpers.py` | Changed | Adds `extract-signals-prepare` and `extract-signals-consolidate` subcommands, plus pure helpers (`load_signal_taxonomy`, `validate_signal_extraction_output`, `fail_closed_signal_set`, `signal_extraction_cache_key`). |
| `plugins/code-review/tools/python/test_code_review_helpers.py` | Changed | 28 new tests across 6 classes. |
| `plugins/code-review/.claude-plugin/plugin.json` | Changed | Version bump to `v2.14.0` (MINOR). |
| `CHANGELOG.md` | Changed | Generated via `/update-documentation --changelog-only --plugin code-review`. |

### Subcommand shape

Modelled on PLN-722's verifier two-step pattern:

1. **`extract-signals-prepare`** — reads `diff_data.json` + optional intent, computes the `(diff_tip, taxonomy_hash, prompt_hash)` cache key.
   - **Cache hit** (`signals/` namespace; 7-day TTL per PLN-719 §9): writes the cached extraction straight to `<cr_dir>/extract_signals.json` and emits a manifest with `status: \"cache_hit\"`. No agent spawn needed.
   - **Cache miss**: writes a bounded agent-input bundle (capped at 25 files, 20 added/removed lines per file) plus a per-run taxonomy snapshot and a manifest with `status: \"needs_agent\"`, `input_path`, `output_path`, `taxonomy_path`, `prompt_path`.

2. **`extract-signals-consolidate`** — validates LLM output against the taxonomy contract (closed vocabulary, evidence required, confidence in `[0.7, 1.0]`, no duplicate names).
   - **All valid** → writes canonical `extract_signals.json` with `status: \"ok\"`; caches.
   - **All rejected / read failure** → writes fail-closed signal set (every taxonomy signal at `0.5` confidence per PLN-725 §2). Emits a MEDIUM `system_marker: \"signal-extraction-failed\"` Coverage finding so the operator footer surfaces the degraded mode. Does **not** cache.
   - **Partial validity** → keeps valid signals, records errors for observability, no fail-closed finding emitted.

### Fail-closed semantics

Per PLN-725 §2: any structural extraction failure surfaces every taxonomy signal at `0.5` confidence so best-effort routing over-triggers. Required reviewers are unaffected because the architecture forbids required rules from keying solely on LLM signals (deterministic floor — Phase 2 enforcement).

## Validation

| | |
|---|---|
| `pytest plugins/` | **1410 passed**, 5 skipped (+28 new tests) |
| CLI registration smoke test | both subcommands registered, `--help` clean |

### Test coverage breakdown (28 new tests)

- `TestSignalTaxonomy` — structural invariants of the v1 asset (6 tests)
- `TestSignalExtractionCacheKey` — tuple-key invariants (4 tests)
- `TestSignalExtractionValidator` — every rejection path enumerated (10 tests: invented name, empty evidence, missing evidence, sub-floor confidence, super-1.0 confidence, non-numeric confidence, duplicate names, non-object output, missing `signals` list, partial-rejection plus sorting)
- `TestFailClosedSignalSet` — fail-closed composition (2 tests)
- `TestExtractSignalsPrepare` — cache hit + cache miss CLI paths (2 tests)
- `TestExtractSignalsConsolidate` — ok / fail-closed / unreadable / partial-validity CLI paths (4 tests)

Phase 1 ships **only** the signal-extraction stage. The following are explicit follow-on PRs:

| Phase | Scope |
|---|---|
| 2 | `critic-gates.json` `coverage[]` schema extension + `resolve-coverage` deterministic subcommand + `migrate-critic-gates` legacy soft-compat |
| 3 | `coverage-critic` LLM stage (additive-only) + constraint enforcement |
| 4 | `start.md` Tasks 5.6–5.9 integration; switch Task 8 to read `coverage_plan.json` |
| 6 | `verify-coverage` + verdict precedence (BLOCKING coverage gaps) |
| 7 | `arbitrate-budget` integration for coverage_plan |
| 9 | `bootstrap` `agent-decomposer` mirror of the taxonomy |
| 10 | Version + docs |

Per PRD-409 ramp: until Phase 4 wires the orchestrator, this PR adds two new subcommands to the helper CLI but **does not change the reviewer set** any existing run dispatches. Existing callers are unaffected.

## Risk

- **New cache namespace**: `signals/` was already declared in `code_review_schema.py` (PLN-719 §9); we are now populating it.
- **Asset size**: taxonomy is 48 entries, ~5 KB; prompt is ~3.5 KB. Both are well under any context budget.
- **Determinism**: signal-extraction is LLM-driven by design; the validator + cache key + fail-closed defaults are the determinism guarantees.

## Test plan

- [x] `ruff check` clean
- [x] `pyright` 0 errors
- [x] Cache hit / miss round-trip exercised in tests
- [x] Validator rejection paths individually pinned (10 enumerated)
- [x] Fail-closed composition pinned
- [x] Operator-visible finding emission pinned

ClosedLoop manual loop: [019e889d-a12f-749f-800c-f0f9034996fc](https://app.closedloop.ai/closedloop-ai/loops/019e889d-a12f-749f-800c-f0f9034996fc)


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5301,6 +5302,640 @@ def cmd_detect_injection(args: argparse.Namespace) -> int:
...
+            "added_overflow": added_overflow,
+            "removed_sample": removed_sample,
+            "removed_overflow": removed_overflow,
+        })
+
+    return {
+        "files": files,
+        "sample_diff_excerpts": excerpts,
+        "intent": intent_summary or {},
+        "change_classes": diff_data.get("change_classes", []) or [],
```
thadeusb: Nothing in the repo writes change_classes into diff_data.json. parse-diff emits files_to_review/file_statuses/file_loc/total_loc/changed_ranges/patch_lines and that's it, so this is always []. The prompt (signal_extraction_prompt.txt:11) tells the agent it'll receive 'deterministic file classifications already detected' like schema_change/infrastructure_change, and it never will. Either wire change_classes into parse-diff or drop it from the input bundle and the prompt. As is, the agent is told a populated field exists, gets nothing, and has to re-derive those signals it was promised were already done for it.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5301,6 +5302,640 @@ def cmd_detect_injection(args: argparse.Namespace) -> int:
...
+
+def signal_extraction_cache_key(
+    diff_tip: str, taxonomy_hash: str, prompt_hash: str,
+) -> str:
+    """Cache key for the ``signals`` namespace (PLN-725).
+
+    Mirrors the verifier's namespace contract: the tuple
+    ``(diff_tip, taxonomy_hash, prompt_hash)`` is the complete set of
+    inputs the extraction is a pure function of. Any change to the diff
+    head, the taxonomy file, or the prompt asset flips the key.
```
thadeusb: You hash the taxonomy from its actual bytes (load_signal_taxonomy -> _taxonomy_hash) so any taxonomy edit busts the key for real. prompt_hash is taken on faith from --prompt-hash, which defaults to "" in prepare. So 'the prompt asset flips the key' only holds if the orchestrator remembers to pass a correct hash, and prepare doesn't even read the prompt file (it only records prompt_path in the manifest). prepare already has prompt_path in hand, so hash the file here the same way you hash the taxonomy, or make --prompt-hash required. Otherwise a Phase 4 wiring that forgets it serves stale extractions across prompt edits with no signal that anything's wrong.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5301,6 +5302,640 @@ def cmd_detect_injection(args: argparse.Namespace) -> int:
...
+    for path, status in sorted(file_statuses.items()):
+        pl = patch_lines.get(path, {}) or {}
+        added = pl.get("added_lines", {}) or {}
+        removed = pl.get("removed_lines", {}) or {}
+        files.append({
+            "path": path,
+            "status": status,
+            "lines_added": len(added),
+            "lines_removed": len(removed),
+            "loc": file_loc.get(path, 0),
```
thadeusb: file_loc[path] is {added, removed}, not an int, so loc ends up a dict here and the dict[str, int] annotation on 5502 is wrong. lines_added/lines_removed already carry this, so the field is redundant too.

thadeusb (approved):
Solid foundation and the tests actually exercise the rejection paths and cache-key invariants, not just the happy path. The one thing I'd fix before Phase 4 leans on it: the prompt advertises a change_classes field that nothing in the repo ever populates, so the agent always gets it empty. Couple smaller notes on the cache key and a type slip.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 11d1a75 (v2.14.1). `change_classes` was advertised in the prompt and defaulted into the input bundle, but `parse-diff` never populates it — agent would have always received `[]`. Removed from both `_build_signal_input` and `signal_extraction_prompt.txt:11`. Phase 2's `resolve-coverage` can re-add it as part of the deterministic file-classification stage. Regression tests `test_input_bundle_omits_change_classes` + `test_prompt_does_not_advertise_change_classes` pin the contract.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 11d1a75 (v2.14.1). The prompt is now content-addressed via a new `_signal_extraction_prompt_hash` helper (mirrors `_taxonomy_hash`), read inside `cmd_extract_signals_prepare`. The `--prompt-hash` flag stays as an orchestrator override but is no longer required for correctness — default behavior reads the prompt file's bytes and busts the cache key on any edit. Cache-key docstring updated to be honest about the source. Regression test `test_prepare_uses_file_bytes_for_cache_key` exercises two runs with different prompt bytes and pins that the manifest cache_keys differ.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 11d1a75 (v2.14.1). The annotation was wrong and the field was redundant — `lines_added`/`lines_removed` already carry the per-file churn. Removed both the annotation and the `loc` field. The test fixture also carried the same bug (bare ints instead of `{added, removed}` dicts) which the PR's tests didn't catch because they never asserted on the field — both corrected. Regression test `test_input_bundle_files_omit_loc_field` pins the removal.

PR #124, opened 2026-06-02
feat(code-review): PLN-725 Phase 2 coverage resolution (v2.15.0)

wongk:
## Summary

Ships **Phase 2 of [PLN-725](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-725)** — the `coverage[]` rule schema, the `resolve-coverage` deterministic subcommand, and the `migrate-critic-gates` legacy soft-compat rewriter. Foundation only; no orchestrator wiring yet (Phase 4 will replace the `route` subcommand's domain-critic selection; Phase 6 will gate the verdict on missing required reviewers). Both new subcommands are additive — existing callers are unaffected.

Second of ten planned rollout steps. Stacked on PR #121 (Phase 1, shipped v2.14.0/v2.14.1).

## What's in this PR

| Layer | New |
|---|---|
| **Schema** | `COVERAGE_TRIGGER_TYPES`, `COVERAGE_DETERMINISTIC_TRIGGERS`, `COVERAGE_LLM_TRIGGERS`, `COVERAGE_SCOPES`, `COVERAGE_CHANGE_CLASSES`, `COVERAGE_CORE_REQUIRED` in `code_review_schema.py` |
| **Classifier** | `classify_file_changes` + `CHANGE_CLASS_PATH_PATTERNS` (4 canonical classes: schema, infra, build-config, dependency) |
| **Triggers** | `_trigger_fires` evaluates all 6 types against diff state |
| **Resolver** | Pure `resolve_coverage()` + `cmd_resolve_coverage` CLI |
| **Migrator** | Pure `migrate_legacy_module_critics()` + `cmd_migrate_critic_gates` CLI |
| **Validation** | `_validate_coverage_rule` lints rules (unknown trigger types, unknown scopes, missing fields) |
| **Tests** | 51 across 7 classes |

### The architectural invariant (PLN-725 §1)

Triggers split into **deterministic** (`always`/`extension`/`path_pattern`/`content_signal`/`change_class`) and **LLM** (`signal`). A rule with `required: true` whose triggers are **only** LLM-signal is **auto-downgraded to best-effort** with an audit warning. LLM signals can ADD reviewers but never solely DRIVE required selection.

This is the deterministic floor for auto-merge — Phase 1's fail-closed signal extraction already guarantees best-effort over-triggers on extraction failure, and this PR's downgrade rule guarantees required selection never depends on LLM output. Both are necessary for PRD-409's 2-round auto-merge.

### Six trigger types

| Type | Determinism | Inputs |
|---|---|---|
| `always` | det | none |
| `extension` | det | `extensions[]`, optional `min_files` (default 1) |
| `path_pattern` | det | `patterns[]` (glob via existing `_glob_to_regex`) |
| `content_signal` | det | `pattern` (regex on **added lines only**), optional `max_scan_lines` cap |
| `change_class` | det | `class` (one of the 4 canonical classes) |
| `signal` | **LLM** | `name` (taxonomy), optional `min_confidence` |

OR semantics within a rule: the first trigger that fires selects the reviewer.

### Soft-compat without a file rewrite

`resolve_coverage` reads **both** `coverage[]` and `moduleCritics[]`. Legacy entries auto-migrate at evaluation time, so a `critic-gates.json` with only legacy entries (e.g. the one this repo currently ships from PR #123) **continues to route reviewers as best-effort** without any manual upgrade. The `[DEPRECATED]` warning surfaces in the Coverage Plan so operators see the call-to-migrate.

The `migrate-critic-gates` subcommand is a one-time helper for operators who want to rewrite their file on disk; it preserves the legacy `moduleCritics[]` block for a one-release back-out window.

### Always-add core

Every Coverage Plan includes `bug_hunter_a`, `bug_hunter_b`, `unified_auditor`, `premise_reviewer`, `test_quality` (reserved slot for PLN-723) labeled `source: "core"`. This makes the floor reproducible from any diff.

### Required-wins-over-best-effort dedup

A reviewer hit by a best-effort rule first and a required rule second ends up in `required[]`. Promotion is deterministic in rule order.

## Validation

| | |
|---|---|
| `pytest plugins/` | **1473 passed**, 5 skipped (+51 new tests) |
| CLI smoke (`--help`) | both subcommands registered |
| Inline trigger smoke (6 types) | all pass |

### Test coverage breakdown (51 new tests)

- `TestClassifyFileChanges` (7) — every change_class detector
- `TestSignalsToConfidenceMap` (4) — Phase 1 output flattening
- `TestTriggerFires` (16) — every trigger type's positive/negative/boundary case + content_signal added-vs-removed lines + max_scan_lines + malformed regex
- `TestMigrateLegacyModuleCritics` (4) — soft-compat translator
- `TestResolveCoverage` (13) — core-always-added, determinism downgrade, mixed-trigger-required-stays-required, scope filter, dedup promotion, legacy soft-compat, signal-trigger-without-extract_signals
- `TestResolveCoverageCLI` (2)
- `TestMigrateCriticGatesCLI` (5) — `--dry-run`, `--in-place`, explicit `--output`, missing-output error, preserves existing canonical entries

Phase 2 ships **only** the rule schema, resolver, and migrator. The following are explicit follow-on PRs:

| Phase | Scope |
|---|---|
| 3 | `coverage-critic` LLM stage (additive-only, constraint enforcement) |
| 4 | `start.md` Tasks 5.6–5.9; switch Task 8 to read `coverage_plan.json`; retire `route` for reviewer selection |
| 6 | `verify-coverage` + BLOCKING coverage-gap verdict precedence |
| 7 | `arbitrate-budget` integration for coverage_plan |
| 9 | `bootstrap` `agent-decomposer` taxonomy mirror |

Per PRD-409 ramp: until Phase 4 wires the orchestrator, this PR adds two new helper CLI subcommands but **does not change the reviewer set** any existing run dispatches.

## Risk

- **Legacy `critic-gates.json` continues to work**: the repo's `.closedloop-ai/settings/critic-gates.json` (added in PR #123) uses only `moduleCritics[]`. When Phase 4 lands, `resolve-coverage` will read it via soft-compat — no operator action required until they want finer-grained control.
- **File-size impact**: helpers file grew from ~9180 → ~9750 lines (+5.8%); still well under the ~12K threshold we agreed to revisit.

## Test plan

- [x] `ruff check` clean
- [x] `pyright` 0 errors
- [x] Inline pure-function smoke (6 trigger types, classifier, migration, end-to-end resolver)
- [x] Determinism invariant pinned (required+signal-only → best-effort + warning)
- [x] Soft-compat path pinned (legacy-only critic-gates still routes reviewers)
- [x] Dedup promotion path pinned (required wins)

ClosedLoop loop: [019e889d-a12f-749f-800c-f0f9034996fc](https://app.closedloop.ai/closedloop-ai/loops/019e889d-a12f-749f-800c-f0f9034996fc)


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5975,6 +5979,609 @@ def _emit_signal_extraction_failed_finding(
...
+        if not isinstance(entry, dict):
+            skipped += 1
+            continue
+        patterns_raw = entry.get("patterns", [])
+        critics_raw = entry.get("critics", [])
+        if not isinstance(patterns_raw, list) or not isinstance(critics_raw, list):
+            skipped += 1
+            continue
+        patterns = [
+            f"**{p}**" for p in patterns_raw if isinstance(p, str) and p
```
thadeusb, line 6200: The migration says it preserves the legacy '<sub> in path.lower()' semantics but it drops the case-insensitivity. Old route matched pattern.lower() in fp_lower; here you wrap the substring as `**sub**` and run it through _glob_to_regex, which compiles a case-sensitive regex (^.*sub.*$, no IGNORECASE). So a moduleCritics pattern like 'graphql' that used to match src/GraphQL/schema.ts won't after migration, and whatever reviewer was scoped to that module silently drops off the coverage plan. I ran it to confirm. Compile these with re.IGNORECASE or lowercase both sides. Same wrapping also turns any legacy substring with a * or ? in it into a glob (a*b becomes a[^/]*b) instead of a literal, though that's a rarer case.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5975,6 +5979,609 @@ def _emit_signal_extraction_failed_finding(
...
+def _validate_coverage_rule(
+    rule: dict[str, Any], index: int,
+) -> tuple[bool, list[str]]:
+    """Structural validation of a coverage[] rule. Returns (ok, errors).
+
+    Catches the contract violations that would otherwise produce
+    silent-misroute behaviour: unknown trigger type, unknown scope,
+    required-with-only-llm-signals (downgraded by the resolver to
+    best-effort rather than rejected, but the warning is emitted here
+    too so callers can lint critic-gates.json at edit time).
```
thadeusb: Docstring says the required-with-only-llm-signals warning 'is emitted here too so callers can lint at edit time' but this function never emits it, only resolve_coverage does. Either wire the check in here or fix the docstring, otherwise the next person goes hunting for a warning that isn't where it says it is.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5975,6 +5979,609 @@ def _emit_signal_extraction_failed_finding(
...
+
+    legacy = current.get("moduleCritics", [])
+    legacy = legacy if isinstance(legacy, list) else []
+    migrated, warnings = migrate_legacy_module_critics(legacy)
+
+    existing_coverage = current.get("coverage", [])
+    existing_coverage = existing_coverage if isinstance(existing_coverage, list) else []
+
+    new_state = dict(current)
+    new_state["coverage"] = list(existing_coverage) + migrated
```
thadeusb: This isn't idempotent. You append migrated entries onto existing coverage[] and leave moduleCritics[] on disk, so running --in-place twice migrates the legacy block again and coverage[] ends up with two copies of every rule. resolve_coverage dedups by reviewer name so the resolved plan stays correct, but the file grows on each run. Skip entries already tagged _migrated_from, or drop moduleCritics on write.

thadeusb (approved):
The resolver and trigger eval read clean and the tests hit the boundary cases. The one thing that actually bites: the moduleCritics migration goes case-sensitive, so legacy substring rules silently stop matching paths they used to cover. Two smaller notes besides.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 436e691 (v2.15.1). `path_pattern` triggers now accept an opt-in `ignore_case` field (default False — canonical rules keep explicit case-sensitive semantics). The migration emits `ignore_case: True` on every migrated trigger, restoring the legacy `substring.lower() in path.lower()` behavior. End-to-end test `test_migrated_rule_routes_uppercase_path` exercises `moduleCritics: [{patterns: ['graphql'], critics: ['security-privacy']}]` against `src/GraphQL/schema.ts` and pins that the reviewer routes via the migrated rule. Tests `test_path_pattern_default_is_case_sensitive` + `test_path_pattern_explicit_ignore_case_false_is_case_sensitive` pin that canonical rules without the flag stay case-sensitive. The rarer `a*b` literal case isn't addressed in this PR — operator-authored patterns are expected to use canonical glob semantics; the migration only adapts substrings under the established legacy contract.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 436e691 (v2.15.1). The check belongs in the resolver (where the match-vs-no-match distinction lives — see your bha_p0_f1 finding which spotted the related premature-warning bug). The validator's role is edit-time structural problems only. Docstring updated to reflect this split: `_validate_coverage_rule` enumerates what it catches (missing/invalid reviewer, missing/empty triggers, unknown trigger types, **unknown change_class.class values (new)**, unknown scopes) and explicitly defers the required-with-only-LLM-signal invariant to `resolve_coverage`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in 436e691 (v2.15.1). `cmd_migrate_critic_gates` now prunes any prior `_migrated_from="moduleCritics"` entries from `coverage[]` before appending the freshly migrated set. Running `--in-place` N times is equivalent to running it once. Operator-edited canonical entries (no marker) are preserved untouched. Two regression tests pin the behavior: `test_second_in_place_run_does_not_duplicate` runs migration twice and asserts byte-equality of `coverage[]`; `test_operator_edited_canonical_entries_preserved` mixes a canonical `ts-expert` rule with a legacy block and confirms both survive exactly once after two runs. New summary field `prior_migrated_pruned` makes the idempotency visible to operators.

PR #125, opened 2026-06-02
feat(code-review): Batch verifier fleet to reduce agent spawns

mikeangstadt:
## Summary

- Batch the verifier fleet (stage_23) from 1-agent-per-finding to `VERIFY_BATCH_SIZE` findings per agent (default 5)
- For a typical 15-finding review: 15 agents → 3 agents (~80% reduction in verifier spawns)
- Each batch agent shares codebase context across findings, eliminating duplicate Read/Grep work
- Configurable via `.closedloop-ai/settings/code-review.json` key `verify_batch_size` (1-20, default 5)
- Set to `1` to restore one-agent-per-finding behavior

## Motivation

Session log analysis showed the verifier fleet as the largest source of sub-agent spawns in code-review runs (up to 50 agents). Each agent independently bootstraps (~2K tokens), reads the verifier prompt, and explores overlapping codebase files. Batching amortizes all three costs.

## Test plan

- [ ] Run code-review on a PR with 10+ findings, verify verifier agents spawn in batches
- [ ] Verify each batch produces one `agent_verifier_<finding_id>.json` per finding
- [ ] Verify `verify_batch_size: 1` in settings restores 1:1 behavior
- [ ] Verify batch agent failure degrades all findings in batch to `pending_verification[]`

### DIFF
--- a/plugins/code-review/commands/start.md
+++ b/plugins/code-review/commands/start.md
@@ -672,7 +672,19 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
 
 ### Spawn contract
 
-Group entries in `verify_manifest.json.to_verify[]` into batches of up to `VERIFY_BATCH_SIZE` findings each (default `5`). Batching amortizes the per-agent bootstrap cost and lets the verifier reuse codebase context across findings that touch the same files.
+> **Independence trade-off — read before tuning the batch size.** The whole
+> point of this pass is an *independent* second opinion per finding. Packing
+> N findings into one agent means those N verdicts share a single agent
+> context, so they are no longer independent: one lazy, biased, or anchored
+> pass now taints every finding in the batch instead of one. Batching trades
+> the per-finding falsification guarantee for token/spawn savings — it does
+> not preserve it. When that guarantee matters more than cost, set
+> `verify_batch_size` to `1` (see "Batch size" below) to restore one
+> independent agent per finding.
+
+Read the resolved, already-clamped batch size from `verify_manifest.json.verify_batch_size` (written by `cmd_verify_prepare` — see "Batch size" below; do NOT re-parse the settings file in the walker).
+
+**Group `verify_manifest.json.to_verify[]` by `model` first, then slice each model-group into batches of up to `verify_batch_size` findings.** Grouping by model before slicing guarantees a batch never mixes models, so step 3's single-model assumption holds even after the planned cross-model split (a different verifier model per original reviewer) lands. Today every entry is `sonnet`, so this is a single group. Batching amortizes the per-agent bootstrap cost and lets the verifier reuse codebase context across findings that touch the same files.
 
 For each batch:
 
@@ -694,15 +706,15 @@ For each batch:
    Process each finding independently. Write one output file per finding
    at the path each input file specifies. Do not write anywhere else.
    ```
-   Substitute the resolved paths from the manifest entries (the verifier prompt is at `<CR_DIR>/verifier_prompt.txt`, copied by `stage_02_prep_assets`).
-3. Set `model` to the batch entries' `model` field (currently uniform `sonnet`; all entries in a batch share the same model).
+   Substitute the resolved paths from the manifest entries (the verifier prompt is at `<CR_DIR>/verifier_prompt.txt`, copied by `stage_02_prep_assets`). That prompt is batch-aware: it processes each input file in turn, runs the full protocol against each independently, and writes one verdict file per finding.
+3. Set `model` to the batch's shared `model` field. Because the batch was sliced within a single model-group (see grouping above), all entries share one model by construction — currently uniform `sonnet`.
 
-`VERIFY_BATCH_SIZE` is configurable via `.closedloop-ai/settings/code-review.json` key `verify_batch_size` (integer, min 1, max 20, default 5). Set to `1` to restore one-agent-per-finding behavior.
+**Batch size.** `verify_batch_size` is operator-tunable via `.closedloop-ai/settings/code-review.json`. `cmd_verify_prepare` reads it through `_load_code_review_settings`, which clamps it to the integer range `[1, 20]` (default `5`); a non-integer, boolean, or out-of-range value (e.g. `0` or the string `"5"`) falls back to the default exactly the way `bha_unified_threshold_loc` does. The clamped result is surfaced in the manifest as `verify_batch_size` — the walker uses that, never the raw file. Set it to `1` to restore one-agent-per-finding (fully independent) verification.
 
 ### Collection contract
 
 - Call `TaskOutput` (block: true) for every spawned verifier batch agent before letting the walker proceed past `stage_23`.
-- A missing `agent_verifier_<finding_id>.json` is NOT a fatal error — `cmd_verify_consolidate` tags it as `pending_verification[]` so operators see what didn't get verified. When a batch agent fails, ALL findings in that batch degrade to `pending_verification[]`.
+- A missing `agent_verifier_<finding_id>.json` is NOT a fatal error — `cmd_verify_consolidate` tags that finding as `pending_verification[]` so operators see what didn't get verified. Consolidation keys on **per-finding output files, not on batch success**: if a batch agent writes verdicts for some of its findings and then crashes or times out, the verdicts already on disk are consolidated (and cached) normally, and only the findings with no output file degrade to `pending_verification[]`. Partial batch progress is preserved by design — recovering the verdicts a failed batch did finish is strictly better than discarding the whole batch. (Consequence: a verdict written just before a crash is trusted like any other. The batch-aware verifier prompt writes one finding's output completely before starting the next, so each verdict file is atomic; the failure mode is a *missing* file, never a half-written one.)
 - Do NOT retry verifier agents in the walker. If a verifier fails, the finding's downstream handling already covers the gap (pending) — and verifier retries would burn tokens on a finding already flagged for human review.
 - `stage_23.on_failure == "continue"`: a fleet-wide failure does NOT abort the pipeline; `verify-consolidate` and `finalize-result` produce a usable envelope even when zero verifier outputs land on disk.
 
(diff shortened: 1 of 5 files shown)


plugins/code-review/commands/start.md
```diff
@@ -672,28 +672,37 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
 
 ### Spawn contract
```
thadeusb, line 674: The whole point of this pass is an independent second opinion per finding. Five findings sharing one agent context aren't independent anymore, one lazy or biased pass now taints five verdicts instead of one. Fine as a token tradeoff with the batch=1 escape hatch, but be explicit that batching weakens the falsification guarantee, not just the spawn count.

plugins/code-review/commands/start.md
```diff
@@ -672,28 +672,37 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
...
-For each entry in `verify_manifest.json.to_verify[]`:
+Group entries in `verify_manifest.json.to_verify[]` into batches of up to `VERIFY_BATCH_SIZE` findings each (default `5`). Batching amortizes the per-agent bootstrap cost and lets the verifier reuse codebase context across findings that touch the same files.
+
+For each batch:
 
 1. Spawn one background `Task` with `subagent_type: "code-review:code-review-worker"`. The agent's tool allowlist (`Read`, `Write`, `Grep`, `Glob`) is identical to the Reviewer Fleet's — no permission changes needed.
 2. Prompt template:
    ```
    You are the FINDING VERIFIER. Read your prompt at:
      {VERIFIER_PROMPT_PATH}
```
thadeusb, line 695: This wrapper tells the agent to process N findings and write N files, but the prompt it reads at {VERIFIER_PROMPT_PATH} is verifier_prompt.txt, which is single-finding top to bottom: 'falsify a single code-review finding', '<input_file>{INPUT_PATH}</input_file>', 'you write only your single output JSON', 'Write a single JSON object to output_path'. You didn't touch that file, and the worker agent is told to follow the shared prompt exactly. So the agent gets two contradictory sets of instructions. Best case it ignores the singular language; worst case it verifies the first finding, writes one file, stops, and findings 2-5 degrade to pending_verification with no crash to tell you. That failure mode looks identical to success. verifier_prompt.txt has to be made batch-aware before this does what the description claims.

plugins/code-review/commands/start.md
```diff
@@ -672,28 +672,37 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
...
-   per-output JSON shape. Write your verdict JSON to the output path the
-   input file specifies. Do not write anywhere else.
+   Process each finding independently. Write one output file per finding
+   at the path each input file specifies. Do not write anywhere else.
    ```
-   Substitute the resolved paths from the manifest entry (the verifier prompt is at `<CR_DIR>/verifier_prompt.txt`, copied by `stage_02_prep_assets`).
-3. Set `model` to the entry's `model` field (currently uniform `sonnet`; future revisions may split by original-reviewer model for cross-model independence).
+   Substitute the resolved paths from the manifest entries (the verifier prompt is at `<CR_DIR>/verifier_prompt.txt`, copied by `stage_02_prep_assets`).
+3. Set `model` to the batch entries' `model` field (currently uniform `sonnet`; all entries in a batch share the same model).
+
```
thadeusb, line 711: Nothing groups the batch by model. The 'group entries into batches of up to VERIFY_BATCH_SIZE' instruction above just slices to_verify[] in order, so the moment the 'future revisions may split by original-reviewer model' case you deleted actually lands, a batch holds mixed-model entries and you set one model on the Task for all of them. No-op today since everything's sonnet, but the invariant you assert here isn't enforced anywhere. Group by model first, then slice.

plugins/code-review/commands/start.md
```diff
@@ -672,28 +672,37 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
...
+`VERIFY_BATCH_SIZE` is configurable via `.closedloop-ai/settings/code-review.json` key `verify_batch_size` (integer, min 1, max 20, default 5). Set to `1` to restore one-agent-per-finding behavior.
```
thadeusb, line 713: min 1 / max 20 is only stated in prose here, nothing reads verify_batch_size in code the way _load_code_review_settings clamps bha_unified_threshold_loc. What happens when an operator sets it to 0 or "5"?

thadeusb (commented):
Batching's the right lever, the verifier fleet is the biggest spawn source. But as written I don't think it works: the prompt each agent reads is still single-finding all the way through and you didn't update it, so the agent likely verifies one finding and silently drops the rest to pending. Want that and the model-grouping invariant settled before this lands.

plugins/code-review/commands/start.md
```diff
@@ -288,7 +288,7 @@ These notes annotate the run-plan stages with anything not obvious from the plan
...
-- **stage_23_verify_findings** (PLN-722): agent_fleet stage. Dispatch to the "Verifier Fleet" section below. Each spawned agent reads its `verifier_inputs/<finding_id>.json` (containing the finding + the `verifier_prompt_path` + the canonical `output_path`) and emits one verdict file at `<CR_DIR>/agent_verifier_<finding_id>.json`. `on_failure: continue` so a single agent crash never aborts review.
+- **stage_23_verify_findings** (PLN-722): agent_fleet stage. Dispatch to the "Verifier Fleet" section below. Findings are batched (`VERIFY_BATCH_SIZE`, default 5) so each spawned agent verifies multiple findings in one pass, amortizing bootstrap cost and reusing codebase context. Each agent reads `verifier_inputs/<finding_id>.json` for every finding in its batch and emits one verdict file per finding at `<CR_DIR>/agent_verifier_<finding_id>.json`. `on_failure: continue` so a single agent crash never aborts review.
```
shafty023, line 291: `CLAUDE.md` requires every plugin-file change to bump that plugin's `plugin.json` in the same commit. This PR changes the `code-review` command surface but does not include `plugins/code-review/.claude-plugin/plugin.json`, and the `Plugin Version Bump` check is failing. Please add the appropriate semver bump for the `code-review` plugin.

plugins/code-review/commands/start.md
```diff
@@ -672,28 +672,37 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
...
+3. Set `model` to the batch entries' `model` field (currently uniform `sonnet`; all entries in a batch share the same model).
+
+`VERIFY_BATCH_SIZE` is configurable via `.closedloop-ai/settings/code-review.json` key `verify_batch_size` (integer, min 1, max 20, default 5). Set to `1` to restore one-agent-per-finding behavior.
 
 ### Collection contract
 
-- Call `TaskOutput` (block: true) for every spawned verifier agent before letting the walker proceed past `stage_23`.
-- A missing `agent_verifier_<finding_id>.json` is NOT a fatal error — `cmd_verify_consolidate` tags it as `pending_verification[]` so operators see what didn't get verified.
+- Call `TaskOutput` (block: true) for every spawned verifier batch agent before letting the walker proceed past `stage_23`.
+- A missing `agent_verifier_<finding_id>.json` is NOT a fatal error — `cmd_verify_consolidate` tags it as `pending_verification[]` so operators see what didn't get verified. When a batch agent fails, ALL findings in that batch degrade to `pending_verification[]`.
```
shafty023: This says a failed batch makes all findings in that batch degrade to `pending_verification[]`, but the existing consolidate step consumes any `agent_verifier_<finding_id>.json` files that are present. If a batch verifier writes one or two outputs and then times out or crashes, those partial verdicts will still be merged and cached instead of all findings in the failed batch becoming pending. Either document/accept partial outputs, or have the walker delete/ignore every output path for a failed batch before `verify-consolidate` runs.

plugins/code-review/commands/start.md
mikeangstadt: Good call — made the trade-off explicit. Added an **Independence trade-off** callout to the Spawn contract in `start.md`: batching N findings into one agent shares a single context across them, so it weakens the per-finding falsification guarantee (one lazy/biased/anchored pass now taints the whole batch), not just the spawn count. The note points operators at `verify_batch_size: 1` to restore one independent agent per finding. (0b521da)

plugins/code-review/commands/start.md
mikeangstadt: Fixed — `verifier_prompt.txt` is now batch-aware. You were right that it was single-finding top to bottom and the worker is told to follow it exactly, so findings 2–N would silently degrade to pending. The prompt now opens with a **"Your inputs — one finding, or a batch"** section: process every input file you were given, one at a time, run the full protocol against each independently, and write one verdict file per finding; "done" means every input file has its own output. The singular language downstream (`<input_file>`, "single output JSON", "Write a single JSON object") was generalized to per-finding. (0b521da)

plugins/code-review/commands/start.md
mikeangstadt: Fixed. The Spawn contract now says to **group `to_verify[]` by `model` first, then slice each model-group into batches** of up to `verify_batch_size`. That makes "a batch never mixes models" a structural guarantee rather than an assertion, so step 3's single-model `model` set holds even after the future cross-model split lands. No-op today (all `sonnet` → one group), but the invariant is now enforced by the grouping order. (0b521da)

plugins/code-review/commands/start.md
mikeangstadt: Good catch — the bound is now enforced in code, not just prose. `_load_code_review_settings` reads `verify_batch_size` and clamps it to `[1, 20]` the same way it handles `bha_unified_threshold_loc`: a non-int, bool, `0`, the string `"5"`, or any out-of-range value falls back to the default (5). `cmd_verify_prepare` resolves it once (it's the only manifest writer) and surfaces the clamped result as `verify_manifest.json.verify_batch_size`; the walker batches on that, never the raw file. Added `VERIFY_BATCH_SIZE_{DEFAULT,MIN,MAX}` constants plus unit tests for in-range/boundary/0/99/`"5"`/bool and manifest-propagation tests. (0b521da)

plugins/code-review/commands/start.md
mikeangstadt: Done — bumped `code-review` 2.12.1 → 2.13.0 (MINOR: new operator-tunable `verify_batch_size` knob wired through `cmd_verify_prepare` + the batching feature). That clears the failing Plugin Version Bump check, which compares the manifest at `merge-base...HEAD` (base version 2.12.1). (0b521da)

plugins/code-review/commands/start.md
mikeangstadt: You're right that the old prose ("ALL findings in that batch degrade") didn't match how consolidate actually works. Rather than have the walker delete partial outputs, I documented and embraced the per-finding behavior, since recovering the verdicts a failed batch *did* finish is strictly better than discarding them. The Collection contract now states: consolidation keys on **per-finding output files, not batch success** — verdicts already on disk are consolidated/cached normally, and only findings with no output file degrade to `pending_verification[]`. I also noted the atomicity guarantee: the batch-aware verifier prompt writes one finding's output completely before starting the next, so the failure mode is a *missing* file, never a half-written one. (0b521da)

PR #128, opened 2026-06-02
feat(code-review): PLN-725 Phase 4 orchestrator wiring (v2.17.0)

wongk:
## Summary

PLN-725 Phase 4 — wires the previously-deferred prepare/agent/consolidate chain for signal-extraction and coverage-critic end-to-end. Stages 11/11b/14/15/15b are enabled in **dry-run mode** — `coverage_plan.json` is produced on every review but no routing decisions yet consume it (`stage_16_arbitrate_budget` stays disabled — Phase 7 turns that on together with `stage_20` reading the plan instead of using the static spec list).

### Stage graph

| Stage | Before | After |
|---|---|---|
| `stage_11_extract_signals` | disabled | **enabled** (emits manifest) |
| `stage_11b_extract_signals_consolidate` | — | **NEW** (writes `extract_signals.json`) |
| `stage_14_resolve_coverage` | disabled, depends_on=stage_11, on_failure=abort | **enabled**, depends_on=stage_11b, on_failure=continue_with_coverage_gap |
| `stage_15_coverage_critic` | disabled | **enabled** (emits manifest) |
| `stage_15b_coverage_critic_consolidate` | — | **NEW** (writes `coverage_plan.json`) |
| `stage_16_arbitrate_budget` | disabled, depends_on=stage_15, `--coverage-plan=coverage_plan_initial.json` | disabled, depends_on=stage_15b, `--coverage-plan=coverage_plan.json` (rewired for Phase 7) |

Stage count: 32 → 34.

### Consolidate cache_hit / skipped handling

Both `cmd_extract_signals_consolidate` and `cmd_coverage_critic_consolidate` now read `manifest.status` and **no-op** when prepare's outcome was `cache_hit` (or `skipped` for coverage-critic's `--no-critic`). Prepare already wrote the canonical output in those cases; consolidate emits a one-line `{status, output_path, cache_key}` JSON and returns 0. The walker drives consolidate unconditionally without branching on manifest status.

### Walker contract + dispatch protocol (`start.md`)

- **New walker contract step 6**: after `stage_11` / `stage_15`, read prepare's manifest and run the dispatch protocol before advancing.
- **New "PLN-725 Single-Agent Dispatch" section**: codifies the singleton-agent spawn (`code-review:code-review-worker` subagent type, same allowlist as the verifier fleet; `prompt_path` / `input_path` / `model` from manifest; agent writes by convention to `agent_extract_signals.json` / `agent_coverage_critic.json`).
- **Per-Stage Notes** added for `stage_11`, `stage_11b`, `stage_14`, `stage_15`, `stage_15b`.

## Architectural invariants preserved

- **Required floor untouched.** Phase 2's `required[]` resolution is deterministic-only; the critic can only append to `best_effort[]` (locked by `merge_critic_additions`). LLM signal extraction cannot drive required reviewers.
- **Fail-closed everywhere.** Missing agent outputs surface as `signal-extraction-failed` / `coverage-critic-failed` findings via the existing v2.16.1 emitter machinery — never a silent drop.
- **Dispatch failure cannot be masked.** The `cache_hit` / `skipped` short-circuit fires only when those statuses are explicitly present. A `needs_agent` manifest with no agent output still routes through fail_closed (locked by `test_needs_agent_with_missing_output_still_fails_closed`).

## Validation

- pytest plugins/: **1560 passed**, 5 skipped (+18 new)
- ruff: clean
- pyright: 0 errors

## Test plan

- [ ] Run a local review (`/code-review start`) against a branch with real changes; confirm `extract_signals.json` and `coverage_plan.json` land in the CR dir.
- [ ] Re-run the same review; confirm prepare emits `cache_hit` and the sibling consolidates no-op (one-line stdout, file untouched).
- [ ] Spot-check the `agent_coverage_critic.json` / `agent_extract_signals.json` paths against the sibling stage's `--agent-output` arg.


plugins/code-review/commands/start.md
```diff
@@ -717,6 +724,88 @@ Entries in `verify_manifest.json.cache_hits[]` are already on disk at `agent_ver
...
+     {INPUT_PATH}
+
+   Read the bundle for the contract you are validating against. Write
+   your output JSON to:
+     {OUTPUT_PATH}
+
+   Do not write anywhere else. Do not read source files in the
+   repository unless the system prompt explicitly says so.
+   ```
+   Placeholder substitution map (do NOT substitute every placeholder from the manifest — `output_path` in the manifest is the consolidate target, not the agent target):
```
thadeusb, line 787: {OUTPUT_PATH} is never bound in the Field/Use table above, and this line tells the walker to pull the paths from the manifest. But the agent's write target isn't in the manifest. The manifest's output_path is extract_signals.json / coverage_plan.json, which the paragraph at line 758 spends three sentences warning is NOT where the agent writes. So the literal reading drops the agent output into the canonical file, then stage_11b goes looking for agent_extract_signals.json, doesn't find it, and fails closed. Net effect: every cache-miss review emits a MEDIUM signal-extraction-failed (or coverage-critic-failed) finding even though Phase 4 is meant to be invisible dry-run. Bind {OUTPUT_PATH} explicitly to the convention path in the table (agent_extract_signals.json / agent_coverage_critic.json) and pull it out of 'substitute the resolved paths from the manifest'.

CHANGELOG.md
```diff
@@ -4,6 +4,32 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code-review v2.17.1
```
thadeusb, line 16: v2.16.3 got swallowed into this heading. The _load_available_reviewers fix and its loader tests shipped in 59f5f7d as their own v2.16.3 section, but now they're sitting under v2.17.0 behind a second #### Added block and the v2.16.3 heading is gone entirely. The duplicate #### Added is the tell that this was hand-edited. CLAUDE.md says regenerate with /update-documentation instead of editing this by hand.

thadeusb (approved):
The prepare/consolidate no-op wiring is clean and the tests actually cover the needs_agent-with-missing-output trap, not just the happy path. Two things to tighten: the dispatch prompt points the walker at the wrong output path, and the changelog swallowed v2.16.3 into this release.

plugins/code-review/commands/start.md
wongk: Fixed in v2.17.1 (commit 909319d). Replaced the ambiguous "Substitute the resolved paths from the manifest" line with an explicit per-placeholder mapping table — `{PROMPT_PATH}` and `{INPUT_PATH}` come from the manifest, but `{OUTPUT_PATH}` is bound to the by-convention agent write target (`<CR_DIR>/agent_extract_signals.json` / `<CR_DIR>/agent_coverage_critic.json`) with **NOT** `manifest.output_path` called out explicitly. Same root cause your comment identified; same fix shape your comment recommended.

CHANGELOG.md
wongk: Fixed in v2.17.1 (commit 909319d). Restored the v2.16.3 heading as its own block — the `_load_available_reviewers` Fixed entry and its loader-tests Added entry are back under their original v2.16.3 section, and v2.17.0 reduced to just the Phase 4 content (single `#### Added` + single `#### Changed` block). You're right that the duplicate `#### Added` was the tell — hand-editing slipped through because I was patching v2.17.0 in place; the corrected flow re-ran `/update-documentation` after restoring the boundary.

plugins/code-review/commands/start.md
```diff
@@ -717,6 +724,88 @@ Entries in `verify_manifest.json.cache_hits[]` are already on disk at `agent_ver
...
+|---|---|
+| `prompt_path` | The agent's system prompt path — pass as `{PROMPT_PATH}` below; the agent reads it itself. |
+| `input_path` | The bounded input bundle (taxonomy + diff summary + AVAILABLE list / etc.) — pass as `{INPUT_PATH}`; the agent reads it itself. |
+| `model` | Model tier for the dispatch (`haiku` for signal-extraction, `sonnet` for coverage-critic). |
+
+The manifest's `output_path` field is the **canonical sibling consolidate output** (`extract_signals.json` / `coverage_plan.json`), NOT where the agent writes. The agent's write target is fixed by convention:
+
+| Stage | Agent writes to |
+|---|---|
+| `stage_11_extract_signals` | `<CR_DIR>/agent_extract_signals.json` |
```
shafty023, line 762: These singleton outputs should not live under the `agent_*.json` namespace. Later `stage_20_spawn_reviewers` still uses `<CR_DIR>/agent_*.json` as its reviewer-fleet success glob, and `collect-findings` globs every `agent_*.json` as findings input. A successful `agent_extract_signals.json` or `agent_coverage_critic.json` can therefore mask a totally failed reviewer fleet, and an extra top-level `findings[]` in these raw protocol outputs can alter the final verdict. Move these protocol outputs out of the `agent_` namespace, or make both stage_20 validation and `collect-findings` explicitly exclude them.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -8353,13 +8427,37 @@ def _build_run_plan_stages(
...
+            # AFTER the orchestrator's PLN-725 agent-dispatch step has
+            # written agent_extract_signals.json. On a cache_hit
+            # manifest, prepare wrote extract_signals.json itself, no
+            # agent runs, and stage_11b is a no-op (handled inside
+            # cmd_extract_signals_consolidate).
             "expected_outputs": [f"{cr_dir}/extract_signals_manifest.json"],
             "depends_on": ["stage_06_extract_patches", "stage_10_classify_intent"],
             "on_failure": "continue_with_coverage_gap",
-            "enabled": False,  # plan 05
+            "enabled": True,  # PLN-725 Phase 4
```
shafty023, line 8439: Now that this stage is enabled, its prepare args need to use the resolved `<DIFF_TIP>` and include `--cache-dir <CACHE_DIR>`; `stage_15_coverage_critic` needs the same treatment. The prepare commands compute/read cache hits from `diff_tip` and `cache_dir`, but the run plan still passes a literal `HEAD` and only the consolidate siblings receive `<CACHE_DIR>`. That makes the documented `cache_hit` / no-op path unreachable through the walker and writes cache entries under a constant diff-tip key. Pass the resolved diff tip and cache dir to both prepare stages, and pin those args in the stage-graph tests.

plugins/code-review/commands/start.md
```diff
@@ -187,9 +187,11 @@ If a token's source file does not exist yet (a prior stage that produces it was
    - `continue` — log a warning and proceed to the next stage.
    - `continue_with_coverage_gap` — emit a `system_marker: "agent-failure"` finding (see "Agent Failure Recovery" below) and proceed.
 
-6. **Run gates.** After completing a stage, scan `GATES` for any entry whose `after_stage` matches the stage just finished. Each gate checks `outputs` exist and are well-formed; `on_failure_action` follows the same `abort | continue | emit_coverage_gap` semantics.
+6. **PLN-725 singleton agent dispatch.** When the stage just finished is `stage_11_extract_signals` or `stage_15_coverage_critic`, read the manifest it just wrote and run the protocol in the "PLN-725 Single-Agent Dispatch" section below before proceeding to the next stage. The manifest's `status` field decides whether an agent spawn is needed — `cache_hit` / `skipped` skips, `needs_agent` spawns. The downstream sibling stage (`stage_11b` / `stage_15b`) walks normally as the next array entry; its `cmd` no-ops on `cache_hit` / `skipped` manifests so the walker doesn't need to branch.
```
shafty023, line 190: This dispatch step runs before Gate A, but Gate A is where `--hygiene-only` exits. With stage_11/11b enabled, hygiene-only reviews can still reach the singleton signal-extraction dispatch and spend an LLM call before the hygiene gate, despite `start.md` documenting hygiene-only as a zero-LLM deterministic check. Skip/disable the PLN-725 singleton stages when `FLAGS.hygiene_only` is true, or move the hygiene-only exit before any LLM-capable stage; add a prepare-run test for that flag.

plugins/code-review/commands/start.md
```diff
@@ -717,6 +724,88 @@ Entries in `verify_manifest.json.cache_hits[]` are already on disk at `agent_ver
...
+The manifest's `output_path` field is the **canonical sibling consolidate output** (`extract_signals.json` / `coverage_plan.json`), NOT where the agent writes. The agent's write target is fixed by convention:
+
+| Stage | Agent writes to |
+|---|---|
+| `stage_11_extract_signals` | `<CR_DIR>/agent_extract_signals.json` |
+| `stage_15_coverage_critic` | `<CR_DIR>/agent_coverage_critic.json` |
+
+These paths match the `--agent-output` arg the sibling stage's `args` declare, so the consolidate cmd finds the file with no walker substitution.
+
+Spawn one foreground `Task`:
```
shafty023, line 767: This dispatch contract mixes `foreground Task` with a required `TaskOutput` call below. The existing reviewer/verifier fleet sections spawn background tasks before using `TaskOutput`; a foreground task can complete synchronously without a task handle for `TaskOutput`, so a literal walker can attempt the dispatch and still fail to collect the singleton output before the sibling consolidate stage fail-closes. Choose one protocol explicitly: either spawn with `run_in_background: true` and block with `TaskOutput`, or run synchronously and remove the `TaskOutput` step.

shafty023:
`plugins/code-review/README.md` is stale after this PR changes the pipeline shape and bumps `code-review` to v2.17.1. The generated docs still describe `prepare-run` as a 30-stage pipeline and show an old plugin version, while the repo guardrails require plugin docs to be regenerated by `/update-documentation`. Please rerun `/update-documentation` without changelog-only mode and include the README updates for the new stage count/current version.

plugins/code-review/commands/start.md
wongk: Fixed on PR #129 (v2.18.1, commit 3fb324e). Renamed `agent_extract_signals.json` → `pln725_extract_signals.json` and `agent_coverage_critic.json` → `pln725_coverage_critic.json` to keep the singleton protocol outputs out of the `agent_*.json` namespace. Cleaner than exclusion lists — `stage_20_spawn_reviewers.expected_outputs` and `cmd_collect_findings`'s findings glob now physically cannot match these files. The agent_*.json fail-closed-finding emitters (`agent_signal-extraction-failed.json`, `agent_coverage-critic-failed.json`) stay where they are — they're designed to be collected. Pinned by `test_pln725_singleton_outputs_are_not_in_agent_namespace` and `test_collect_findings_glob_does_not_match_pln725_outputs`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed on PR #129 (v2.18.1, commit 3fb324e). `stage_11_extract_signals` and `stage_15_coverage_critic` prepare args now declare `--diff-tip <DIFF_TIP>` (walker-resolved from scope.json) instead of literal `"HEAD"`, and both receive `--cache-dir <CACHE_DIR>`. Pinned by four targeted tests in `TestPLN725Phase5PostMergeHardening` — each prepare's diff_tip is `<DIFF_TIP>` (not `HEAD`) and each has a `--cache-dir` arg bound to `<CACHE_DIR>`.

plugins/code-review/commands/start.md
wongk: Fixed on PR #129 (v2.18.1, commit 3fb324e). `stage_12_hygiene` moved from after `stage_11_extract_signals` to immediately after `stage_10_classify_intent`. Gate A (the `--hygiene-only` early exit) now fires before any PLN-725 LLM-capable stage — hygiene-only reviews exit at zero LLM cost as the contract documents. `stage_12.depends_on` is `stage_05_parse_diff` only so the reorder is safe; the `_12_` prefix is a stable label not an ordinal. Pinned by `test_stage_12_hygiene_runs_before_any_pln725_llm_stage` which enumerates every PLN-725 LLM stage and asserts `stage_12` precedes all of them, plus `test_stage_12_hygiene_position_after_classify_intent` for the lower-bound.

plugins/code-review/commands/start.md
wongk: Fixed on PR #129 (v2.18.1, commit 3fb324e). Dispatch protocol switched from "foreground Task + `TaskOutput` block-true" to **synchronous Task, no `TaskOutput`**. The previous wording mixed the two protocols — `TaskOutput` is for background tasks (the reviewer/verifier fleets); a synchronous Task completes before control returns to the walker, so there's no task handle to wait on. The new contract explicitly says: do **not** set `run_in_background: true`; advance the walker after the Task returns; no `TaskOutput` step. The pln725-prefix output paths in step 3 are also rebound to the new namespace from the related finding.

PR #129, opened 2026-06-02
feat(code-review): PLN-725 Phase 5 agent-definition loading (v2.18.0)

wongk:
## Summary

PLN-725 Phase 5 — produces the AVAILABLE roster the coverage-critic enforces against. The missing input that kept Phase 4 permanently in the no-roster skipped fallback. Pipeline stage_count: 34 → 35.

Smoke-tested against this repo's `.claude/agents/`: **14 reviewers loaded with 0 warnings.**

## What landed

| Layer | Component |
|---|---|
| Helper | `_parse_agent_name` — regex frontmatter parser (no PyYAML dep for one field). Tolerant by design — returns `None` for missing/unclosed frontmatter, missing name, prose-only files. |
| Helper | `_scan_agent_definitions` — walks agents dir in sorted order (stable cache keys), dedups names, surfaces per-file warnings without aborting. |
| CLI | `load-available-reviewers` — scans `--agents-dir` (default `.claude/agents`), writes a flat sorted+dedup JSON list. Empty/missing dir → empty list + exit 0 (preserves Phase 4 no-roster fall-through). |
| Stage | `stage_14a_load_available_reviewers` — helper kind, `depends_on=[]`, `on_failure=continue_with_coverage_gap`. Inserted between `stage_14` and `stage_15`. |
| Wiring | `stage_15_coverage_critic.depends_on` adds `stage_14a` — the data dependency on the roster is now explicit on the wire. |
| Docs | `start.md` Per-Stage Note for `stage_14a`. |

## Contract invariants

1. **File shape locked.** The cmd writes the same flat-list shape `_load_available_reviewers` (Phase 4 helper used by both critic prepare and consolidate) already accepts. Round-tripped in a regression test.
2. **Phase 4 fall-through preserved.** Empty `.claude/agents/` or missing dir produces empty list + exit 0 — the critic's no-roster skipped semantics still fire and the rest of the review still ships.
3. **Stable cache key.** The roster hash dimension added in v2.16.2 sorts before hashing, but the scan also sorts. Belt-and-braces — filesystem traversal order cannot flip the critic cache key.

## Validation

- pytest: **1581 passed**, 5 skipped (+19 new)
- ruff: clean
- pyright: 0 errors
- Smoke test against `.claude/agents/`: 14 reviewers, 0 warnings, output round-trips cleanly through `_load_available_reviewers`

## Progress on the 10-phase rollout

| Phase | Status |
|---|---|
| 1–4 | ✅ shipped |
| **5 — Agent-definition loading** | ✅ **shipped here** |
| 6 — `verify-coverage` + BLOCKING verdict | next |
| 7 — `arbitrate-budget` integration + spawn_reviewers reading coverage_plan.json | pending |
| 8 — Coverage Plan rendering | pending |
| 9 — Bootstrap taxonomy mirror | pending |


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6581,6 +6581,154 @@ def cmd_resolve_coverage(args: argparse.Namespace) -> int:
...
+# what ``critic-gates.json`` mentions). A filename/name mismatch
+# would surface here as a roster entry that doesn't match anything
+# the validator enforces against.
+# ---------------------------------------------------------------------------
+
+DEFAULT_AGENTS_DIR = Path(".claude/agents")
+
+_FRONTMATTER_BOUNDARY = re.compile(r"^---\s*$", re.MULTILINE)
+_FRONTMATTER_NAME = re.compile(
+    r"^name\s*:\s*([A-Za-z0-9][A-Za-z0-9_.-]*)\s*$",
```
shafty023, line 6607: This loader advertises YAML frontmatter, but the `name` parser only accepts an unquoted bare scalar. Valid frontmatter such as `name: "security-reviewer"`, `name: 'security-reviewer'`, or `name: security-reviewer # primary` is treated as missing, so that agent is dropped from `available_reviewers.json` and the run can fall into `no-roster`/`no-candidates` even though a reviewer is configured. Please parse the fenced frontmatter as YAML or extend the parser to accept YAML scalar forms, then add regression coverage for quoted/commented `name` values.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6581,6 +6581,154 @@ def cmd_resolve_coverage(args: argparse.Namespace) -> int:
...
+    visibility without aborting the load.
+    """
+    reviewers: list[str] = []
+    seen: set[str] = set()
+    warnings: list[str] = []
+    if not agents_dir.is_dir():
+        return [], [f"agents dir not found: {agents_dir}"]
+    for path in sorted(agents_dir.glob("*.md")):
+        try:
+            text = path.read_text()
```
shafty023, line 6657: This follows symlinks and reads each matching agent file to EOF. Because the review runs against a repository checkout, a PR can add `.claude/agents/x.md` as a symlink to something like `/dev/zero` or a very large file and hang/OOM the runner before the roster fallback can degrade safely. Please skip symlinks/non-regular files and read only a bounded frontmatter prefix, warning and skipping oversized files; add negative tests for symlink and oversized-agent inputs.

plugins/code-review/commands/start.md
```diff
@@ -284,11 +284,12 @@ These notes annotate the run-plan stages with anything not obvious from the plan
...
+- **stage_15_coverage_critic** (PLN-725 Phase 3, wired in Phase 4): runs `coverage-critic-prepare`. Writes `<CR_DIR>/coverage_critic_manifest.json`. On `status: "cache_hit"` or `status: "skipped"` (the `--no-critic` flag, not wired in Phase 4), prepare wrote `<CR_DIR>/coverage_plan.json` itself — no agent runs, downstream `stage_15b` no-ops. On `status: "needs_agent"`, the walker invokes the "PLN-725 Single-Agent Dispatch" protocol with the manifest's paths and target `pln725_coverage_critic.json`. `on_failure: continue` — coverage-critic failure surfaces as `critic_status: "fail_closed"` on the final plan; the deterministic floor from stage_14 still routes reviewers correctly.
```
shafty023, line 291: This parenthetical is stale for the current implementation. `coverage-critic-prepare` now also returns `status: "skipped"` for an empty loaded roster (`reason: "no-roster"`) and for a fully subscribed plan after dedupe (`reason: "no-candidates"`), not just the `--no-critic` flag. Since this file is the walker-facing dispatch contract, please list all skipped reasons so future walker changes do not accidentally dispatch the singleton critic for the no-roster/no-candidates paths.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -8512,6 +8708,41 @@ def _build_run_plan_stages(
...
+            # (it doesn't read coverage_plan_initial.json), but it slots
+            # between stage_14 and stage_15 so the data dependency is
+            # explicit: stage_15 coverage-critic-prepare reads the
+            # roster. An empty .claude/agents/ (no project agents) or
+            # missing directory produces an empty roster — stage_15
+            # then falls back to the Phase 4 no-roster skipped semantics.
+            "id": "stage_14a_load_available_reviewers",
+            "kind": "helper",
+            "subcommand": "load-available-reviewers",
+            "args": [
```
shafty023, line 8725: The real stage omits `--agents-dir`, so it relies on the default `.claude/agents` path and the runner cwd. The added CLI tests call `cmd_load_available_reviewers` with an explicit `agents_dir`, so they do not lock the actual walker path used here. Please add an existing-pattern regression test that executes this stage/subcommand without `--agents-dir` from a temp repo containing `.claude/agents/*.md`, then asserts `available_reviewers.json` is the expected flat roster and round-trips through `_load_available_reviewers`.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6581,6 +6581,154 @@ def cmd_resolve_coverage(args: argparse.Namespace) -> int:
...
+    per-file diagnostics — unreadable files, missing frontmatter,
+    duplicate names — surfaced to stderr by the caller for operator
+    visibility without aborting the load.
+    """
+    reviewers: list[str] = []
+    seen: set[str] = set()
+    warnings: list[str] = []
+    if not agents_dir.is_dir():
+        return [], [f"agents dir not found: {agents_dir}"]
+    for path in sorted(agents_dir.glob("*.md")):
```
thadeusb, line 6655: except OSError won't catch UnicodeDecodeError (it's a ValueError subclass), so one non-utf8 file in .claude/agents/ throws straight out of the loop and aborts the whole scan. The docstring right above promises unreadable files get a warning and get skipped, not that they kill the load. Catch ValueError too, or read with errors="replace".

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7154,6 +7302,19 @@ def cmd_coverage_critic_prepare(args: argparse.Namespace) -> int:
         print(load_err or "Error reading available_reviewers", file=sys.stderr)
         return 1
 
+    # PLN-725 Phase 5 fix: stage_14a_load_available_reviewers now
```
thadeusb, line 7305: The comment block just above on the exists() check still says "not-yet-shipped Phase 5 stage" and "Phase 4 surveys the pipeline without Phase 5 wired" — but Phase 5 is this PR. So now two adjacent comments contradict each other. That exists() path only fires if stage_14a fails to write the file at all, not because the loader doesn't exist. Update the old one so the next person doesn't go looking for a missing stage.

thadeusb (approved):
Empty-roster short-circuit is the right fix and the regression tests actually split no-roster from no-candidates instead of lumping them. Walked the gate reorder, the consolidate no-op, and the pln725 rename, all hold. Two small nits inline.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.19.1 (PR #130). `_FRONTMATTER_NAME` now uses named alternation (`dq` / `sq` / `bare`) accepting all three YAML scalar forms — `name: foo`, `name: "foo"`, `name: 'foo'` — and trailing `\s*(?:#.*)?$` strips inline comments from any form. `_parse_agent_name` reads `match.group("dq") or match.group("sq") or match.group("bare")` so consumers get the unquoted value regardless of how it was authored. Regression tests in `TestPLN725Phase5LoaderHardening`: `test_parse_agent_name_accepts_double_quoted`, `test_parse_agent_name_accepts_single_quoted`, `test_parse_agent_name_strips_inline_comment_from_bare`, `test_parse_agent_name_strips_inline_comment_from_quoted`, `test_parse_agent_name_handles_quoted_dashes_underscores`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.19.1 (PR #130). `_scan_agent_definitions` now does an `lstat()` check before opening anything — `stat.S_ISLNK(lst.st_mode) or not stat.S_ISREG(lst.st_mode)` skips with a warning, so a symlink (or FIFO, device, etc.) never gets dereferenced. The read itself is bounded by `_AGENT_FILE_READ_LIMIT_BYTES` (64 KiB, exported as a module constant): we open as binary, `fh.read(LIMIT + 1)`, and if the result exceeds `LIMIT` we surface an `oversized` warning and truncate to the prefix. A symlink to `/dev/zero` or a multi-GB file can no longer hang or OOM the runner before the no-roster fallback fires. Frontmatter parsing is unaffected because frontmatter is always near the top of the file — when the closing `---` boundary lies past the limit, the parser sees an unclosed block and drops the agent rather than partial-matching against a truncated value. Regression tests: `test_scan_skips_symlinks` (symlink to out-of-tree file is skipped with a `link.md … symlink` warning while a legit sibling agent loads), `test_scan_truncates_oversized_files_to_frontmatter_prefix` (head < limit, padding > limit → name parses, `oversized` warning emitted), `test_scan_skips_oversized_file_with_late_frontmatter` (head + padding > limit, tail past limit → no agent, `oversized` warning).

plugins/code-review/commands/start.md
wongk: Fixed in v2.19.1 (PR #130). The stage_15 per-stage note in `commands/start.md` now enumerates all three `"skipped"` reasons the prepare step emits: `"no-critic"` (operator flag), `"no-roster"` (loaded `available_reviewers.json` is empty — no project agents configured), and `"no-candidates"` (roster non-empty but every reviewer is already in the initial plan). The note explicitly tells the walker "MUST NOT dispatch the singleton critic, no matter which of these skip reasons fired," and re-asserts that `stage_15b` no-ops on any `"skipped"` manifest. That makes the dispatch contract self-contained for a walker-implementation reader without requiring a round-trip to `_emit_skipped_coverage_plan`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.19.1 (PR #130). Two new tests in `TestPLN725Phase5StageGraphDefaults` lock the walker path:

1. `test_stage_14a_args_omit_agents_dir_flag` asserts `"--agents-dir" not in stage["args"]` on the shipped stage, so any future refactor that adds the flag (and would let the existing CLI tests that pass an explicit `agents_dir` keep passing) trips here.
2. `test_stage_14a_default_path_round_trips_through_loader` mirrors the actual walker: seeds `.claude/agents/alpha.md` + `beta.md` in a temp repo root, uses `monkeypatch.chdir(repo_root)` so `DEFAULT_AGENTS_DIR` (a relative `Path(".claude/agents")`) resolves correctly, invokes `cmd_load_available_reviewers` with `agents_dir=None`, then round-trips the written `available_reviewers.json` through `_load_available_reviewers` (the same loader the critic uses) and asserts `["alpha-reviewer", "beta-reviewer"]`. The fixture deliberately uses one bare-scalar name and one double-quoted name so the regression doubles as integration coverage for the v2.19.1 quoted-name fix.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.19.1 (PR #130). You're right — `UnicodeDecodeError` is a `ValueError` subclass, not an `OSError` subclass, so the previous `except OSError` couldn't catch it and a single non-UTF8 file aborted the entire scan even though the docstring promised per-file degradation. The fix switches to a binary read (`open(path, "rb")`) followed by `raw.decode("utf-8", errors="replace")`. With `errors="replace"`, non-UTF8 bytes become `U+FFFD` and the decoder never raises — so the for-loop continues across the bad file. A defensive `except ValueError` guards the (theoretically unreachable) replacement-decode failure path, so the contract "no single file aborts the scan" is preserved end-to-end. Regression test: `test_scan_does_not_abort_on_non_utf8_file` seeds a real agent + a file with a raw `\xe9` byte (Latin-1 `é`) in the prose body, then asserts both agents load — the previous implementation would have raised straight out of the first iteration and left the roster with only one entry (or zero, depending on lexicographic order).

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.19.1 (PR #130). The "PLN-725 Phase 4 dry-run tolerance / not-yet-shipped Phase 5 stage" comment is replaced with prose that matches the code path as it actually exists now: Phase 5 (v2.18.0) ships `stage_14a_load_available_reviewers`, so the `available_reviewers.json` file is always produced under normal operation. The `not available_path.exists()` short-circuit is now the safety net for the case where stage_14a's write failed or was lost, not the dry-run tolerance for an absent stage. The two adjacent comments no longer contradict each other.

PR #130, opened 2026-06-02
feat(code-review): PLN-725 Phase 6 verify-coverage + BLOCKING verdict (v2.19.0)

wongk:
## Summary

- Adds `stage_15c_verify_coverage` — a deterministic post-LLM verifier that runs immediately after `stage_15b_coverage_critic_consolidate` and validates the final `coverage_plan.json` against the closed-vocabulary, additive-only, best-effort-only-critic, evidence-required, 5-cap, and no-duplicates contracts. Emits `coverage_verify.json` with `verdict: PASS | BLOCKING` plus per-check violations, and a HIGH `coverage-verify-blocking` system finding on BLOCKING.
- Phase 6 ships observational: `on_failure: continue`, exit 0 on both verdicts. Phase 7 will gate `stage_16_arbitrate_budget` / `stage_20_spawn_reviewers` by reading the verdict from `coverage_verify.json` rather than by changing the verifier's exit code.
- Removes the stale `stage_24_verify_coverage` placeholder whose shape (no roster / initial-plan args, post-verifier-stage dependency) no longer fits where the verifier actually lives; rewires `stage_16.depends_on` to `stage_15c` so the producer chain stays transitive.

## Test plan

- [x] `pytest plugins/code-review/tools/python/` — 881 passed, 5 skipped (added `TestPLN725Phase6VerifyCoveragePure`, `TestPLN725Phase6VerifyCoverageCommand`, `TestPLN725Phase6StageGraph`; updated `test_stage_16_depends_on_post_consolidate_chain_not_stage_15`, `test_plan_dependent_stages_disabled`, `test_arbitrate_budget_gated_on_phase_7`).
- [x] `pytest plugins/` — 1620 total, 0 failures.
- [x] `ruff check plugins/code-review/tools/python/` — clean.
- [x] `pyright plugins/code-review/tools/python/` — 0 errors.


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+            "message": (
+                f"initial-plan reviewers missing from final plan: {dropped} "
+                f"(coverage critic may only ADD; deletions violate the contract)"
+            ),
+        })
+
+    # closed_vocabulary — only enforced when a roster is present and
+    # non-empty, AND only against source="critic" entries. The core/rule
+    # reviewer labels in coverage_plan.json (e.g. ``bug_hunter_a``,
+    # ``unified_auditor``) are plugin-internal identifiers that the
```
thadeusb, line 7799: closed_vocabulary checks final_names, which is required[] + best_effort[] across every source. Problem is the always-add core floor (bug_hunter_a, bug_hunter_b, unified_auditor, premise_reviewer, test_quality) lives in required[] on every run and is never in .claude/agents/ — available_reviewers.json only ever holds the target repo's project agents. So any review of a repo that has a non-empty agents dir trips this and goes BLOCKING every time. I ran it: core floor + one legit critic addition + a normal roster returns closed_vocabulary on all five core names. validate_critic_additions already gets this right at line 7164 by only checking source=critic entries against the roster. Scope this check the same way, otherwise the verifier fires a HIGH coverage-verify-blocking finding on basically every real run (and blocks them outright once Phase 7 gates on it). The tests miss it because they always pass a roster containing every reviewer name including the rule/core ones.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+    if violations:
+        return violations  # downstream checks assume shape is valid
+
+    required_final, best_effort_final = _plan_reviewer_buckets(plan)
+    required_initial, best_effort_initial = _plan_reviewer_buckets(plan_initial)
+    final_names = set(_reviewer_names(required_final)) | set(_reviewer_names(best_effort_final))
+
+    # additive
+    initial_names = set(_reviewer_names(required_initial)) | set(_reviewer_names(best_effort_initial))
+    dropped = sorted(initial_names - final_names)
```
thadeusb, line 7786: additive only looks at required[] + best_effort[]. Fine now since 15c runs before arbitrate-budget. Heads-up for Phase 7 though: stage_16 moves cut reviewers into dropped_required/deferred_for_budget, both outside the two buckets this check reads. If a verify ever runs against a post-arbitration coverage_plan.json it'll flag every budgeted-out reviewer as a contract-violating deletion. Confirmed it misfires that way. Since this PR re-anchors stage_16.depends_on to 15c specifically to feed Phase 7, worth keeping in mind when you wire the gate.

thadeusb (approved):
Verifier's the right addition and the missing-input/empty-roster fallbacks are handled well. One real bug: closed_vocabulary checks every final reviewer against the roster, but the core floor (bug_hunter_a and friends) never lives in .claude/agents/, so it goes BLOCKING on every repo with a configured agents dir. Scope that check to source=critic like the consolidate validator does and it's good.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+        })
+    if violations:
+        return violations  # downstream checks assume shape is valid
+
+    required_final, best_effort_final = _plan_reviewer_buckets(plan)
+    required_initial, best_effort_initial = _plan_reviewer_buckets(plan_initial)
+    final_names = set(_reviewer_names(required_final)) | set(_reviewer_names(best_effort_final))
+
+    # additive
+    initial_names = set(_reviewer_names(required_initial)) | set(_reviewer_names(best_effort_initial))
```
shafty023, line 7785: Additivity is still bucket-insensitive here: it unions initial `required` + `best_effort` and final `required` + `best_effort`, so an initial required reviewer can move into final `best_effort[]` and the verifier returns PASS. Phase 7 will rely on this artifact to preserve the required-reviewer floor, so a corrupted cached/manual final plan could silently downgrade mandatory coverage. Please compare buckets explicitly, at least enforcing `initial.required ⊆ final.required`, and add a required-to-best-effort demotion regression.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+      - additive: every initial reviewer survives in some bucket of final
+      - closed_vocabulary: every final reviewer is in roster (when roster)
+      - critic_best_effort_only: source="critic" entries never in required[]
+      - critic_evidence: every source="critic" entry has non-empty evidence
+      - critic_cap: critic_addition count <= COVERAGE_CRITIC_MAX_ADDITIONS
+      - no_duplicates: each reviewer appears at most once across both buckets
+    """
+    violations: list[dict[str, str]] = []
+
+    # shape
```
shafty023, line 7766: The shape check only verifies that `required` and `best_effort` are lists. `_plan_reviewer_buckets` then drops non-dict entries and `_reviewer_names` ignores entries without a non-empty string `reviewer`, so malformed plans such as `{"required": [{}], "best_effort": []}` can still PASS. Since this verifier is the last guard before downstream spawning reads the plan, please reject malformed bucket entries as `check: "shape"` and cover non-object / missing-reviewer entries in pure and command-level tests.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+
+    # Missing inputs are not a verifier failure — they mean an upstream
+    # stage didn't produce its artifact (likely because a prior stage
+    # aborted). Treat as PASS with a single advisory violation so
+    # operators see why verification was vacuous, without blocking
+    # downstream observability.
+    try:
+        with open(plan_path) as f:
+            plan = json.load(f)
+    except (OSError, json.JSONDecodeError) as exc:
```
shafty023, line 8008: Missing or unreadable verifier inputs should not produce `verdict: "PASS"`. Phase 6 can still exit 0 for observational rollout, but the artifact is what Phase 7 will read before allowing spawn/arbitration. A PASS with an input violation makes "no plan was verified" indistinguishable from a real pass. Please emit `verdict: "BLOCKING"` and the HIGH `coverage-verify-blocking` finding for unreadable `coverage_plan.json` or `coverage_plan_initial.json`, while keeping exit 0.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+    Surfaces violations to the operator so the failing check can be
+    diagnosed. Fail-open on write error — telemetry is observational
+    and a write failure shouldn't crash the stage.
+    """
+    summary = [f"{v['check']}: {v['message']}" for v in violations[:10]]
+    if len(violations) > 10:
+        summary.append(f"… {len(violations) - 10} more")
+    finding = {
+        "reviewer": "coverage-verify",
+        "source": "coverage-verify",
```
shafty023, line 7897: This emitted finding uses `source: "coverage-verify"`, but `code_review_schema.SOURCES` allows `coverage-verifier`, not `coverage-verify`. When `agent_coverage-verify-blocking.json` is collected/finalized, schema validation rejects the finding exactly when the verifier needs to surface a BLOCKING result. Please use the existing canonical `coverage-verifier` source/reviewer, or intentionally add the new source to the schema, and add a collect/validate/finalize regression for a BLOCKING verifier finding.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7607,6 +7679,367 @@ def _emit_coverage_critic_failed_finding(
...
+        return _write_verdict("PASS", [{
+            "check": "input",
+            "message": f"coverage_plan_initial unreadable: {exc}",
+        }])
+    if not isinstance(plan_initial, dict):
+        plan_initial = {}
+
+    available_reviewers: list[str] | None = None
+    if available_path is not None and available_path.exists():
+        loaded, _ = _load_available_reviewers(available_path)
```
shafty023, line 8032: `_load_available_reviewers` returns a diagnostic for malformed or unreadable roster files, but this path discards it and treats the roster as absent. A present corrupted `available_reviewers.json` can therefore skip closed-vocabulary enforcement and still produce PASS. Missing/empty roster can keep no-roster semantics, but a present invalid roster should become a BLOCKING verifier input/roster violation while still exiting 0.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6603,11 +6604,34 @@ def cmd_resolve_coverage(args: argparse.Namespace) -> int:
...
+# the post-process step below. The bare-scalar regex stays
+# conservative (must start with [A-Za-z0-9], YAML-safe chars only)
+# so a typo on a different key doesn't accidentally swallow content
+# from a later line.
 _FRONTMATTER_NAME = re.compile(
-    r"^name\s*:\s*([A-Za-z0-9][A-Za-z0-9_.-]*)\s*$",
-    re.MULTILINE,
+    r"""^name\s*:\s*
+        (?:
+            "(?P<dq>[^"\n]*)"
```
shafty023, line 6619: Quoted `name:` values are no longer constrained to the same identifier grammar as bare names. A repo can declare `name: "../x"`, `name: "bad reviewer"`, or another non-canonical ID; that string is written to `available_reviewers.json`, can be selected by the critic, and will pass closed-vocabulary because it came from the roster. Please validate the unquoted value against the canonical reviewer-id grammar, ideally with a practical length cap, before adding it to the roster.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6653,11 +6683,51 @@ def _scan_agent_definitions(agents_dir: Path) -> tuple[list[str], list[str]]:
...
+            continue
+        # Bound the read at _AGENT_FILE_READ_LIMIT_BYTES so a hostile
+        # `name`-less multi-GB file can't exhaust memory. Frontmatter
+        # is always near the top of the file; truncation cannot
+        # silently produce a wrong `name` because the parser requires
+        # the closing ``---`` boundary to be present.
         try:
-            text = path.read_text()
+            with open(path, "rb") as fh:
+                raw = fh.read(_AGENT_FILE_READ_LIMIT_BYTES + 1)
```
shafty023, line 6707: The per-file read is bounded, but aggregate roster loading is still unbounded: the scanner walks every `.claude/agents/*.md`, writes every parsed reviewer to `available_reviewers.json`, and later includes the full roster in critic input. A PR can add many agent files or very long names to grow CPU, memory, artifacts, and LLM prompt size despite this per-file cap. Please add deterministic caps for scanned files, reviewer count, reviewer-id length, and/or serialized roster/input size, with tests for the cap behavior.

plugins/code-review/tools/python/code_review_helpers.py
wongk: This was already addressed in v2.19.2 (cr-97997's BHB HIGH surfaced the same scope bug — Phase 6's verifier caught its own design defect via the no-PR-comment finding path). The merged code at this line already scopes the closed_vocabulary check to source="critic" entries via:

```python
critic_names = {
    e.get("reviewer", "")
    for e in best_effort_final
    if e.get("source") == "critic"
}
unknown = sorted(critic_names - roster - {""})
```

The violation message also updates to "critic-added reviewers not in AVAILABLE roster" so operators see the scope. The previous tests' gap (always passing rosters containing every reviewer including core/rule labels) is closed by `test_closed_vocabulary_ignores_core_reviewers_outside_roster` and `test_closed_vocabulary_ignores_rule_source_outside_roster` in `TestPLN725Phase6VerifyCoveragePure`, plus an end-to-end version in `TestPLN725Phase6VerifyCoverageCommand.test_core_reviewers_outside_roster_do_not_block`. So this thread's concern is already covered on `main`; resolving.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good catch — and you're right that it's a non-issue at present but worth pinning. Phase 7 (PR #131, v2.20.0, just merged) keeps the call-order strict: stage_15c runs once, BEFORE stage_16. stage_16 reads the verifier verdict from `coverage_verify.json` but does NOT re-invoke the verifier on its output. So the post-arbitration plan (with deferred/dropped buckets) is never fed back through `verify_coverage_plan`.

If Phase 8/9 ever wanted a second verification pass against the post-arbitration plan, the additive check would need a separate mode (or to compare against the consolidate output rather than the original initial plan). To make that contract explicit on disk, the bucket-aware additive rewrite in v2.20.1 enforces `initial.required ⊆ final.required` separately from best_effort preservation, which means the next iteration of this code is forced to be deliberate about what "initial" and "final" mean in a multi-pass world. Resolving — the heads-up landed.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). The additive check is now bucket-aware:

```python
# initial.required ⊆ final.required — demotion to best_effort BLOCKs
required_initial_names = set(_reviewer_names(required_initial))
required_final_names = set(_reviewer_names(required_final))
demoted_or_dropped = sorted(required_initial_names - required_final_names)
if demoted_or_dropped:
    violations.append({"check": "additive", "message": (
        f"initial required reviewers missing from final required[]: "
        f"{demoted_or_dropped} (required reviewers MUST stay required; "
        f"demotion to best_effort or deletion violates the additive-only contract)"
    )})

# initial.best_effort ⊆ final.required ∪ final.best_effort —
# promotion (be → req) stays additive
best_effort_initial_names = set(_reviewer_names(best_effort_initial))
dropped_best_effort = sorted(best_effort_initial_names - final_names)
if dropped_best_effort:
    violations.append({"check": "additive", "message": ...})
```

Two new pure-fn regressions in `TestPLN725Phase6VerifyCoveragePure`:
- `test_additive_blocks_required_demoted_to_best_effort` — exact scenario you flagged: initial.required=[a], final.required=[], final.best_effort=[a] → BLOCKING with "required reviewers MUST stay required" in the message.
- `test_additive_allows_best_effort_promoted_to_required` — promotion direction stays PASS so we don't over-constrain.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). After the top-level list check, every entry in `required[]` and `best_effort[]` is now validated:

```python
for bucket_name in ("required", "best_effort"):
    for idx, entry in enumerate(plan.get(bucket_name, []) or []):
        if not isinstance(entry, dict):
            violations.append({"check": "shape", "message": (
                f"coverage_plan.{bucket_name}[{idx}] is not an object")})
            continue
        reviewer = entry.get("reviewer")
        if not isinstance(reviewer, str) or not reviewer.strip():
            violations.append({"check": "shape", "message": (
                f"coverage_plan.{bucket_name}[{idx}] missing or empty 'reviewer' field")})
if violations:
    return violations  # short-circuit
```

Shape failures short-circuit the additive / closed_vocabulary / critic_* checks — running them on entries we already know are malformed would generate misleading cascading violations.

Four pure-fn regressions plus shared coverage for the short-circuit:
- `test_shape_rejects_non_dict_bucket_entry` (`[{"reviewer": "a"}, "not-a-dict"]`)
- `test_shape_rejects_empty_reviewer_field` (`[{"reviewer": ""}]`)
- `test_shape_rejects_missing_reviewer_field` (`[{}]` — the canonical case you mentioned)
- `test_shape_failure_short_circuits_other_checks` (plan has both `[{}]` AND a real critic_evidence violation — only shape surfaces)

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). Missing/unreadable `coverage_plan.json` and `coverage_plan_initial.json` now both return `verdict: "BLOCKING"` with check `input`, emit the canonical HIGH `coverage-verify-blocking` system finding, and exit 0. The walker's observational posture is preserved (exit code, on_failure: continue) — the artifact verdict is what changes, because that's what Phase 7's gate now reads.

```python
try:
    with open(plan_path) as f: plan = json.load(f)
except (OSError, json.JSONDecodeError) as exc:
    return _write_verdict("BLOCKING", [{
        "check": "input",
        "message": f"coverage_plan unreadable: {exc}",
    }])
```

(Same shape for `coverage_plan_initial.json`.)

The docstring is updated to spell out the split: "Exit code is 0 on PASS and BLOCKING. … Missing or unreadable verifier inputs BLOCK with an ``input`` check (v2.20.1). The earlier behavior — PASS with an advisory ``input`` violation — made 'no plan was verified' indistinguishable from a real PASS in the artifact, and Phase 7's gate would have silently bypassed the cap on every upstream-aborted run."

Regressions: existing `test_missing_plan_degrades_to_pass_with_input_advisory` renamed to `test_missing_plan_blocks_with_input_violation` and inverted, plus new `test_unreadable_coverage_plan_initial_blocks` for the initial-plan path.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). The emitted finding now uses the canonical source registered in `code_review_schema.SOURCES`:

```python
finding = {
    "reviewer": "coverage-verifier",
    "source": "coverage-verifier",
    ...
}
```

Both fields updated for consistency. New regression `test_blocking_finding_uses_canonical_coverage_verifier_source` in `TestPLN725Phase6VerifyCoverageCommand` triggers the BLOCKING path (initial.required dropped → additive violation), reads `agent_coverage-verify-blocking.json` end-to-end, and asserts BOTH `finding["source"] == "coverage-verifier"` AND `finding["source"] in SOURCES`. If a future SOURCES rename or schema-side change drops `coverage-verifier`, the test surfaces it immediately instead of silently losing the finding at stage_22.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). Present-but-malformed roster is now distinguished from missing/empty:

```python
if available_path is not None and available_path.exists():
    loaded, err = _load_available_reviewers(available_path)
    if loaded is None:
        # Present-but-malformed → BLOCK with check "roster".
        # Missing file or empty list still PASS (no-roster semantics).
        return _write_verdict("BLOCKING", [{
            "check": "roster",
            "message": f"available_reviewers unparseable: {err}",
        }])
    available_reviewers = loaded if loaded else None
```

The `roster` check name is distinct from `input` (which is reserved for the verifier's plan inputs) so operators can tell from `coverage_verify.json` which file is the actual fault.

Regression: `test_corrupted_roster_blocks_with_roster_check` in `TestPLN725Phase6VerifyCoverageCommand` writes `[{this is broken` to `available_reviewers.json`, runs `cmd_verify_coverage`, and asserts BLOCKING with the `roster` check in violations plus the canonical HIGH finding artifact on disk. The existing `test_empty_roster_file_skips_closed_vocabulary_check` and `test_missing_roster_file_skips_closed_vocabulary_check` continue to pass — empty/absent still preserve no-roster semantics.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). After the permissive regex matches, the unquoted value is now validated against the canonical agent-name grammar:

```python
_AGENT_NAME_RE = re.compile(r"^[a-z][a-z0-9_-]{0,62}$")

def _parse_agent_name(text: str) -> str | None:
    ...
    value = (match.group("dq") or match.group("sq") or match.group("bare") or "").strip()
    if not value:
        return None
    if not _AGENT_NAME_RE.match(value):
        return None
    return value
```

The grammar mirrors the existing `_REVIEWER_ID_RE` (defined later in the same module for collect-findings) which checks against `make_finding_id`'s `^[a-z][a-z0-9_-]*$` requirement. Keeping them at the same level of strictness guarantees that any name surviving the loader will also survive `make_finding_id` downstream — without this alignment a name that passed here could still poison the collect-findings stage. The 63-char cap mirrors DNS-label sizing; every actual project agent in this repo is lowercase-with-hyphens well under that.

Six regressions in `TestPLN725Phase5LoaderGrammarAndCaps`:
- `test_quoted_path_traversal_rejected` (`name: "../x"`)
- `test_quoted_whitespace_rejected` (`name: "bad reviewer"`)
- `test_quoted_uppercase_rejected` (`name: "FooBar"` — would crash `make_finding_id`)
- `test_quoted_dot_rejected` (`name: "foo.bar"`)
- `test_quoted_exceeding_length_cap_rejected` (64 chars → None)
- `test_quoted_at_length_cap_accepted` (63 chars → accepted; pins the boundary)
- `test_bare_uppercase_still_rejected` (no asymmetry between bare and quoted paths)

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in v2.20.1 (pushed to PR #131). Two deterministic caps applied at scan time, before any bytes from the cap-exceeding files are read:

```python
_AGENTS_DIR_MAX_FILES = 200   # truncate file scan
_ROSTER_MAX_ENTRIES = 200      # cap resulting roster

# In _scan_agent_definitions:
all_paths = sorted(agents_dir.glob("*.md"))
if len(all_paths) > _AGENTS_DIR_MAX_FILES:
    warnings.append(f"agents dir exceeds {_AGENTS_DIR_MAX_FILES}-file cap …")
    all_paths = all_paths[:_AGENTS_DIR_MAX_FILES]
for path in all_paths:
    if len(reviewers) >= _ROSTER_MAX_ENTRIES:
        warnings.append(f"roster size cap reached at {_ROSTER_MAX_ENTRIES} …")
        break
    ...
```

Reviewer-id length is bounded by the 63-char grammar cap (the previous reply); roster JSON size and downstream critic prompt size are now bounded by `_ROSTER_MAX_ENTRIES × 63 ≤ ~12.6 KiB` of names. Excess files truncate deterministically in sorted-filename order so the cap behavior is reproducible across reviews. Both warnings surface so operators see why a short roster came back.

Two regressions in `TestPLN725Phase5LoaderGrammarAndCaps`:
- `test_scan_caps_files_to_max` — seeds `_AGENTS_DIR_MAX_FILES + 5` files, asserts the roster is exactly `_AGENTS_DIR_MAX_FILES` long and the truncation warning fires.
- `test_scan_caps_roster_size` — pins the roster-size cap independently in case the file-count and roster caps ever diverge.

PR #131, opened 2026-06-02
feat(code-review): PLN-725 Phase 7 arbitrate-budget + BLOCKING gate (v2.20.0)

wongk:
## Summary

- Enables `stage_16_arbitrate_budget` in the run plan. The cost-arbitration cap (held disabled since v2.16.x while Phase 6 caught up) now runs on every review: drops overflow `required` reviewers with `budget-exceeded` findings, prunes lowest-priority best-effort, computes the final `bha_partitions` count.
- Wires the Phase 6 verifier's BLOCKING verdict (from `stage_15c_verify_coverage`) into `cmd_arbitrate_budget` via a new `--coverage-verify` arg. On BLOCKING, arbitration short-circuits: input plan flows through unchanged, `budget.gated_by_verify: true`, `arbitrate_status: "blocked_by_verify"`. PASS/missing/malformed verify file all run normal arbitration so the verifier stays observational from the gate's perspective.
- Out of scope (deferred to Phase 8): `stage_20_spawn_reviewers` still consumes the static reviewer table in `start.md`. The orchestrator rewire to read `coverage_plan.json` is the natural next phase; this PR delivers the BLOCKING-verdict teeth without that larger surface change.

## Test plan

- [x] `pytest plugins/code-review/tools/python/` — 902 passed, 5 skipped (added `TestPLN725Phase7ArbitrateBudgetGate` × 6; renamed `test_stage_16_stays_disabled_in_phase_4` → `test_stage_16_enablement_history`; updated `test_arbitrate_budget_gated_on_phase_7` → `test_arbitrate_budget_enabled_in_phase_7` to assert enabled + `--coverage-verify` wired + `on_failure: abort`).
- [x] `pytest plugins/` — 1641 total, 0 failures.
- [x] `ruff check plugins/code-review/tools/python/` — clean.
- [x] `pyright plugins/code-review/tools/python/` — 0 errors.

## Why BLOCKING short-circuits instead of aborting

`on_failure: "abort"` is intentional for stage_16 — a real I/O or shape error should halt the pipeline. But BLOCKING is not an error: it's a deterministic signal from the verifier that the coverage plan failed a contract check. Treating it as failure would conflate two concepts (artifact verdict vs. process error) and would force every BLOCKING run into a full pipeline halt before findings are even collected, costing operators the rest of the run's signal. The short-circuit (exit 0, preserves rule floor, annotates the plan) is the right shape: downstream stages run as normal, finalize-result can show the gate in the summary, and the BLOCKING finding from stage_15c is already in the run's findings bucket so it surfaces in the presenter footer.


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -8020,19 +8178,33 @@ def _write_verdict(verdict: str, violations: list[dict[str, str]]) -> int:
...
+        }])
 
     available_reviewers: list[str] | None = None
     if available_path is not None and available_path.exists():
-        loaded, _ = _load_available_reviewers(available_path)
-        # Roster read errors degrade to None (skip closed-vocabulary check)
-        # rather than emitting a violation — the verifier's job is to
-        # check the plan, not the roster file shape.
+        loaded, err = _load_available_reviewers(available_path)
+        if loaded is None:
```
thadeusb, line 8248: _load_available_reviewers only returns None for a bad top-level type. A present-but-wrong-shape roster like {"reviewers": [...]} (wrong key) or [1,2,3] comes back as ([], None), so `loaded is None` is False and it falls through to `available_reviewers = None`, i.e. no-roster PASS, not the BLOCKING you just added for the unparseable case. The wrong-key one is a pretty realistic operator hand-edit. Intended, or should present-but-empty-after-parse also block?

thadeusb (approved):
Gate wiring and the verifier hardening hang together, and catching the wrong 'coverage-verify' source before it hit stage_22 validation was a good find. One roster edge I want to check inline, nothing blocking.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good catch — that's a real gap. Fixed in v2.20.3 (`cbc6771`). `_load_available_reviewers` now returns `(None, diagnostic)` for the three present-but-wrong-shape cases:

```python
# dict missing the "available" key entirely (the hand-edit you flagged)
if "available" not in raw:
    return None, (
        "Error: available_reviewers dict missing required 'available' "
        f"key (got keys: {sorted(raw.keys())[:10]})"
    )
# dict with inner non-list (e.g. {"available": "foo"})
if not isinstance(inner, list):
    return None, (f"Error: ... must be a list (got {type(inner).__name__})")
# list whose entries are all non-strings (e.g. [1,2,3])
filtered = [a for a in inner if isinstance(a, str) and a]
if not filtered:
    return None, "Error: ... contains no usable string entries"
```

Truly-empty rosters (`[]` and `{"available": []}`) still return `([], None)` and keep no-roster skip semantics — the fix doesn't over-rotate. Five new unit regressions in `TestLoadAvailableReviewers` (wrong-key, list-of-non-strings, inner-list-of-non-strings, plus the two truly-empty positive cases that pin the discriminator), and one end-to-end `test_wrong_key_roster_blocks_with_roster_check` in `TestPLN725Phase6VerifyCoverageCommand` that writes `{"reviewers": [...]}` to disk and asserts the full pipeline emits BLOCKING with the `roster` check.

Validation: 926 passed (+6), ruff clean, pyright 0 errors.

PR #132, opened 2026-06-03
feat(code-review): relax line-scope filter for companion-change findings (v2.21.0)

wongk:
## Summary

The pre-v2.21 `_filter_scope_and_range` unconditionally discarded every P2+ finding whose line was outside the file's changed range as `DISCARD_LINE_NOT_CHANGED`. That was a pre-verification heuristic from before PLN-722's per-finding verifier shipped — and modern reviewer agents legitimately surface companion-change findings (a function signature change in the diff window leaving stale sibling call sites just outside it is the canonical case). The unconditional discard was silently killing real bugs.

This PR replaces the unconditional drop with a confidence floor:

- Out-of-hunk P2+ findings survive when `confidence ≥ out_of_hunk_confidence_floor` (default `0.80`).
- Survivors get tagged `out_of_hunk_kept: true` so presenters can label them as companion-change without re-deriving hunk membership.
- New `DISCARD_OUT_OF_HUNK_LOW_CONFIDENCE` reason for sub-floor drops keeps the stats separable from the in-hunk discard path.
- New stats keys `kept_out_of_hunk` and `discarded_out_of_hunk_low_confidence` (plus the echoed `out_of_hunk_confidence_floor`) give A/B observability against historical strict runs.
- Retired `discarded_line_not_changed` key preserved at `0` for one release so external telemetry dashboards see the transition without crashing.
- New `--settings` arg on `cmd_validate` for test isolation; floor is operator-tunable via the same `code-review.json` that already hosts `bha_unified_threshold_loc`.

## Design notes

- **Default floor (0.80)** is intentionally tighter than the in-hunk floor (`CONFIDENCE_DISCARD_THRESHOLD = 0.5`). Out-of-hunk findings have to demonstrate causation, not just colocation.
- **P1 (BLOCKING/HIGH)** findings are unaffected — they pass regardless of hunk membership, same as before. Only P2+ goes through the new gate.
- **Kill switch**: setting the floor to `1.0` effectively restores pre-v2.21 strict behavior. Setting `0.0` lets every out-of-hunk P2+ through and leans on the PLN-722 verifier for noise filtering.
- **Verifier still runs.** Per-finding verification (stage_23) applies on top — anything that surfaces here can still be REJECTED by the verifier on a second pass.

## Test plan

- [x] pytest plugins/code-review/tools/python/ — 937 passed, 5 skipped (+11 net new).
- [x] ruff + pyright clean.
- [x] pytest plugins/ — full suite passes.


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1358,7 +1413,20 @@ def _filter_scope_and_range(
...
         in_range = _line_in_range(line, added) or _line_in_range(line, removed)
         if not in_range and priority > 1:
-            discarded.append({"finding": finding, "reason": "DISCARD_LINE_NOT_CHANGED"})
+            # v2.21.0: relaxed from unconditional discard to
+            # confidence-gated. The floor is intentionally tighter
+            # than the in-hunk floor because out-of-hunk findings need
+            # to demonstrate causation. Findings that survive get
+            # tagged so presenters can label them as companion-change
+            # without re-deriving hunk membership.
+            if confidence >= out_of_hunk_confidence_floor:
```
thadeusb: The whole risk argument here is "the verifier does noise filtering on top," but a MEDIUM out-of-hunk finding at confidence >= 0.85 never reaches it. _should_verify returns `confidence < 0.85` for MEDIUM (line 2078), so anything at or above that skips stage_23 and lands in the presenter as a confirmed finding. Your own canonical case, the 0.9 companion-change finding in test_out_of_hunk_high_confidence_survives, is exactly one of those. So the safety net only actually covers the 0.80-0.84 band, and the high-confidence findings you're most deliberately letting through get no second opinion, which is backwards from what you'd want for cross-region causation claims an LLM is shaky on. If you want the verifier to really backstop this, add an `out_of_hunk_kept` check to the tier table so these always verify.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1553,29 +1621,65 @@ def cmd_validate(args: argparse.Namespace) -> int:
...
     validated = _group_cross_file(deduped)
 
     cross_file_grouped = sum(
         len(f.get("other_locations", [])) for f in validated
     )
+    # v2.21.0 telemetry: count out-of-hunk findings that survived the
+    # relaxation so operators can A/B the change against historical
+    # runs where DISCARD_LINE_NOT_CHANGED hit every one of these.
+    kept_out_of_hunk = sum(
+        1 for f in validated if f.get("out_of_hunk_kept") is True
```
thadeusb: This is the only place out_of_hunk_kept gets set, but _normalize_findings passes the raw reviewer dict through verbatim, so an in-hunk finding that arrives already carrying out_of_hunk_kept: true keeps it and ticks kept_out_of_hunk here. Low odds today since it's a brand new internal field, but the telemetry and the future presenter label both trust it. Set it False on the normal exit path so we own the field, not the reviewer.

thadeusb (approved):
Clean relaxation and the settings validation is thorough. One thing to chew on: the high-confidence out-of-hunk findings this is built to admit are exactly the ones stage_23 skips, so the verifier-backstop argument has a hole. Flagged inline, not blocking.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Sharp catch. You're right that the relaxation makes the canonical 0.9 companion-change finding *less* covered, not more — the verifier was the load-bearing safety net argument in the design and it wasn't actually firing.

Fixed in v2.21.1 (`1ec780c`). `_needs_verification` at [code_review_helpers.py:2167-2188](plugins/code-review/tools/python/code_review_helpers.py#L2167-L2188) now adds an `out_of_hunk_kept: True` check ahead of the severity tiers so confidence-based skips never apply:

```python
# v2.21.1 backstop: every out-of-hunk-kept finding gets verified.
# Placed after the deterministic-producer guards (Hygiene findings
# never carry the tag in practice, but the ordering keeps the
# never-verify producers absolute) and before the severity tiers
# so confidence-based skips don't apply.
if finding.get("out_of_hunk_kept") is True:
    return True
```

Ordering matters and is pinned by `test_out_of_hunk_kept_does_not_resurrect_hygiene_or_injection`: Hygiene + injection-detector findings stay unverifiable even if they somehow carry the tag (Hygiene comes from a regex catalogue, not from agent reviewers, so in practice it never does — but pinning the order keeps a future refactor honest). The canonical case is locked by `test_out_of_hunk_kept_always_verified_even_at_high_confidence` — `_needs_verification` on `{severity: MEDIUM, confidence: 0.9, out_of_hunk_kept: True}` now returns True (without the backstop it returned False).

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fair point — agreed the validator should own the field rather than passing reviewer values through. Fixed in v2.21.1 (`1ec780c`).

The fix pops the key on every diff-scope exit path that isn't "out-of-hunk and survived":

```python
# In-hunk diff-scope path
finding.pop("out_of_hunk_kept", None)
result.append(finding)

# system/pr_metadata path (same treatment — the field is meaningless
# for them; the schema convention should hold universally)
finding.pop("out_of_hunk_kept", None)
result.append(finding)
```

Chose pop over `= False` so the schema stays "tag present (and True) IFF companion-change". That keeps the golden fixtures unchanged (none of them carry the field) and avoids polluting every persisted envelope with a meaningless `out_of_hunk_kept: False`. The reviewer-supplied case is locked by `test_validator_overrides_reviewer_supplied_out_of_hunk_kept` — pre-populates the field on an in-hunk finding and asserts both the field absence post-validation AND `kept_out_of_hunk == 0`.

I also realized while doing this that `kept_out_of_hunk` was undercounting in a different way — it was summing on the post-grouping `validated` set, but `_group_cross_file` absorbs siblings into `other_locations[]` where the tag is lost. The counter now sums on `filtered` (post-filter, pre-grouping). Regression: `test_kept_out_of_hunk_counts_grouped_companions`.

PR #133, opened 2026-06-03
feat(code-review): PLN-725 Phase 8 spawn-spec gate at stage_20 (v2.22.0)

wongk:
## Summary

PLN-725 Phase 8 — closes the deterministic-coverage loop. Before this PR, the post-arbitrate `coverage_plan.json` was effectively ignored at spawn time: `stage_20_spawn_reviewers` walked a static reviewer table baked into `start.md`, and the rule-resolved + verified coverage signal never actually shaped the spawned fleet. Phase 8 wires the plan into the orchestrator's dispatch path.

- **New `derive-spawn-spec` subcommand** translates the post-arbitrate `coverage_plan.json` + `partitions.json` + `route.json` into a flat `spawn_spec.json` keyed by canonical AGENT_IDs (`bha_p<N>`, `bhb`, `auditor`, `premise`, `domain_<N>`, `fast`). BHA expands per partition with test-only model routing; `best_effort[]` critics map to sequential `domain_<N>` slots; `test_quality` surfaces in `skipped[]` (deferred to PLN-723).
- **New `stage_19b_derive_spawn_spec` walker entry** between `stage_17_partition` and `stage_20_spawn_reviewers`. `on_failure: continue` — a derive bug writes a sentinel spec with `arbitrate_status: "fallback"` and `stage_20` walks the static table instead. Review is never blocked by a spawn-spec failure.
- **`start.md` Reviewer Fleet section** now opens with the spawn-spec consumption protocol; the static table is preserved as the documented fallback path. Fast-path passthrough (`route.fast_path == true` → single `fast` agent) and the Phase 7 `gated_by_verify` flag both propagate so the present step can warn when arbitration was bypassed.

## Test plan
- [x] `pytest plugins/` — 1693 passed, 5 skipped (+12 new Phase 8 tests)
- [x] `pyright plugins/code-review/tools/python/code_review_helpers.py` — 0 errors
- [x] `TestPLN725Phase8DeriveSpawnSpec` covers bucket-to-spec mapping, critic mapping to `domain_<N>`, fast-path passthrough, BLOCKING-verify propagation, all-files-cached BHA skip, unknown-reviewer skip, route-model overrides, missing-route safe defaults
- [x] `TestPLN725Phase8StageGraph` pins walker wiring (subcommand, depends_on, `on_failure: continue`, stage_20 dep on stage_19b, array ordering)
- [x] `TestPrepareRun` stage count expectation updated 35 → 36


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -10425,6 +10466,391 @@ def cmd_arbitrate_budget(args: argparse.Namespace) -> int:
...
+                "partitioned": False,
+                "patches_file": role_cfg["patches_template"],
+                "source": source or "core",
+                "bucket": bucket,
+            })
+            return
+        # Unknown reviewer name. If the rule-resolved entry was a critic
+        # (closed-vocabulary checks have already passed by stage_15c),
+        # treat it as a domain critic and assign a sequential ID.
+        if source == "critic":
```
thadeusb: This only rescues source=='critic', but rule-resolved domain reviewers come through with source=='rule' and fall straight to the unknown_reviewer skip below, so they never spawn.

Two ways that happens for real: a canonical coverage[] rule like {reviewer: "ts-expert", required: true, triggers: [...]} resolves to a required[] entry with source 'rule' (resolve_coverage sets that at the rule loop), and any legacy moduleCritics[] entry, which migrate_legacy_module_critics turns into coverage[] rules that also resolve to source 'rule'. Both are roster-valid so stage_15c's closed_vocabulary check passes them, and arbitrate-budget keeps them in required/best_effort. The old static-table path spawned the moduleCritics ones via route.json -> domain_critics, so for those repos this is a regression: the critic that used to run now sits in skipped[] and start.md tells stage_20 not to re-add it. No warning surfaces either.

The comment right here says "if the rule-resolved entry was a critic" but the check is source=='critic', which rule-resolved entries never are. Widen it to treat any non-core roster reviewer as a domain_<N> (source in {"rule","critic"}). And test_unknown_required_reviewer_lands_in_skipped currently pins the wrong behavior with a source:'rule' entry, so that test asserts the bug.

thadeusb (approved):
Core + LLM-critic path translates cleanly. One real gap: derive only rescues source=='critic', so rule-resolved domain reviewers (coverage[] rules, migrated moduleCritics) carry source 'rule' and land in skipped[] as unknown_reviewer, never spawning. For moduleCritics repos that's a straight regression from the old route path. Fixable without reworking the approach.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -10425,6 +10466,391 @@ def cmd_arbitrate_budget(args: argparse.Namespace) -> int:
...
+                # BHA expands per partition. When partitions is empty
+                # (all cached or docs-only post-arbitrate), no BHA spawn.
+                if not partitions:
+                    skipped.append({
+                        "reviewer": reviewer,
+                        "bucket": bucket,
+                        "reason": "no_partitions",
+                    })
+                    return
+                for part in partitions:
```
shafty023: `coverage_plan["budget"]["bha_partitions"]` is the post-arbitrate cap, but this loop emits one BHA descriptor for every partition. When arbitration reserves fewer BHA slots than the partitioner produced, or sets `bha_partitions: 0`, stage_20 will still spawn those BHA agents and bypass the final coverage plan. Please derive the BHA slice from the final budget, or feed that budget into partitioning, and add a regression where partitions exceed `bha_partitions`.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -10425,6 +10466,391 @@ def cmd_arbitrate_budget(args: argparse.Namespace) -> int:
...
+                "model": "sonnet",
+                "partitioned": False,
+                "patches_file": "patches_all.txt",
+                "source": "critic",
+                "bucket": bucket,
+                "priority": int(entry.get("priority", 2)),
+            })
+            critic_index += 1
+            return
+        skipped.append({
```
shafty023, line 10807: Entries from `required[]` can reach this path and only land in `spawn_spec.skipped[]`; `start.md` then tells stage_20 not to re-add skipped reviewers, and no coverage gap is written for the final verdict. That lets an unspawnable required reviewer disappear instead of failing closed. Please synthesize a required Coverage gap for non-benign required skips, and keep benign cases like deferred `test_quality` clearly separated.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -10425,6 +10466,391 @@ def cmd_arbitrate_budget(args: argparse.Namespace) -> int:
...
+                "bha_count": 0,
+                "domain_critic_count": 0,
+                "from_required": 0,
+                "from_best_effort": 0,
+            },
+            "generated_at": now_iso,
+        }
+        return _write_spawn_spec(spec, output_path)
+
+    partitions_blob = _read_optional_json(Path(args.partitions), None)
```
shafty023: Missing or malformed `partitions.json` is collapsed into `partitions = []`, which is indistinguishable from a valid empty partition set. In standard flow that suppresses BHA as `no_partitions` while still dispatching the rest of the spec, instead of taking the documented fallback path. Please only treat an empty partition list as valid after successfully reading a well-formed artifact; otherwise emit a fallback sentinel or a blocking coverage gap.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9812,17 +9812,58 @@ def _build_run_plan_stages(
...
+            "subcommand": "derive-spawn-spec",
+            "args": [
+                "--cr-dir", cr_dir,
+                "--coverage-plan", f"{cr_dir}/coverage_plan.json",
+                "--partitions", f"{cr_dir}/partitions.json",
+                "--route", f"{cr_dir}/route.json",
+            ],
+            "stdout": None,
+            "expected_outputs": [f"{cr_dir}/spawn_spec.json"],
+            "depends_on": [
```
shafty023: Gate B says `FAST_PATH == true` skips `stage_17_partition`, but this new stage depends on `stage_17_partition`, and `stage_20` also keeps that dependency. A dependency-aware walker can therefore skip the derive/spawn path exactly when it needs the `fast` descriptor. Please adjust the dependency model or materialize a fast-path partition sentinel, and add a run-plan test that fast path reaches `stage_20`.

plugins/code-review/commands/start.md
```diff
@@ -309,7 +310,24 @@ These notes annotate the run-plan stages with anything not obvious from the plan
...
+
+- `agent_id` → orchestrator-assigned ID; agent writes to `<CR_DIR>/agent_{agent_id}.json`.
+- `model` → resolved per-agent model string (already accounts for BHA test-only routing and route.json overrides — do not re-derive).
+- `partitioned: true` + `partition_id` → patches file is `patches_p{partition_id}.txt`; use the partition's `files[]` from `partitions.json` for `<files_assigned>`.
+- `partitioned: false` → patches file is `patches_all.txt`; `<files_assigned>` is the full `files_to_review` list.
+- `source` ∈ {`core`, `critic`, `fast_path`} → picks the per-agent prompt suffix below (`bug_hunter_a` → BHA, `bug_hunter_b` → BHB, `unified_auditor` → Auditor, `premise_reviewer` → Premise, `critic` → Domain Critic, `fast_path` → Fast Path).
+- `spec.fast_path: true` → spec emits exactly one agent (`agent_id: "fast"`); skip the standard-flow tables and use the Fast Path suffix below.
+- `spec.gated_by_verify: true` → arbitration was bypassed by a BLOCKING verify verdict from stage_15c (the canonical finding already lives in `agent_coverage-verify-blocking.json`). Spawn the spec as-is; surface a one-line warning in the present step that the review ran against an unbudgeted plan.
```
shafty023: `gated_by_verify` means the coverage verifier already rejected the plan, but this tells stage_20 to spawn the unverified spec as-is. Since `derive-spawn-spec` turns any `source: "critic"` name into a domain agent, a BLOCKING closed-vocabulary/evidence/cap failure can still drive task dispatch. Please gate spawn on a verified PASS or sanitize the plan to a known-safe subset while leaving the verifier's coverage gap to block the verdict.

plugins/code-review/commands/start.md
```diff
@@ -309,7 +310,24 @@ These notes annotate the run-plan stages with anything not obvious from the plan
 
 ## Reviewer Fleet (stage_20_spawn_reviewers)
 
-This stage runs when the walker reaches `stage_20`. It branches on `FAST_PATH` from Gate B.
+This stage runs when the walker reaches `stage_20`.
+
+**PLN-725 Phase 8 — spawn-spec consumption (preferred path).** Before walking the static tables below, Read `<CR_DIR>/spawn_spec.json`. If the file exists and `arbitrate_status != "fallback"`, dispatch one Task per entry in `agents[]`, using the descriptor fields directly:
```
shafty023, line 316: The new spec gives stage_20 the exact `agents[]` it intended to run, but the existing validation gate still only checks for the presence of some `agent_*.json`. If one required descriptor fails after another agent writes output, the walk can continue without a required-reviewer coverage gap. Please compare `spawn_spec.agents[]` against `agent_{agent_id}.json` outputs after stage_20 and emit blocking coverage gaps for missing required agents.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in `80f0ed9` (v2.22.2). You're right — `_resolve_coverage` emits `source: "rule"` for every rule-matched entry (line 6633), so any canonical `coverage[]` rule or migrated `moduleCritics[]` entry naming a non-core reviewer was landing in `skipped[]` with `reason: "unknown_reviewer"` instead of spawning.

The dispatch is now widened:

```python
if source in {"rule", "critic"}:
    agent_id = f"domain_{critic_index}"
    seen_agent_ids.add(agent_id)
    agents.append({
        "agent_id": agent_id,
        "reviewer": reviewer,
        "model": "sonnet",
        "partitioned": False,
        "patches_file": "patches_all.txt",
        "source": source,  # preserved: presenters distinguish rule vs critic
        "bucket": bucket,
        "priority": int(entry.get("priority", 2)),
    })
    critic_index += 1
    return
```

Knock-on fixes in the same commit:
- `domain_critic_count` telemetry now sums both `rule` + `critic` sources (both spawn as domain agents).
- `SPAWN_SPEC_SOURCES` in `code_review_schema.py` adds `"rule"` as a first-class value alongside `core`/`critic`/`fast_path`; `SCHEMA.md §6b` updated to match.
- `start.md` two-level dispatch bullet records that both sources select the Domain Critic suffix.
- The misnamed `test_unknown_required_reviewer_lands_in_skipped` (which used `source: "rule"` and was pinning the bug) is renamed `test_unknown_source_lands_in_skipped` and switched to use an empty source — only the genuine defense-in-depth branch fires now.
- New `test_rule_resolved_domain_reviewer_spawns_as_domain_critic` pins the correct behavior end-to-end with a `ts-expert` rule in `required[]` and a legacy-migrated entry in `best_effort[]` — both spawn as `domain_<N>`, source is preserved, neither lands in `skipped[]`.

1698 passed, ruff clean, pyright 0 errors.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `70c51db` (v2.22.3). `_derive_spawn_agents_from_plan` now takes a `bha_partitions_cap` keyword arg sourced from `coverage_plan.budget.bha_partitions`; when the cap is < len(partitions), the first cap partitions spawn (prefix-take preserves the partitioner's bin-packed ordering) and the rest land in `skipped[]` with `reason: "budget_capped"`, `partition_id`, `budget_cap`, and `partition_count` for operator visibility. `cap == 0` (docs-only post-arbitrate) suppresses all BHA spawns regardless of partition count. Regression test `TestPLN725Phase8DeriveSpawnSpecBudgetCap` pins three boundary cases: cap < count, cap == 0, cap == count.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `70c51db` (v2.22.3). `cmd_derive_spawn_spec` now synthesizes coverage-gap findings for every `skipped[]` entry with `bucket == "required"` and a non-benign reason, then appends them to `coverage_gaps.json` (preserving any pre-existing `arbitrate-budget` entries — append-not-overwrite contract). The benign set is `{deferred_pln723, no_partitions, gated_by_verify}` — the first is the PLN-723 placeholder; the second is all-cached/docs-only legitimate omission; the third is already surfaced via the canonical BLOCKING finding from stage_15c.

I also added a **runtime symmetric pair**: new `stage_20b_verify_spawn` runs after stage_20, globs `agent_*.json`, and emits one coverage-gap finding per required descriptor with no on-disk output. Together stage_19b catches required reviewers the spec couldn't describe; stage_20b catches required agents that crashed at runtime — the "required reviewer missing" failure mode is now observable in both halves.

Tests: `TestPLN725Phase8RequiredCoverageGaps` (unknown emits, deferred doesn't, gated doesn't, append-not-overwrite) + `TestPLN725Phase8VerifySpawn` (all-present, missing-required, missing-best-effort, fallback no-op, missing no-op).

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `70c51db` (v2.22.3). Missing or malformed `partitions.json` now emits a fallback sentinel (`arbitrate_status: "fallback"`, `fallback_reason: "partitions_missing_or_malformed"`) instead of collapsing into the present-but-empty path. The all-files-cached / docs-only case (where `partitions.json` exists and contains `{"partitions": []}`) still flows through the documented `no_partitions` skip — that's a legitimate "BHA has nothing to do" outcome. The reviewer's distinction was right: a missing file is a stage_17 crash signal, not an empty partition set.

This required inverting the v2.22.1 test (`test_missing_partitions_file_treated_as_no_partitions`) that was pinning the old behavior — that test was added at the same `domain_4_f1` reviewer's request in cr-14619 and was specifically asking the missing-file path be tested, but the contract it pinned was the wrong contract. New test `test_missing_partitions_file_emits_fallback_sentinel` pins the corrected contract; the docstring explains the v2.22.1 → v2.22.3 reversal explicitly so future readers see why.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `70c51db` (v2.22.3). `stage_19b_derive_spawn_spec` walker entry now declares only `stage_16_arbitrate_budget` as a hard dep; `stage_17_partition` is intentionally NOT a hard dep so Gate B's fast-path branch (which skips stage_17) can still reach stage_19b → stage_20 with the fast descriptor. The cmd handles the missing partitions.json internally: fast-path returns before reading partitions; non-fast-path falls through to the new `partitions_missing_or_malformed` fallback sentinel (per the parallel fix on the same review).

New regression test `test_fast_path_reaches_stage_20_without_stage_17` pins the invariant — stage_19b precedes stage_20 in the run plan array, and stage_19b.depends_on does NOT contain stage_17_partition. The chain back from stage_16 reaches setup through stages that all run regardless of fast_path.

plugins/code-review/commands/start.md
wongk: Fixed in `70c51db` (v2.22.3). Went with the "sanitize to known-safe subset" option: under `gated_by_verify: true`, `cmd_derive_spawn_spec` now drops every non-`source: "core"` entry from required[] + best_effort[] before passing the plan to the agent resolver. The dropped entries land in `skipped[]` with `reason: "gated_by_verify"` and the entry's original `source` preserved (rule vs critic) so presenters can distinguish operator-configured suppressions from LLM-proposed ones.

This keeps Phase 7's "review still runs" semantics (the canonical static fleet — BHA per partition, BHB, Auditor, Premise — still spawns) while refusing to action any closed-vocabulary / shape / evidence / cap violation the verifier flagged. The canonical BLOCKING finding from stage_15c remains the operator-facing signal of the rejection — we don't double-emit. start.md `gated_by_verify` bullet rewritten to spell out the sanitization contract.

Tests: `TestPLN725Phase8DeriveSpawnSpecBlockingSanitization` covers mixed core+rule+critic (only core spawns; both rule and critic land in skipped with source preserved) and the no-op path (BLOCKING with only core reviewers → no spurious gated_by_verify skips).

plugins/code-review/commands/start.md
wongk: Fixed in `70c51db` (v2.22.3). New `stage_20b_verify_spawn` walker stage runs after `stage_20_spawn_reviewers`: reads `spawn_spec.json`, globs `<CR_DIR>/agent_*.json`, identifies every descriptor with `bucket == "required"` that has no on-disk output, and appends one coverage-gap finding per missing required agent to `coverage_gaps.json`. The walker dep chain has `stage_21_collect_findings` depending on `stage_20b_verify_spawn` so the gap findings land before `cmd_finalize_result` reads them into the canonical envelope.

Missing best-effort descriptors are recorded in `spawn_verification.json` for telemetry but emit no finding — best-effort omissions are budget-driven, not coverage gaps. No-ops cleanly when the spec is missing (`spec_missing`), marks fallback (`spec_fallback`), or contains zero agents (`spec_empty`) so the static-table path doesn't trigger spurious gaps.

The reviewer's framing of stage_19b + stage_20b as a symmetric pair captures it exactly: derive-spawn-spec catches required reviewers the spec couldn't describe; verify-spawn catches required agents that crashed at runtime. Both producers write to the same `coverage_gaps.json` so the verdict computation sees one unified list.

Tests: `TestPLN725Phase8VerifySpawn` covers all-required-present (no gap), missing-required (gap appended), missing-best-effort (no gap), fallback no-op, missing no-op. `test_stage_20b_verify_spawn_present_and_wired` + `test_collect_findings_depends_on_verify_spawn` pin the walker wiring.

PR #134, opened 2026-06-03
feat(code-review): PLN-725 Phase 9 presenter spawn-spec consumption (v2.23.0)

wongk:
## Summary

PLN-725 Phase 9 — operator-facing side of the spawn-spec pipeline. Before this PR, the `present-local` skill and `github-review.md` derived the Reviewers + Model Routing lines from a hardcoded static list keyed on `FAST_PATH`. The spawn-spec (Phase 8 v2.22.0) and runtime verification artifact (v2.22.3) carried the actual deterministic-coverage signal at dispatch and runtime — and the presenter ignored both.

- **New `render-fleet-summary` subcommand** on `code_review_helpers.py` reads `spawn_spec.json` + `spawn_verification.json` + `route.json` and emits a deterministic 2–6 line markdown block: **Reviewers** (actual fleet that spawned, with rule-resolved vs LLM-proposed split on domain critics), **Model Routing** (per-agent assignments from route.json), **Fleet** (`N intended | N ran | N required missing`), and a conditional **Notes** block for non-default outcomes.
- **`present-local` SKILL.md and `prompts/github-review.md`** now invoke the renderer at the slot previously occupied by hand-authored Reviewers / Model Routing prose. The Notes block surfaces fleet outcomes the presenters previously had no way to show — operators reading the local report or PR summary see _Arbitration bypassed: suppressed ts-expert (rule), graphql-architect (critic)_ or _1 required reviewer(s) did not produce output: Unified Auditor_ without having to scroll into Verifier Stats or read `coverage_gaps.json` by hand.
- **Domain critic provenance** surfaces inline: `2 domain critics (1 rule-resolved, 1 LLM-proposed)` so operators can tell operator-configured coverage apart from one-off LLM proposals — same audit-trail rationale as the v2.22.2 source-preservation fix.
- **Fallback path is explicit:** when the renderer reports `spawn-spec unavailable` or `spawn-spec fell back`, the presenter walks the static fleet table in `start.md` (the documented source of truth for that case) and never fabricates a reviewer list.

## Test plan
- [x] `pytest plugins/` — 1730 passed, 5 skipped (+15 net new Phase 9 tests)
- [x] `pyright plugins/code-review/tools/python/code_review_helpers.py` — 0 errors
- [x] `TestPLN725Phase9RenderFleetSummaryHappyPath` — canonical 5-agent flow, multiple BHA partitions (`× N`), rule vs critic provenance, single critic inline, fast-path branch
- [x] `TestPLN725Phase9RenderFleetSummaryNotes` — BLOCKING sanitization 🛡️, runtime missing required ⚠️ (display-name mapping verified), budget cap, PLN-723 deferral ℹ️, malformed-plan skips, four-note composition order
- [x] `TestPLN725Phase9RenderFleetSummaryFallbacks` — missing spec, fallback sentinel + reason, missing verification (runtime tally unavailable), missing route (Model Routing omitted, not fabricated)


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+        if rule_critics:
+            provenance.append(f"{len(rule_critics)} rule-resolved")
+        if llm_critics:
+            provenance.append(f"{len(llm_critics)} LLM-proposed")
+        suffix = (
+            f" ({', '.join(provenance)})" if provenance else ""
+        )
+        if domain_total == 1:
+            critic_entry = rule_critics[0] if rule_critics else llm_critics[0]
+            critic_name = str(critic_entry.get("reviewer", "domain critic"))
```
shafty023, line 11484: This renderer writes reviewer names from `spawn_spec.json` directly into Markdown that can land in the GitHub summary. Rule-resolved reviewer names originate from repo-controlled `critic-gates.json`, and the current coverage rule validator only checks for a non-empty string, so names containing newlines, headings, HTML, or mentions can inject misleading content into the summary. Please validate reviewer IDs before they reach the spawn spec and/or escape artifact-derived strings before rendering, with adversarial tests for Markdown/mention injection.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+        reason = str(spec.get("fallback_reason", "unknown"))
+        out_lines.append(
+            f"**Reviewer Fleet:** spawn-spec fell back (`{reason}`); the "
+            "orchestrator walked the static reviewer table for this run.",
+        )
+        out_lines.append("**Model Routing:** (see `route.json`)")
+        return _write_fleet_summary(out_lines, args)
+
+    if spec.get("fast_path") is True:
+        out_lines.extend(_render_fast_path_fleet(spec, route))
```
shafty023, line 11745: Fast-path rendering returns before the common Fleet/Notes block, so it never consumes `spawn_verification.json`. A run with one intended fast-path reviewer but no `agent_fast.json` would still show a normal-looking fast-path summary with no `1 intended | 0 ran` signal. Please render the Fleet runtime tally for fast-path specs too, and update the test that currently asserts `Fleet:` is absent.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+
+    # Model Routing — derived from route.json; mirror the existing
+    # presenter shape. We don't re-derive route ourselves here, just
+    # echo the size category + a compact model summary so the
+    # presenter doesn't need a second read.
+    size_category = str(route.get("size_category", "")).strip()
+    models = route.get("models") or {}
+    if isinstance(models, dict) and size_category:
+        bha_default = ""
+        bha_models = models.get("bug_hunter_a")
```
shafty023: The model summary is being re-derived from `route.json` defaults instead of the actual `spawn_spec.agents[].model` values used for dispatch. This can report `BHA=opus` even when a test-only BHA partition actually ran on Sonnet, drops BHA entirely when `models.bug_hunter_a` is the supported string form, and omits domain critic model assignments. Please build Model Routing from the spawned descriptors, using route only as a fallback.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+                summary_parts.append(f"{label}={value}")
+        summary = ", ".join(summary_parts) if summary_parts else "see route.json"
+        out_lines.append(f"**Model Routing:** {size_category} — {summary}")
+
+    # Fleet outcome line — surface the intended vs ran counts so the
+    # operator sees runtime tally without scrolling to Verifier Stats.
+    stats = spec.get("stats") or {}
+    intended = int(stats.get("agent_count", 0)) if isinstance(stats, dict) else 0
+    if isinstance(verification, dict) and verification.get("verified") is True:
+        present = int(verification.get("present_count", 0))
```
shafty023: `present_count` is not safe to expose as the fleet `ran` count unless it is scoped to the intended spawn spec. The producer currently globs every `agent_*.json` in `CR_DIR`, including earlier system/cached artifacts such as injection detector or coverage-verifier outputs, so the summary can report more agents ran than were intended. Please count only IDs present in `spawn_spec.agents[].agent_id` before rendering or writing the verification tally.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+            value = models.get(slot)
+            if isinstance(value, str) and value:
+                summary_parts.append(f"{label}={value}")
+        summary = ", ".join(summary_parts) if summary_parts else "see route.json"
+        out_lines.append(f"**Model Routing:** {size_category} — {summary}")
+
+    # Fleet outcome line — surface the intended vs ran counts so the
+    # operator sees runtime tally without scrolling to Verifier Stats.
+    stats = spec.get("stats") or {}
+    intended = int(stats.get("agent_count", 0)) if isinstance(stats, dict) else 0
```
shafty023: The fallback behavior handles missing or invalid JSON files, but malformed fields inside otherwise-valid artifacts can still abort presentation. For example, a non-numeric `stats.agent_count` or `spawn_verification.present_count`, or a non-object `route.models` in the fast-path branch, raises instead of degrading to an intended-only or omitted-routing block. Please add safe coercion/shape checks and tests for malformed nested telemetry fields.

plugins/code-review/tools/python/code_review_helpers.py
wongk: do not take action on this comment. the recommended changes are overly defensive.

plugins/code-review/tools/python/code_review_helpers.py
wongk: this sounds overly defensive, but use your best judgement if it's very straightforward to handle.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11329,6 +11329,343 @@ def _emit_no_op(reason: str) -> int:
...
+                f"- ⚠️ **{len(missing_required)} required reviewer(s) did not "
+                f"produce output:** {missing_names}. "
+                "See `coverage_gaps.json`.",
+            )
+
+    # BHA budget cap — collapse multiple partition_id entries into one note.
+    capped_entries = [
+        s for s in skipped
+        if isinstance(s, dict) and s.get("reason") == "budget_capped"
+    ]
```
thadeusb, line 11643: This undercounts in the docs-only case. When the post-arbitrate cap is 0, `_derive_spawn_agents_from_plan` emits a single `budget_capped` skipped entry with `partition_count=N` and no `partition_id`, meaning all N BHA partitions were suppressed. Here `len(capped_entries)` is 1, so the note reads "1 partition(s) exceeded ... (0/N)" when really N got dropped. The cap>0 path is fine because it emits one entry per dropped partition. When cap is 0, drive the count off `partition_count` instead of `len(capped_entries)`.

thadeusb (approved):
Approach is right, a Python renderer over the typed spawn-spec is exactly what Phase 9 should be. One thing on top of the open comments: the BHA cap note miscounts the docs-only case. Otherwise fixable as targeted edits.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Acknowledged — deferring per your guidance. The user evaluated the markdown-injection scenario for rule-resolved reviewer names from `critic-gates.json` and flagged the recommended hardening (name validation + adversarial escaping) as overly defensive for the threat model. The renderer continues to write `reviewer` strings verbatim into the summary block. If we ever surface this on a multi-tenant or untrusted-actor surface this is worth revisiting.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `282ddb5` (v2.23.2). The fast-path branch now falls through to the shared Fleet + Notes block instead of returning early:

```python
fast_path = spec.get("fast_path") is True
if fast_path:
    out_lines.extend(_render_fast_path_fleet(spec, route))
else:
    out_lines.extend(_render_fleet_breakdown(spec))
# ... shared Fleet + Notes rendering below
```

`_render_fast_path_fleet` still owns the Reviewers + Model Routing lines (the bucket/role logic doesn't apply when one agent runs every pass), but the runtime telemetry now consumes `spawn_verification.json` the same way the standard flow does. New `TestPLN725Phase9FastPathFleetTally::test_fast_path_runtime_crash_surfaces_in_fleet_line` pins it: a `fast_path: true` spec + verification showing `present_count: 0` + `missing_agents: [{agent_id: "fast", ...}]` produces `**Fleet:** 1 intended | 0 ran` in the output. Updated the original `test_fast_path_branch` (which asserted `"Fleet:" not in out` and was pinning the buggy behavior) to assert the new line present instead.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `282ddb5` (v2.23.2). New `_render_model_summary` walks `spawn_spec.agents[].model` to build the routing summary; route.json is consulted only as a fallback when the spec has no descriptor for a slot. All three failure modes you flagged are now handled:

- **Test-only BHA on Sonnet:** mixed model assignments render as `BHA=opus/sonnet` (deduplicated, first-appearance order preserved) instead of just the route default. `test_mixed_bha_models_show_both` pins this.
- **Plain-string `route.models.bug_hunter_a`:** when the spec lacks a BHA descriptor, the fallback now handles both the dict shape (`{default: ..., test_only: ...}`) and the plain-string shape. `test_plain_string_route_models_no_longer_dropped` covers it (drops BHA from the spec to force the fallback, sets `models.bug_hunter_a: "sonnet"` as a string, asserts `BHA=sonnet` appears).
- **Domain critic models omitted:** critics (`source in {"rule", "critic"}`) now contribute to a `Critics=<model>` aggregate. `test_critic_model_appears_in_routing` covers it.

Plus a regression test (`test_per_descriptor_model_overrides_route_default`) that pins spec-as-source-of-truth: when the spec's BHB descriptor carries `model: "haiku"` but route still says `sonnet`, the summary echoes `BHB=haiku` — what actually dispatched.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in `282ddb5` (v2.23.2). `cmd_verify_spawn` now scopes `present_ids` to the intersection of `spawn_spec.agents[].agent_id` and the on-disk `agent_*.json` glob:

```python
intended_ids: set[str] = {
    str(d.get("agent_id", "") or "")
    for d in agents
    if isinstance(d, dict) and d.get("agent_id")
}

present_ids: set[str] = set()
for path in cr_dir.glob("agent_*.json"):
    name = path.stem
    if not name.startswith("agent_"):
        continue
    candidate = name[len("agent_"):]
    if candidate in intended_ids:
        present_ids.add(candidate)
```

Files outside the spawn-spec contract (`agent_coverage-verify-blocking.json` from stage_15c, `agent_cached_bha.json` from the cache replay, the upcoming `agent_injection.json` from the injection detector, etc.) no longer inflate `present_count`. Missing-required detection was already correct (it walked the spec list looking for absent files); only the present-side counters were off.

`TestPLN725Phase9VerifySpawnIntendedScoping::test_non_spec_agent_files_excluded_from_present_count` writes two spec-intended outputs + three non-spec artifacts on disk and asserts `present_count == 2` (not 5) and `present_agents == {bhb, auditor}`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed the two trivial cases per your "use your best judgement if very straightforward" guidance:

- Added a tiny `_safe_int` helper that wraps `int()` in a `try/except (TypeError, ValueError)` returning a default. Used at the two places the renderer reads numeric telemetry: `stats.agent_count` and `verification.present_count`. A malformed value (string, `None`, list) now degrades to `0` instead of raising mid-output and suppressing the entire Fleet line.
- Added an `isinstance(dict)` guard on the fast-path `route.models` lookup so a wrong-shape route (e.g. `models: ["malformed"]` from a half-written file) falls back to the spec's per-agent model instead of raising `AttributeError`. The standard flow already had this guard via `_spawn_resolve_models`; fast-path bypassed it.

I did NOT add general "validate every shape recursively" coverage — that's where the line on "overly defensive" sits for me too. The two added guards are scoped to fields the renderer actually reads during normal flow.

`TestPLN725Phase9MalformedTelemetry` covers the three cases: non-numeric `agent_count` degrades to `0 intended`, non-numeric `present_count` degrades to `0 ran`, fast-path survives a non-dict `route.models` (commit `282ddb5`).

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and fixed in `6234368` (v2.23.3). You're right — `_derive_spawn_agents_from_plan` has two `budget_capped` emission shapes and the renderer was conflating them:

- `cap > 0`: one entry per dropped partition, each carrying `partition_id`. `len(capped_entries)` is the correct dropped count.
- `cap == 0` (docs-only): one aggregate entry covering all N suppressed partitions, no `partition_id`, `partition_count` = N. `len(capped_entries) == 1` undercounts.

The renderer now detects the aggregate shape via the absence of `partition_id` on the first entry and drives the dropped count off `partition_count` in that branch:

```python
if "partition_id" not in first:
    dropped = _safe_int(total, len(capped_entries))
else:
    dropped = len(capped_entries)
```

New `test_budget_capped_docs_only_aggregate_uses_partition_count` seeds a single `{budget_cap: 0, partition_count: 5}` entry and asserts the note renders `5 partition(s) ... (0/5)`. The existing `test_budget_capped_partitions_emits_warning` continues to exercise the `cap > 0` branch via per-partition entries with `partition_id`, so both shapes are pinned.

Also fixed the typecheck failure in the same commit (CI's pyright was stricter than my local — the inline `spec.get("stats") if isinstance(spec.get("stats"), dict) else {}` calls `spec.get` twice in one expression and the narrowing doesn't carry across separate calls; binding via a local first fixes it).

PR #137, opened 2026-06-03
refactor(code-review): CRS Phase A — declarative stages + CLI config (v2.24.0)

wongk:
## Summary

CRS Phase A: replace the two largest config-as-code functions in `code_review_helpers.py` with thin loaders over declarative JSON files.

- `_build_run_plan_stages` (894 LOC) → ~30 LOC loader + `tools/python/config/stages.json` (37 stages, 842 LOC)
- `_register_subparsers` (706 LOC) → ~30 LOC loader + `tools/python/config/cli.json` (44 parsers, 199 args, 1,690 LOC)

Net change in `code_review_helpers.py`: **-1,431 LOC** (1,579 removed, 148 added). The file goes from 13,152 LOC to 11,721 LOC. The two largest functions in the helpers file are eliminated.

## How runtime behavior is preserved

The stage loader substitutes template variables at call time:
- `{cr_dir}`, `{mode}` — function arguments
- `{schema_version}` — `str(SCHEMA_VERSION)` from `code_review_schema`
- `{flags_scope_args}`, `{flags_base_ref_override}`, `{flags_full_review}`, `{flags_since_last_review}` — derived from the `flags` dict
- `@pr_flag` — splat marker, inserts `["--pr-number", str(pr_number)]` when `pr_number` is truthy, `[]` otherwise (matches the original conditional `*pr_flag` splat exactly)

Angle-bracket walker tokens (`<CACHE_DIR>`, `<DIFF_TIP>`, `<PROMPT_HASH>`, etc.) pass through as literal strings — they remain the orchestrator/walker's responsibility to resolve from prior-stage outputs.

The CLI loader resolves `$$NAME` references in `cli.json` via `_resolve_cli_constant` at parser-build time:
- `$$DEFAULT_MAX_BHA_AGENTS`, `$$BUDGET_TOTAL_CAP_DEFAULT`, `$$CACHE_GC_TTL_DAYS_DEFAULT`, `$$CACHE_GC_MAX_PER_FILE_DEFAULT`, `$$_EXTRACT_PATCHES_BATCH_SIZE`, `$$SIGNAL_EXTRACTION_MODEL_DEFAULT`, `$$COVERAGE_CRITIC_MODEL_DEFAULT` → imported integer/string constants
- `$$COVERAGE_SCOPES_SORTED` → `sorted(COVERAGE_SCOPES)` from `code_review_schema`

Constant edits at the import site propagate to argparse defaults without touching `cli.json`. The 7 constant slots and 1 computed-choice slot are hand-curated against the source to avoid value-collision false positives (e.g. `--max-files default=20` collides with `BUDGET_TOTAL_CAP_DEFAULT=20` by value but is a literal in the source).

The single mutually-exclusive group (`migrate-critic-gates --output` vs `--in-place`) is declared via `mutex_groups[].actions` on the parser entry; the loader creates the group and routes the listed flags into it.

## Verification

- `TestCRSPhaseADeclarativeStagesConfig` (5 tests) asserts the loader-based `_build_run_plan_stages` produces output equal to two pre-refactor snapshots:
  - `local_no_pr_empty_flags.json` — every conditional in its "off" state
  - `github_pr42_all_flags.json` — every conditional in its "on" state
- Also pins `@pr_flag` semantics (omitted/inserted at correct positions in stages 03/04/08/25) and that `{schema_version}` resolves to `SCHEMA_VERSION` rather than a literal.
- The existing argparse-introspection regression tests (`test_enabled_helper_stages_include_all_required_argparse_args`, `test_enabled_helper_stages_parse_via_argparse_after_token_substitution`, `test_runtime_placeholders_present_for_orchestrator`) still pass — they're the structural safety net.

All checks green:
- 1,008 code-review tests pass (5 skipped)
- 1,747 monorepo tests pass (5 skipped)
- `uv run ruff check plugins/code-review/` clean
- `uv run pyright plugins/code-review/` 0 errors / 0 warnings / 0 informations

## Test plan

- [ ] Confirm the JSON-vs-TOML decision (plan recommended JSON, kept JSON, design rationale comments now belong in the source review for individual entries rather than inline since JSON doesn't carry them)
- [ ] Confirm `tools/python/config/` is the right home (vs `plugins/code-review/config/` or `tools/config/`)
- [ ] Spot-check `cli.json` against the source for any missed constant references — current curated slots cover the 7+1 references identified during the catalogue pass

## Next: CRS Phase B

Phase B (singleton-dispatch unification) is the next phase. It depends on Phase A only in that it edits stage definitions in `stages.json` to collapse the 4 `prepare`/`consolidate` pairs.


plugins/code-review/tools/python/test_code_review_helpers.py
```diff
@@ -17491,3 +17491,79 @@ def test_needs_agent_with_missing_output_still_fails_closed(
         assert (cr_dir / "agent_coverage-critic-failed.json").exists()
         canonical = json.loads((cr_dir / "coverage_plan.json").read_text())
         assert canonical["critic_status"] == "fail_closed"
+
+
+class TestCRSPhaseADeclarativeStagesConfig:
```
thadeusb, line 17496: The stages config got a real byte-equality snapshot guard here, but cli.json didn't get an equivalent and it's the bigger, more fragile of the two (199 args, 8 hand-curated $$ constant slots). The existing _register_subparsers tests only assert that stage args parse through the built parser, not what the defaults/choices/types actually resolve to. So if someone bumps --max-files default in cli.json or fat-fingers a $$ slot, nothing fails. A snapshot of the resolved parser spec (the same dump you'd get introspecting sub._actions) would close that gap and match what you did for stages.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9221,898 +9222,79 @@ def cmd_verdict(args: argparse.Namespace) -> int:
...
-            "enabled": True,
-        },
-    ]
+    def resolve_args(template_args: list[str]) -> list[str]:
+        out: list[str] = []
+        for arg in template_args:
+            if arg == "@pr_flag":
+                out.extend(pr_flag)
+            else:
+                out.append(arg.format(**ctx))
```
thadeusb, line 9349: Heads up for whoever edits stages.json next: a literal { or } in an arg/stdout/expected_outputs string will blow up here with a KeyError/ValueError from .format(), since these all run through it. Runtime flag values are safe (format doesn't re-scan them), it's only the template strings. The old f-string code had the same trap but at least it was visibly Python. Non-obvious now that it's hand-authored JSON.

thadeusb (approved):
Pulled both versions and diffed them: 44/44 parsers identical down to defaults and choices, 37 stages identical across every mode/pr/flags combo, mutex membership matches. The byte-for-byte claim holds. Two forward-looking notes on test coverage, nothing blocking.

wongk:
v2.24.1 pushed (commit \`99f0612\`) — addresses both review comments plus 5 verified findings from the local cr-97797 review.

**Re: snapshot guard for cli.json (this comment + cr-97797 finding domain_0_f1)**
Added \`tools/python/fixtures/run_plan_snapshots/cli_parser_resolved.json\` — the same shape as the stages snapshot but for the resolved argparse spec across all 44 subparsers (name, dest, default, type, choices, required, action, func, mutex groups). \`test_resolved_parser_spec_matches_snapshot\` diffs the live loader against it, so a default/type/choices/func edit or a misrouted mutex member now fails the suite. \`TestCRSPhaseACLIConfigLoader\` also covers the constant resolution explicitly (one assertion per \`$$\` slot), the mutex success/failure paths, and the loader's error paths.

**Re: literal \`{\`/\`}\` trap in stages.json templates**
Added \`_validate_stages_config\` inside the lru_cached loader. It scans every arg / expected_output / stdout string at load time and rejects unknown \`{key}\` patterns or unbalanced braces with a clear \`stages.json stage_id.field[index]: ...\` diagnostic. The known key set lives in \`_STAGES_TEMPLATE_KEYS\` (a frozenset of the seven substitutable variables). \`@pr_flag\` is recognized as a splat marker, not a template key. \`TestCRSPhaseAStageTemplateValidator\` covers the four cases: unknown key, unbalanced braces, valid input, and diagnostic naming.

**Additional fixes from cr-97797:**
- \`copy.deepcopy\` stage templates so lru_cached list refs aren't shared with returned stages (finding domain_0_f0 / domain_1_f0 / domain_3_f0; pinned by \`test_returned_stage_lists_are_not_aliased_to_lru_cache\`)
- \`_cli_cmd_registry()\` static allowlist replaces direct \`globals()[func_name]\` lookup (domain_2_f0); descriptive \`ValueError\` instead of opaque \`KeyError\` past \`main()\`'s except (domain_4_f0)
- Reject \`func\` key in \`extra_defaults\` to prevent handler override (additional concern in domain_2_f0)
- \`_resolve_cli_constant\` error message now uses \`\$\$NAME\` to match the cli.json encoding convention (bhb_f0)

1,022 code-review tests pass; \`uv run ruff\` and \`uv run pyright\` clean.

PR #138, opened 2026-06-03
refactor(code-review): CRS Phase B — shared helpers for singleton prepare/consolidate (v2.25.0)

wongk:
## Summary

CRS Phase B: extract shared helpers that de-duplicate the boilerplate replicated across the PLN-725 Phase 1 (`extract-signals`) and Phase 3 (`coverage-critic`) prepare/consolidate pairs.

**Scope reduction from the plan.** The original plan claimed Phase B would collapse 4 prepare/consolidate pairs into a unified `SingletonDispatchSpec` shape, saving ~600 source + ~1,200 test LOC. Reading the code revealed the 4 pairs split into 2 patterns, not 1:

- **Pattern A — singleton + cache**: `extract-signals`, `coverage-critic`. Manifest has `{cache_hit | needs_agent | skipped}` status semantics.
- **Pattern B — fleet dispatch**: `verify-findings`, `review-dismissed`. Per-finding inputs to `to_verify[]`, no cache-hit shortcut.

Pattern A and Pattern B don't unify into one shape — forcing them would have created a fat callback interface heavier than the duplication it removed. This PR ships Pattern A only and leaves Pattern B unchanged, following the plan's documented escape hatch (\"if forcing it costs more than it saves, leave some out\"). Net actual delta: helpers.py +33 LOC (5 new helpers + better-structured fail-closed builders, partially offset by cache-helper collapse). Tests +183 LOC (17 new unit tests).

## Five new shared helpers

| Helper | Replaces |
|---|---|
| `_read_manifest_dict(path)` | The 7-line `if path.exists(): try: load + isinstance check + except: {}` pattern that both consolidates duplicated |
| `_emit_summary(payload)` | The `json.dump(payload, sys.stdout, indent=2); sys.stdout.write(\"\\n\"); return 0` triple at every prepare/consolidate exit |
| `_read_agent_output_or_error(path)` | The 6-line agent-output read pattern that both consolidates used before routing into fail-closed |
| `_write_and_emit_manifest(path, manifest)` | The manifest-write + stdout-dump combo at every prepare exit |
| `_read_simple_cache_entry` + `_write_simple_cache_entry` | The 40+15 LOC `written_at`-stamped cache lifecycle that `_read_cached_signals` / `_write_cached_signals` / `_read_cached_coverage_critic` / `_write_cached_coverage_critic` previously implemented twice |

## Call-site cleanup

- `cmd_extract_signals_consolidate`: read-error and validation-rejected fail-closed branches now both route through one `_fail_closed` closure instead of duplicating the canonical-builder + finding-emit + write + dump pattern.
- `cmd_coverage_critic_consolidate`: same — existing `_fail_closed` closure remains; manifest read and stdout dumps now use the shared helpers.
- `cmd_extract_signals_prepare` + `cmd_coverage_critic_prepare` + `_emit_skipped_coverage_plan`: manifest writes route through `_write_and_emit_manifest`.
- `_read_cached_signals` / `_write_cached_signals` / `_read_cached_coverage_critic` / `_write_cached_coverage_critic`: each becomes ~10 LOC of thin-wrapper over the shared cache helpers.

## Out of scope

- `_read_cached_verification` stays on its own implementation. The verifications cache uses a different timestamp field (`cached_at` vs `written_at`), a payload envelope wrapping `verdict_data`, and atomic tmp+rename writes — unifying it would have widened the shared interface past the savings.
- `cmd_verify_prepare` / `cmd_verify_consolidate` / `cmd_review_dismissed_prepare` / `cmd_review_dismissed_consolidate` (Pattern B) are untouched.

## Verification

- **1,039 tests pass** (1,022 pre-existing + 17 new `TestCRSPhaseBSharedHelpers` covering each helper's contract — missing/malformed/empty/valid manifest, missing path/missing `written_at`/expired-TTL/fresh cache, fail-open OSError behavior, etc.)
- `uv run ruff check plugins/code-review/` — clean
- `uv run pyright plugins/code-review/` — 0 errors / 0 warnings / 0 informations
- All golden fixtures unchanged

## Test plan

- [ ] Confirm the scope reduction is acceptable (Pattern A only, Pattern B deferred)
- [ ] Spot-check the helper signatures for any contract that should be tightened further before merge

## Next: Phase C

Phase C (coverage artifact consolidation) is independent of B and ready to start after merge.


wongk:
v2.25.1 pushed (commit \`d1adc29\`) — fixes the stage_14 stdout-redirect race + 3 latent siblings in the same bug class.

**The reported bug.** \`cmd_resolve_coverage\` writes \`<cr_dir>/coverage_plan_initial.json\` directly via \`--cr-dir\` AND prints a small summary JSON to stdout. The pre-v2.25.1 stage_14 entry redirected stdout to the same file path, so the shell's \`>\` opened the file at offset 0, the helper's \`open(...,\"w\")\` truncated and wrote the full plan, then the helper's stdout write overwrote the head of the file with the 8-line summary. The result: summary at the top, trailing garbage from the longer plan, \`Extra data: line 9 column 6\` on the next stage's parse. Predates CRS (the bug was in the original \`_build_run_plan_stages\` Python literal and got carried into \`config/stages.json\` in v2.24.0) but lives squarely in the file this PR refactors, so it's natural to fold in.

**Three latent siblings I found via a systematic audit.** I scanned every stage with non-null stdout against its cmd's source for the \"writes own file via cr_dir + prints to stdout\" shape:

| Stage | Cmd | Verdict |
|---|---|---|
| stage_11_extract_signals | extract-signals-prepare | RACE-RISK (works by coincidence) |
| stage_14_resolve_coverage | resolve-coverage | **REPORTED BUG** |
| stage_15_coverage_critic | coverage-critic-prepare | RACE-RISK (works by coincidence) |
| stage_22b_verify_prepare | verify-prepare | RACE-RISK (works by coincidence) |

The three siblings use \`_write_and_emit_manifest\` (introduced in v2.25.0 by this same PR), which writes the canonical manifest file to disk AND prints the same manifest to stdout. With the stdout redirect active, both writes produced the same bytes modulo trailing newline — so the file ended up correct **by coincidence, not by design**. A future edit that makes one write path differ from the other (extra field, different indent, conditional content) would have re-introduced the race silently. Setting \`stdout: null\` removes the redundant redirect across all four; the helper is now the sole writer for the canonical file.

**Regression test.** \`TestRunPlanStdoutRedirectRaceClass\` pins the invariant: stages in \`CMD_WRITES_OWN_FILE_STAGES\` (the 4 above) must have \`stdout: null\`. A second test pins \`stage_08_fetch_intent\` (the canonical example documented in \`start.md\`) for consistency. A future stage in the same class must be added to the list and given \`stdout: null\` simultaneously.

**Verification:**
- 1,041 code-review tests pass (1,039 + 2 new race-class tests)
- Snapshots \`local_no_pr_empty_flags.json\` and \`github_pr42_all_flags.json\` regenerated to match the 4 new \`stdout: null\` fields
- \`uv run ruff check\` clean, \`uv run pyright\` 0 errors/warnings/informations

wongk:
v2.25.2 pushed (commit \`522f296\`) — both MEDIUM findings resolved.

**Finding 1 (test masks the cache_dir guard):** \`test_read_simple_cache_entry_returns_none_when_cache_dir_is_none\` now seeds a valid fresh cache entry on disk, asserts \`path.exists()\` before the call (sanity-checking we're not testing the path-missing branch), and only then asserts the read returns \`None\`. A regression that dropped the \`cache_dir is None\` check would now fail this test — the file-missing branch can't mask it because the file exists with a fresh \`written_at\` stamp. Matches the sibling write-side test's side-effect-assertion pattern.

**Finding 2 (fail-open contract dropped from wrapper docstrings):** \`_write_cached_signals\` and \`_write_cached_coverage_critic\` docstrings now explicitly restate the fail-open contract, naming both the delegation to \`_write_simple_cache_entry\` and the canonical \`<cr_dir>/extract_signals.json\` / \`<cr_dir>/coverage_plan.json\` that makes cache write failure a re-run cost issue rather than a pipeline halt. Callers reading the wrapper docstring no longer have to chase the contract one indirection deeper.

1,041 tests pass; \`uv run ruff check\` clean, \`uv run pyright\` 0 errors/warnings/informations.

PR #139, opened 2026-06-03
refactor(code-review): CRS Phase D — consolidate spawn artifacts (v2.26.0)

wongk:
## Summary

CRS Phase D: consolidate the three spawn-decision artifacts (`route.json`, `spawn_spec.json`, `spawn_verification.json`) into one `spawn.json` with three sections.

**Phase ordering note:** plan said C and D could land in either order. After surveying both, Phase C (coverage consolidation) was substantially larger (7 cmd functions, 30+ test refactors, orchestrator-prompt protocol changes for `pln725_coverage_critic.json`) than D (4 cmd functions, ~30 test seed updates, one Gate B prose update). D first establishes the consolidation pattern at lower risk; C is queued next.

## The consolidation

| Before | After |
|---|---|
| `<CR_DIR>/route.json` (Gate B) | `<CR_DIR>/spawn.json` → `.route` |
| `<CR_DIR>/spawn_spec.json` (stage_19b) | `<CR_DIR>/spawn.json` → `.spec` |
| `<CR_DIR>/spawn_verification.json` (stage_20b) | `<CR_DIR>/spawn.json` → `.verification` |

Three wire-format artifacts → one. The presenter (`cmd_render_fleet_summary`) now opens one file with three section lookups instead of three separate file reads.

## Helpers

- **`_read_spawn_state(cr_dir)`** — tolerant read of `<cr_dir>/spawn.json`, returns `{}` on missing/malformed/non-dict so consumers chain `state.get("route", {})` etc.
- **`_write_spawn_section(cr_dir, section, payload)`** — atomic read-modify-write via tmp + `os.replace`. A crash mid-write leaves the prior state intact rather than a half-written file. `_SPAWN_STATE_SECTIONS` is a frozenset closed-vocabulary check; passing an unknown section raises `ValueError` at the call site.

## Race-class lesson applied

v2.25.1 fixed `stage_14_resolve_coverage` where the cmd wrote a file via `--cr-dir` AND the orchestrator's stdout redirect raced the helper's own write, corrupting the file. Phase D applies the same lesson preemptively: `cmd_route` gains `--cr-dir` and writes `spawn.json.route` directly via atomic section update. The stdout summary still fires for operator visibility, but the canonical write target is no longer the orchestrator's redirected stdout. Gate B's prose updates from `> <CR_DIR>/route.json` to `--cr-dir <CR_DIR>`, eliminating the race surface entirely.

## Cmd refactors

[...]

PR #140, opened 2026-06-04
refactor(code-review): CRS Phase C — consolidate coverage artifacts (v2.27.0)

wongk:
## Summary

Final phase of the CRS (Code Review Simplification) refactor sequence: consolidate four standalone coverage-decision artifacts into one `coverage.json` with sections `initial` / `critic` / `final` / `verify`. Reuses the Phase D `_read_spawn_state` / `_write_spawn_section` atomic section-write pattern; mirrors that PR's shape and scope discipline.

### Before → After

| Before | After |
|---|---|
| `<CR_DIR>/coverage_plan_initial.json` | `<CR_DIR>/coverage.json` → `.initial` |
| `<CR_DIR>/coverage_critic_manifest.json` | `<CR_DIR>/coverage.json` → `.critic` |
| `<CR_DIR>/coverage_plan.json` | `<CR_DIR>/coverage.json` → `.final` |
| `<CR_DIR>/coverage_verify.json` | `<CR_DIR>/coverage.json` → `.verify` |

`coverage_gaps.json` stays standalone (multi-writer append doesn't fit atomic sections). `coverage_critic_input.json` + `coverage_critic_diff_summary.json` stay standalone (large LLM-prompt material, single-consumer via Task tool).

### Scope discipline

The original CRS plan listed Phase C as "consolidate 6 → 2"; surveying the actual landscape narrowed scope to **4 → 1**. The LLM-input artifacts and gaps file genuinely don't belong in a tightly-managed decision aggregate — they have different access patterns (large/single-reader vs. multi-writer append) and tens of KB of irrelevant blob would burden every section-read consumer.

### Why the pattern works here

1. Four distinct stages, each writing its own section. No two writers race for the same section.
2. Atomic read-modify-write via `tmp + os.replace` so a crash mid-write leaves the prior state intact.
3. `_COVERAGE_STATE_SECTIONS` frozenset catches typo'd section names at the call site (`ValueError`).
4. `stage_16_arbitrate_budget` continues to mutate `.final` in place — same shape as the pre-Phase-C in-place overwrite of `coverage_plan.json`, atomicity now guaranteed.

### Backwards-compatibility

Legacy CLI args (`--coverage-plan-initial`, `--coverage-plan`, `--manifest`, `--coverage-verify`, `--output`) remain accepted as explicit-path overrides for callers (tests, ad-hoc CLI use) that haven't switched. Production stages drop them and use `--cr-dir` + section defaults.

### Audit-rename sweep

Per the conversation's standing concern about Phase D's trailing docstring drift: after the code rewrite I ran a mechanical grep for every legacy filename across the plugin (`code_review_helpers.py`, `cli.json`, `stages.json`, `start.md`, `SCHEMA.md`, test docstrings) and audited each hit's ±20 lines of surrounding context. Every survivor is now either:
- Correctly framed as historical context (`Phase C / pre-v2.27.0` / `legacy`); or
- A real test fixture writing the legacy file as input to the legacy `--coverage-plan-initial` arg (still accepted).

Pinned the same `--cr-dir` test coverage that Phase D's PR #139 follow-up flagged was missing — Phase C's production paths are covered by the existing prepare/consolidate/verify/arbitrate tests, which all run through the cmds end-to-end.

## Test plan

- [x] `uv run ruff check plugins/code-review/` — clean
- [x] `uv run pyright plugins/code-review/` — 0 errors / 0 warnings / 0 informations
- [x] `uv run pytest plugins/code-review/tools/python/` — 1044 passed, 5 skipped (unchanged from v2.26.1)
- [x] Run `cmd_route` / `cmd_resolve_coverage` / `cmd_coverage_critic_prepare` / `cmd_coverage_critic_consolidate` / `cmd_verify_coverage` / `cmd_arbitrate_budget` end-to-end against the existing test corpus
- [x] Three snapshot fixtures regenerated (cli_parser_resolved + local + github)
- [x] Audit-rename sweep for stale docstrings/comments

## CRS sequence status

| Phase | Scope | Status |
|---|---|---|
| A | Declarative `stages.json` + `cli.json` configs | shipped (v2.24.0) |
| B | Shared helpers for singleton prepare/consolidate | shipped (v2.25.0) |
| D | Spawn artifacts → `spawn.json` sections | shipped (v2.26.0 + v2.26.1 follow-up) |
| C | Coverage artifacts → `coverage.json` sections | this PR (v2.27.0) |

CRS is now complete. The wire-format-artifact count across `<CR_DIR>` drops from the pre-CRS-Phase-A baseline by 4 (Phase D) + 3 (Phase C, net of the LLM-input + gaps files kept standalone) = 7 artifacts consolidated into 2 aggregates (`spawn.json` + `coverage.json`).


thadeusb (approved):
Straight port of the Phase D section-write pattern onto the coverage artifacts, and the piece that actually keeps the rewire safe, dropping required:true off the legacy CLI args so the --cr-dir-only stages don't trip argparse, is done right. Traced arbitrate/derive/finalize to the right sections and ran the suite at HEAD, 954 pass. Nothing to add.

plugins/code-review/tools/python/config/stages.json
```diff
@@ -327,7 +327,7 @@
       ],
       "stdout": null,
       "expected_outputs": [
-        "{cr_dir}/coverage_plan_initial.json"
+        "{cr_dir}/coverage.json"
```
shafty023, line 330: `stage_14_resolve_coverage` is now the producer for `coverage.json.initial`, but the stage args above still omit `--critic-gates`. `cmd_resolve_coverage` falls back to an empty manifest when that arg is absent, so production runs ignore configured `coverage[]` and legacy `moduleCritics[]` rules and only route the core floor. Please wire `.closedloop-ai/settings/critic-gates.json` or the documented tokenized equivalent into this stage and add a run-plan regression for the production args.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9420,7 +9578,10 @@ def _build_validation_gates(cr_dir: str) -> list[dict[str, Any]]:
         {
             "after_stage": "stage_16_arbitrate_budget",
             "gate": "coverage_plan_well_formed",
-            "outputs": [f"{cr_dir}/coverage_plan.json"],
+            "outputs": [f"{cr_dir}/coverage.json"],
+            "required_sections": {
```
shafty023, line 9582: This adds `required_sections`, but the walker contract in `start.md` still says gates only check `outputs` existence/well-formedness. Because `coverage.json` exists from stage 14, the stage-16 gate can still pass with no `.final` section. Please update the runtime gate contract or deterministic helper to load each `required_sections` file and require the listed top-level sections, with a regression where the file exists but `.final` is missing.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7669,72 +7703,102 @@ def cmd_coverage_critic_prepare(args: argparse.Namespace) -> int:
...
-
     # PLN-725 Phase 4: orchestrator wiring runs consolidate unconditionally as
     # the stage after the agent-dispatch step. On ``cache_hit`` or ``skipped``,
-    # prepare already wrote coverage_plan.json (cache hit serves the cached
+    # prepare already wrote coverage.json.final (cache hit serves the cached
     # plan; ``--no-critic`` stamps the initial plan + critic_status="skipped").
     # Either way no agent was spawned and the work is done — no-op so the
     # walker stays mechanically-driven without conditional dispatch.
-    if manifest_status in ("cache_hit", "skipped") and output_path.exists():
+    final_already_written = isinstance(
```
shafty023, line 7795: This no-op check only looks for `coverage.json.final`, even when the caller supplied a legacy `--manifest`. A resumed or mixed-version run can have `coverage_critic_manifest.json` with `status: cache_hit` or `skipped` and a standalone `coverage_plan.json`; this path will miss that final plan, try to read a non-existent agent output, and fail closed. Please honor the legacy manifest output path or `<cr-dir>/coverage_plan.json` and seed `coverage.json.final` before returning.

plugins/code-review/tools/python/config/cli.json
```diff
@@ -591,7 +589,7 @@
     },
     {
       "name": "verify-coverage",
-      "help": "Deterministic verifier for coverage_plan.json. Emits coverage_verify.json with verdict PASS|BLOCKING and per-check violations; emits a HIGH system finding on BLOCKING.",
+      "help": "Deterministic verifier for coverage.json.final. Writes verdict PASS|BLOCKING and per-check violations to coverage.json.verify; emits a HIGH system finding on BLOCKING.",
```
shafty023, line 592: The PR body says legacy CLI args remain accepted, but this command drops `--output`. Existing callers that still invoke `verify-coverage --coverage-plan ... --coverage-plan-initial ... --output coverage_verify.json` now fail argparse, and downstream legacy `arbitrate-budget --coverage-verify` can see a missing verifier artifact. Please restore `--output` as an explicit-path fallback and write the standalone verdict when it is supplied.

plugins/code-review/commands/start.md
```diff
@@ -286,16 +286,16 @@ These notes annotate the run-plan stages with anything not obvious from the plan
...
...r drop), prunes lowest-priority best-effort, and computes the final `bha_partitions` count. BLOCKING verdict → arbitration is bypassed entirely; the input plan flows through unchanged with `budget.gated_by_verify: true` and `arbitrate_status: "blocked_by_verify"`. No new finding is emitted for the gate — the canonical BLOCKING finding already lives in `agent_coverage-verify-blocking.json` from stage_15c, and double-counting would inflate the run summary. Missing `coverage.json.verify` (verifier didn't run, upstream aborted) is treated as PASS so the arbitration path remains operable when Phase 6 telemetry is degraded. `on_failure: abort` — arbitrate-budget has been stable since v2.16.x; flipping it on means a real I/O or shape error here halts the pipeline. The BLOCKING short-circuit is exit 0, not a failure. **Note:** as of Phase 8 (v2.22.0), `stage_20_spawn_reviewers` consumes `spawn.json` (`spec` section, derived by `stage_19b_derive_spawn_spec` from this stage's output — Phase D v2.26.0 consolidated the legacy `spawn_spec.json` into a section of `spawn.json`) and falls back to the static reviewer table below only when the spec is missing or marks `arbitrate_status: "fallback"`.
```
shafty023, line 294: This docs line says a BLOCKING verdict lets the input plan flow through and review is only annotated, but `cmd_arbitrate_budget` currently writes `budget.bha_partitions: 0` on that path and `derive-spawn-spec` treats 0 as a hard cap, skipping all BHA partitions. That drops the core reviewer floor for any blocked verification with real partitions. Please preserve or omit the BHA cap on the gated path unless the diff is actually docs-only, and add a test with `bug_hunter_a` plus partitions under a BLOCKING verifier verdict.

PR #141, opened 2026-06-04
refactor(code-review): strip historical-framing bloat + remove dead args (v2.27.2)

wongk:
## Summary

Two interleaved cleanups identified by parallel audits of the code-review plugin:

- **Historical-framing prose strip** — removed phase labels (`Phase A/B/C/D`), version-tagged "as of" markers, and `pre-vN.M.K` references across `start.md`, `code_review_helpers.py`, `code_review_schema.py`, `SCHEMA.md`, and `README.md`. Stage descriptions, docstrings, and inline comments now describe the code as it currently stands. Legitimate backward-compatibility shims are framed as such instead of as "legacy". Net `+495/-826` lines.
- **Dead-arg removal** — `--no-verify`, `--no-verify-reason`, and `--no-patch-lines` were defined in `cli.json` but never plumbed through `stages.json`. No orchestration path could invoke them. Removed the args, manifest fields, audit-banner sections in `github-review.md` / `present-local/SKILL.md`, the operator-flag README table entry, and the `pending_verification.md` template bullet.

## Test plan
- [x] `uv run ruff check .` — All checks passed!
- [x] `uv run pyright` — 0 errors / 0 warnings / 0 informations
- [x] `uv run pytest plugins/` — 1061 passed, 5 skipped
- [x] Regenerated `cli_parser_resolved.json` snapshot to reflect removed flags
- [x] Deleted `TestNoVerifyBypass` (3 tests) and dropped `no_verify` / `no_verify_reason` kwargs from `_run_verify_prepare` helper


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -505,7 +482,7 @@ def cmd_parse_diff(args: argparse.Namespace) -> int:
     total_loc = sum(v["added"] + v["removed"] for v in file_loc.values())
```
thadeusb, line 482: Removing `--no-patch-lines` here orphans a test name in `test_code_review_helpers.py:231` (not in this diff): `test_no_patch_lines_flag`. Its body is fine — it calls `_parse_u0_output(include_patch_lines=False)` directly and that param survives — but the name now points at a CLI flag that no longer exists. Consider renaming it to something like `test_parse_u0_omit_patch_lines`.

thadeusb (approved):
Chased the dead-flag claim end to end and it holds: --no-verify/--no-verify-reason/--no-patch-lines never had stages.json wiring, so nothing could invoke them, and include_patch_lines=True is identical behavior since the flag was never set. Removal is consistent across both presenters, README, schema, template, and the regenerated snapshot. Only thing left dangling is a test named after a flag that no longer exists.

shafty023 (commented):
Guardrail issue: root `AGENTS.md` says plugin changes should not manually edit `CHANGELOG.md` or `README.md`; after code changes, run the repo's `/update-documentation` workflow, and PR descriptions should include exact commands run. This PR changes both `CHANGELOG.md` and `plugins/code-review/README.md`, but the test plan lists `ruff`, `pyright`, `pytest`, fixture regeneration, and test deletion only. Please run `/update-documentation` for `code-review` or document the exact equivalent if it was already run, then update the PR test plan so docs/changelog provenance is explicit.

plugins/code-review/prompts/github-review.md
```diff
@@ -296,14 +296,6 @@ The Reviewer column keys off the `reviewer` field, which `cmd_collect_findings`
 </details>
 ```
 
-If `<CR_DIR>/verify_manifest.json` exists and has `no_verify: true`, prepend the audit banner BEFORE the collapsible block:
```
shafty023, line 299: Removing the `no_verify: true` banner drops the old-producer/new-presenter compatibility path. A `verify_manifest.json` produced by v2.27.1 can still carry `no_verify: true`; with this presenter, both the verifier-stats comment and summary would omit the audit signal and make a bypassed run look like an ordinary verified run. Please keep a read-only banner path for existing manifests in the GitHub presenter, and mirror it in `skills/present-local/SKILL.md`, without re-adding the CLI flag.

plugins/code-review/README.md
```diff
@@ -54,10 +54,10 @@ canonical schema documented in [SCHEMA.md](SCHEMA.md). Key contracts:
 - **Canonical `prompt_hash`** folds in `schema_version`: a MAJOR schema bump
   invalidates every cache namespace at once.
 
-The legacy `validate_output.json` is still emitted alongside `review_result.json`
-during Phase A so existing consumers keep working. Phase B will delete the
-legacy file in lockstep with the `code` plugin's `run-loop.sh` and the `/fix`
-skill.
+For backward compatibility, `validate_output.json` is still emitted alongside
+`review_result.json` so existing consumers (`code` plugin's `run-loop.sh`,
+the `/fix` skill) keep working. Future versions may discontinue this dual
```
shafty023, line 59: This compatibility rationale still lists `/fix` as a `validate_output.json` consumer, but `plugins/code-review/skills/fix/SKILL.md` says `/fix` reads only `<CR_DIR>/review_result.json` and has no legacy fallback. Keeping `/fix` in this list makes the artifact contract misleading. Please remove `/fix` from the `validate_output.json` consumer list or state that `/fix` consumes `review_result.json`.

plugins/code-review/README.md
```diff
@@ -249,7 +249,6 @@ When the verifier dismisses a finding the operator believes is real, three flags
 |---|---|
 | `--re-assert <id>[,<id>...]` | Write an override file at `<CACHE_DIR>/overrides/<finding_id>.json`. On the next run, `cmd_verify_prepare` honors the override (synthesizes a `RE_ASSERTED` verdict and skips the agent spawn) so long as the cited file's content hash still matches. Optional `--re-assert-reason='<why>'` is recorded in the override and in `pending-learnings/verifier-overrides.jsonl`. |
 | `--review-dismissed` | Run a second opinion via the haiku verifier against the prior run's `rejected[]`. Any verdict that is NOT `REJECTED` auto-promotes via a `REVIEW_DISMISSED` override. Side-by-side diff lands at `<CR_DIR>/review_dismissed_diff.json`. |
-| `--no-verify` + `--no-verify-reason='<why>'` | Emergency bypass — verifier skipped entirely; every eligible finding lands in `verified[]` with `verifier_verdict: null`. Footer carries an explicit audit banner so the bypass is never silent. Mutually exclusive with `--re-assert` / `--review-dismissed`. |
```
shafty023, line 252: After removing this row, the section lists only two override flows. The lead-in still says "three flags," so the operator docs no longer match the CLI surface. Please change this to "two flags" or rephrase it as "the following flags."

plugins/code-review/tools/python/fixtures/run_plan_snapshots/cli_parser_resolved.json
```diff
@@ -1810,4 +1787,4 @@
       ]
     }
   ]
-}
+}
```
shafty023: The regenerated snapshot lost its trailing newline (`\ No newline at end of file`). Please regenerate or write this fixture with a final newline so the generated artifact stays normalized and avoids avoidable fixture churn.

wongk:
Re: AGENTS.md guardrail on manual CHANGELOG/README edits — the workflow was followed.

CHANGELOG entry was produced via `/update-documentation --changelog-only --plugin code-review` (the skill the guardrail points to), not hand-authored. README.md edit was a single-section accuracy fix removing the false `<pr_verdict>` tag claim and the `--no-verify` / `--no-patch-lines` operator-flag rows for flags this PR deletes — the same workflow's Step 5 README verification covers that and explicitly allows minimal accuracy edits to inaccurate sections. PR description updated with the explicit commands.

Follow-up review fixes for cr-14379 + your other-PR callouts just landed in commit 6d1914d as v2.27.3.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Renamed to `test_include_patch_lines_false` in 6d1914d. Slightly different from your suggestion (`test_parse_u0_omit_patch_lines`) — went with `include_patch_lines_false` because it reads directly off the kwarg the test passes, but the spirit (don't reference a dead CLI flag in the test name) is identical.

plugins/code-review/prompts/github-review.md
wongk: Pushing back on this one with evidence. The `--no-verify` / `--no-verify-reason` pair was never wired through `stages.json` — thadeusb chased the dead-flag claim end-to-end in his PR approval: "never had stages.json wiring, so nothing could invoke them." `cmd_verify_prepare` accepted the args but no orchestration stage ever passed them, so `verify_manifest.json` on disk has never carried `no_verify: true` in any released v2.x — not v2.27.1, not earlier. There are no in-the-wild manifests with this field for the presenter to detect.

Keeping a read-only banner path would be cargo-culted defense against a producer that never existed. If the bypass becomes a real feature in the future, the producer side gets wired through stages.json AND the presenter gets the banner at the same time.

plugins/code-review/README.md
wongk: Correct on substance — but the whole paragraph is fiction. `validate_output.json` is never written by any subcommand; the README claim that it's "still emitted alongside review_result.json" is documentation drift, and `/fix` is not actually a consumer. PR #142 deletes the entire paragraph (1c8a8a0 / now 903d601), so the misleading `/fix` entry disappears along with the false dual-emission claim.

plugins/code-review/README.md
wongk: Fixed in 927255a (v2.27.4): "three flags" → "two flags".

plugins/code-review/tools/python/fixtures/run_plan_snapshots/cli_parser_resolved.json
wongk: Fixed in 6d1914d (v2.27.3): trailing newline restored on `cli_parser_resolved.json`. All three snapshot fixtures verified.

PR #142, opened 2026-06-04
refactor(code-review): remove all unproven legacy support surfaces (v2.27.3)

wongk:
## Summary

Evidence-based audit of every backward-compat / legacy surface in the plugin. Each surface was retained only when grep found a real external consumer; everything else was deleted.

**Stacked on top of #141** (v2.27.2) — that PR's bloat cleanup needs to land first.

### Removed
- `cmd_migrate_critic_gates` subcommand + 4 args + mutex group
- `<pr_verdict>` XML tag + `verdict.json.tag` field
- 10 explicit-path "fallback" CLI args that `stages.json` never passes
- `_read_coverage_verify_verdict`, `discarded_line_not_changed: 0`, partitions bare-list shape acceptance, missing-`partition_count` fallback
- `[DEPRECATED]` warning-text prefix

### Renamed
- `--validate-output` → `--findings-validated` (the flag was misnamed; `stages.json` has always passed `findings_validated.json` to it)

### Kept with grep evidence
- `verdict.json` `.verdict` field — `plugins/code/scripts/run-loop.sh:987`
- `moduleCritics[]` migration — `plugins/bootstrap` actively generates it
- Per-cr-dir cache fallback path — active when `CR_GLOBAL_CACHE=0`
- `normalize_legacy_finding` — 7+ call sites
- Plain-string `route.models` form — `cmd_route` actively produces it

Net: **+299 / -996** lines across 15 files.

## Test plan
- [x] `uv run ruff check .` — All checks passed
- [x] `uv run pyright` — 0 errors / 0 warnings / 0 informations
- [x] `uv run pytest plugins/` — 1051 passed, 5 skipped
- [x] Regenerated 3 snapshot fixtures (cli_parser_resolved, local_no_pr_empty_flags, github_pr42_all_flags)
- [x] Deleted obsolete test classes and refactored shared helpers to seed canonical sections


plugins/code-review/commands/start.md
```diff
@@ -207,7 +207,7 @@ If `FLAGS.hygiene_only` is true (or the equivalent `--hygiene-only` was passed):
 2. Mark "Present hygiene findings" `in_progress`.
 3. Parse `<CR_DIR>/hygiene.json` and render using the "Hygiene Findings Format (Gate A render target)" section below.
 4. If MODE=github, write hygiene findings to `.closedloop-ai/code-review-summary.md` and `.closedloop-ai/code-review-findings.json`; the workflow handles posting.
-5. **EXIT.** Do not run any remaining stages — not route, partition, agents, validate, finalize, cache-update, review-state-write, verdict, or footer. Hygiene-only runs do not emit a `<pr_verdict>` tag (there is no findings_validated.json or review_result.json for verdict to read; invoking it would crash the walker via `on_failure: abort`). This matches the pre-Phase-4b orchestrator behavior.
+5. **EXIT.** Do not run any remaining stages — not route, partition, agents, validate, finalize, cache-update, review-state-write, verdict, or footer. Hygiene-only runs do not emit a `verdict.json` (there is no findings_validated.json or review_result.json for verdict to read; invoking it would crash the walker via `on_failure: abort`).
```
shafty023, line 222: [P1] This update removes the hygiene-only `<pr_verdict>` reference, but the final `PR Verdict` section in this same command still tells the walker to read `verdict.json.tag` and print it. `cmd_verdict` now writes `verdict`, `canonical_verdict`, and `reason` without `tag`, so a literal `/code-review:start` run can omit or fail the UI verdict banner. Update the final PR Verdict section to read the current `verdict.json` fields, or restore the `tag` field if that UI contract still exists.

plugins/code-review/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code-review",
   "description": "Code review plugin",
-  "version": "2.27.2",
+  "version": "2.27.3",
```
shafty023: [P1] This is a breaking interface change under the repo semver rule, but the manifest only moves from `2.27.2` to `2.27.3`. The PR deletes helper CLI surfaces (`migrate-critic-gates` and the legacy coverage/spawn explicit-path flags) and renames `--validate-output` to `--findings-validated`; `CLAUDE.md` requires a MAJOR bump for breaking orchestration/API/skill interface changes. Either bump this plugin to the next major version and update the release notes, or keep compatibility aliases for the 2.x line.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -1670,9 +1670,6 @@ def cmd_validate(args: argparse.Namespace) -> int:
         "cross_file_grouped": cross_file_grouped,
         "kept_out_of_hunk": kept_out_of_hunk,
         "discarded_file_not_changed": sum(1 for d in discarded if d["reason"] == "DISCARD_FILE_NOT_CHANGED"),
-        # Retained at 0 for backward compat with downstream telemetry
-        # dashboards; superseded by ``discarded_out_of_hunk_low_confidence``.
-        "discarded_line_not_changed": 0,
```
shafty023, line 1673: [P2] Removing `discarded_line_not_changed` leaves the GitHub presenter contract stale: `prompts/github-review.md` still tells Step 8 to print `Discarded — line not changed` from `findings_validated.json.stats`. After this deletion that field is absent, so PR summaries can show a blank/stale metric while hiding the current out-of-hunk discard signal. Update the GitHub summary prompt/tests to use `discarded_out_of_hunk_low_confidence` and/or `kept_out_of_hunk`, or keep emitting a compatibility stat.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -7406,27 +7304,16 @@ def cmd_coverage_critic_prepare(args: argparse.Namespace) -> int:
...
-            return 1
-    else:
-        plan_initial = _read_coverage_state(cr_dir).get("initial")
-        if plan_initial is None:
-            print(
-                "Error: coverage.json.initial missing — has stage_14 run?",
-                file=sys.stderr,
-            )
-            return 1
+    # Initial plan lives in coverage.json.initial.
```
shafty023, line 7303: [P3] The implementation now reads only `coverage.json.initial`, but the docstring immediately above still advertises an explicit `--coverage-plan-initial` fallback. Similar stale fallback promises remain for `verify-coverage` / `derive-spawn-spec`, and the coverage section header still lists the deleted `migrate-critic-gates` subcommand. Since this PR removes those CLI surfaces, update the helper docs/comments to match the section-only contract or restore the advertised aliases.

plugins/code-review/tools/python/test_code_review_helpers.py
```diff
@@ -18306,38 +18077,6 @@ def test_computed_choices_resolve_to_sorted_coverage_scopes(self) -> None:
...
-
-    def test_mutex_group_rejects_both_choices(self) -> None:
-        import argparse as _argparse
-        from code_review_helpers import _register_subparsers
-        parser = _argparse.ArgumentParser()
-        sub = parser.add_subparsers(dest="command", required=True)
-        _register_subparsers(sub)
-        with pytest.raises(SystemExit):
-            parser.parse_args([
-                "migrate-critic-gates", "--input", "/i",
```
shafty023, line 18742: [P2] This removes the CLI-loader mutex tests for `migrate-critic-gates`, but an older `TestPR124MutuallyExclusiveDestArgs` still parses the removed command and only asserts `SystemExit`. That test now passes because argparse rejects an unknown command, not because the mutex contract is enforced. Delete or replace the remaining test with an explicit absence assertion for `migrate-critic-gates`, or test mutex routing with a live/synthetic command.

shafty023 (commented):
Please address the inline contract, versioning, and validation-surface issues before merge.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6871,99 +6863,6 @@ def cmd_load_available_reviewers(args: argparse.Namespace) -> int:
...
-        print(f"Error writing {out_path}: {exc}", file=sys.stderr)
-        return 1
-    summary["output"] = str(out_path)
-    json.dump(summary, sys.stdout, indent=2)
-    sys.stdout.write("\n")
-    return 0
-
-
 # ---------------------------------------------------------------------------
 # PLN-725 — Coverage critic
```
thadeusb, line 6864: You deleted `cmd_migrate_critic_gates` in this PR, but the "Phase 2 ships" component list (around line 6017, just above `CHANGE_CLASS_PATH_PATTERNS`) still calls out `migrate-critic-gates` as item #3. Drop item 3 — same cleanup as the docstring shafty flagged at 7307.

(The component-list comment block sits outside this PR's diff hunks, so I've anchored this note at the deletion site for the removed subcommand.)

thadeusb (approved):
Chased the consumer claims and they hold: run-loop.sh only reads .verdict, pr_verdict has zero consumers, findings_validated.json is what stage_22 actually writes so the rename is just fixing a misnamed flag, and cmd_partition always emits the wrapped form. shafty already flagged the drift cluster (orphaned mutex test, the 7307 docstring, discarded_line_not_changed). One more stale spot to add.

plugins/code-review/commands/start.md
wongk: Fixed in 903d601. The "PR Verdict" section in start.md now describes the current `verdict.json` field set (`verdict`, `canonical_verdict`, `reason`) and drops the directive to print a `tag` final line. The UI banner was a no-op contract — `tag` was an XML wrapper around `verdict`, never a separate stable field; the legacy `verdict` field (`approve`/`needs_attention`/`decline`) is what `run-loop.sh:987` actually reads.

plugins/code-review/.claude-plugin/plugin.json
wongk: Bumped to v2.28.0 in 903d601 (MINOR, not PATCH).

Pushing back on MAJOR. The CLAUDE.md rule reads "MAJOR for breaking changes to orchestration/hook API/skill interfaces" — but recent precedent treats internal-only orchestration consolidations as MINOR even when wire formats change:

- v2.27.0: "Coverage decision artifacts consolidated" — 4 wire-format files → 1 sectioned file (`coverage.json`). Strictly breaking for any external consumer; treated as MINOR because stages.json was updated atomically.
- v2.26.0: "CRS Phase D — consolidate spawn artifacts" — same shape: wire-format merger, MINOR.
- v2.25.0: "CRS Phase B — shared helpers for singleton prepare/consolidate" — orchestration internal API change, MINOR.

This PR's removals fit the same pattern: `migrate-critic-gates` and the 10 fallback flags have zero grep hits anywhere in the monorepo outside their own definition + the cli.json spec they were declared in. stages.json (the only invoker) was updated in lockstep. There is no external consumer for the deletion to break.

If you'd like to argue this case is qualitatively different from v2.25.0 / v2.26.0 / v2.27.0, happy to bump to v3.0.0 — but the precedent reads to me as "internal orchestration cleanup is MINOR."

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 903d601. `prompts/github-review.md` line 386 now reads "Discarded — out-of-hunk (low confidence)" and consumes `discarded_out_of_hunk_low_confidence` instead of the removed `discarded_line_not_changed` field.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 903d601. The three stale docstrings advertising removed fallback paths are all cleaned up:

- `cmd_coverage_critic_prepare` no longer mentions `--coverage-plan-initial`
- `cmd_verify_coverage` no longer mentions `--coverage-plan` / `--coverage-plan-initial`
- `cmd_derive_spawn_spec` no longer mentions `--coverage-plan` / `--route`

The PLN-725 component-list comment block at line 6006 was also updated: `migrate-critic-gates` item dropped, replaced with a reference to the inline `migrate_legacy_module_critics` helper.

plugins/code-review/tools/python/test_code_review_helpers.py
wongk: Fixed in 903d601 by deletion. Correct that `TestPR124MutuallyExclusiveDestArgs` was only passing because argparse rejects the unknown `migrate-critic-gates` command, not because the mutex contract is enforced. Deleted the class outright — the mutex contract it tested no longer exists; replacing it with a synthetic-command mutex test would just be exercising argparse itself.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 903d601. The component-list comment block at line ~6006 (you flagged it at 6017) had item #3 enumerate `migrate-critic-gates`; that item is dropped, with a one-liner inline noting that the `migrate_legacy_module_critics` helper still handles the actual moduleCritics[] → coverage[] best-effort migration. Same cleanup as the docstring shafty flagged at 7305.

PR #143, opened 2026-06-04
fix(code-review): close PR #107 staged-scope regression + PR #140 retroactive findings (v2.27.5)

wongk:
## Summary

Four production bugs in current main, each reproduced and pinned by a focused regression test. None of these had test coverage that exercised the production codepath end-to-end.

### Fixes

1. **`/start staged` is broken on main.** PR #107's Phase 4b walker rewrite of `start.md` dropped the `--flag=value` form for scope args. CRS Phase A codified the broken space-separated form into `config/stages.json`. Staged reviews resolve `DIFF_SCOPE` to `--cached`; passed space-separated, argparse rejects it as an unknown option and aborts `stage_05_parse_diff` (`on_failure: abort`). The four scope-consuming stages now emit `--scope=<DIFF_SCOPE>`, restoring the v1.5.5 fix. Reproduced with `python3 code_review_helpers.py parse-diff --scope --cached` → `expected one argument`. From PR #107 review comment.

2. **`stage_14_resolve_coverage` ignored operator coverage config.** The stage's args list omitted `--critic-gates`. `cmd_resolve_coverage` silently falls back to `_EMPTY_CRITIC_GATES` when the arg is absent (no error, no warning), so production runs ignored every configured `coverage[]` rule and migrated `moduleCritics[]` entry. Only the core reviewer floor routed. From PR #140 retroactive comment #1.

3. **Validation-gate `required_sections` was declared-but-unread.** `_build_validation_gates` set `required_sections: {coverage.json: [final]}` on the `stage_16` gate, but the walker contract in `start.md` only checked `outputs` existence. `coverage.json` exists from stage_14 onward (sections accumulate atomically), so the gate fired-true even if stages 15/15b/15c/16 all failed to populate `.final`. Added `evaluate_validation_gate` so the contract is deterministic and testable; the walker shells into it. From PR #140 retroactive comment #2.

4. **BLOCKING verify verdict dropped the BHA core floor.** `cmd_arbitrate_budget` hardcoded `budget.bha_partitions: 0` on the BLOCKING short-circuit. derive-spawn-spec reads that value as `bha_partitions_cap` and skipped every BHA partition with `reason: budget_capped` — contradicting the `start.md` contract that BLOCKING preserves core reviewers and only annotates. BHA is `source: "core"` and survives derive-spawn-spec's plan sanitization; arbitrate-budget now computes BHA with the same floor-honoring formula as the PASS path. Docs-only PRs still get 0 (BHA floor waived on both paths). From PR #140 retroactive comment #5.

### Out of scope

PR #140 retroactive comments #3 (`--manifest` legacy compat) and #4 (`verify-coverage --output` legacy compat) are mooted by PR #142, which deletes those legacy fallback args entirely after grep-evidence audit showed zero consumers. They are not addressed here because their resolution is "delete" not "fix."

## Test plan

- [x] `pytest plugins/` → 1071 pass, 5 skipped, 0 fail
- [x] `uv run ruff check plugins/code-review/` → clean
- [x] `uv run pyright plugins/code-review/tools/python/code_review_helpers.py plugins/code-review/tools/python/test_code_review_helpers.py` → 0/0/0
- [x] Reproduced PR #107 bug: `python3 code_review_helpers.py parse-diff --scope --cached` → `expected one argument` (pre-fix)
- [x] Verified joined form fixes it: `python3 -c "import argparse; p=argparse.ArgumentParser(); p.add_argument('--scope'); print(p.parse_args(['--scope=--cached']).scope)"` → `--cached`

## Commits

Four commits, one per logical fix, so any individual regression can be reverted independently. Version bump and CHANGELOG entry consolidated on the final commit (semver bump covers the whole series).


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9445,6 +9445,52 @@ def _build_validation_gates(cr_dir: str) -> list[dict[str, Any]]:
     ]
 
 
+def evaluate_validation_gate(
```
thadeusb, line 9448: Nothing actually calls this. There's no evaluate-validation-gate subcommand registered, and start.md step 7 tells the walker to load the file as JSON and check the keys itself, not to shell into this helper. So the CHANGELOG line about 'the walker shells into it so the markdown spec and the implementation cannot drift' isn't true, this is only exercised by the tests. And the two have already drifted on day one: this rejects final: None via the isinstance(dict) check, but step 7's prose only requires the section key to be 'present', so the LLM walker would happily pass a null final that this would fail. Either add the subcommand and point start.md step 7 at it, or drop the cannot-drift claim and inline the check, because right now you've got two implementations of the same contract and the Python one is dead.

thadeusb (approved):
Three of the four fixes are solid and the regression tests actually hit the broken codepaths. The gap is fix #3: evaluate_validation_gate is added and tested but nothing calls it, so the 'walker shells into it / cannot drift' claim doesn't hold. Fixable without reworking anything.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Caught it cleanly — went with Option (a) in 211a893.

`cmd_evaluate_gate` added, `evaluate-gate` registered in `cli.json`, `start.md` step 7 rewritten to instruct the walker to invoke `python3 <HELPERS> evaluate-gate --cr-dir <CR_DIR> --after-stage <stage_id>` after every stage. Exit code is the gate verdict; the walker reads `on_failure_action` from `run_plan.json` and applies it. The enforcer is gate-policy-agnostic by design — same helper serves abort, continue, and emit_coverage_gap.

You were also right about the day-one drift: the Python check required `isinstance(state[key], dict)` while the prose only said key presence, so an LLM walker would pass `{"final": null}` while the Python would fail it. Now the walker shells into the Python and the prose-vs-implementation drift is impossible by construction. `start.md` step 7 also spells the dict-value check out explicitly for any future Python-side caller.

Plus two adjacent fixes that the local cr-17347 review surfaced: a type guard on `required_sections[file]` (a bare string would silently iterate per-character) and a word-swap on a test docstring ("erroneously passes" → "erroneously fails").

PR #144, opened 2026-06-05
PLN-807: depth-tiered code review commands + BHA-first budget arithmetic

wongk:
## Summary

- Adds `--depth shallow|standard|deep` flag to `/start` plus `/shallow` and `/deep` command wrappers. shallow runs the built-in fleet only (BHA + BHB + auditor + verifier); standard preserves current behavior; deep reserves a slot for FEA-1401 Impact Analyzer.
- Refactors `cmd_arbitrate_budget` to reserve BHA partitions FIRST (from `_max_bha_partitions_by_loc`) before allocating critics. Fixes the bug where `critic-gates.json` rosters with many triggers crushed BHA to its floor=1 partition on large-LOC PRs.
- Adds `DOMAIN_CRITIC_CAP = 5` capping total domain-critic count across both buckets. Required-bucket critics dropped by the cap emit coverage-gap findings. PRs with sparse critic-gates rosters are unchanged.
- Adds tier-aware `review_state.json`: a cached shallow review does not satisfy a subsequent standard or deep invocation.
- Adds tier-mismatch nudge: shallow on a > 3000 LOC PR or one that touches schema/migrations/public-API surfaces emits a single LOW finding suggesting `--depth standard`.

Implementation plan: [PLN-807](https://app.closedloop.ai/closedloop-ai/implementation-plans/PLN-807).

Bare `/code-review start` (no `--depth`) preserves historical behavior — standard is the default and standard-mode budget changes only narrow the fleet when critic-gates triggers exceed 5.

## Test plan

- [x] 1133 existing + 76 new unit tests pass (`uv run pytest plugins/code-review/tools/python/`)
- [x] `uv run ruff check plugins/code-review/` clean
- [x] `uv run pyright plugins/code-review/tools/python/code_review_helpers.py` clean (0 errors)
- [x] Snapshot fixtures regenerated for new field + new stage + new CLI flags
- [x] Backwards-compat verified: bare `cmd_prepare_run` and bare `cmd_review_state_*` calls (no `--depth`) preserve pre-PLN-807 behavior
- [x] BLOCKING-verdict branch preserves "no required drops on BLOCKING" semantics under the new cap
- [ ] Manual smoke test on a real PR with `--depth shallow`, `--depth standard`, `--depth deep`
- [ ] Manual verification of tier-mismatch nudge firing on a schema-touching shallow PR

## Phase commits

1. v2.28.2 — `min_depth` field on stages.json + filtering
2. v2.28.3 — tier propagation + review_state tier persistence
3. v2.28.4 — static spawn spec for shallow tier
4. v2.28.5 — BHA-first budget arithmetic + DOMAIN_CRITIC_CAP=5
5. v2.28.6 — tier-mismatch nudge in hygiene
6. v2.28.7 — command surfaces + start.md docs
7. v2.29.0 — README + CHANGELOG + final MINOR bump


plugins/code-review/commands/start.md
```diff
@@ -111,9 +123,12 @@ python3 <HELPERS> prepare-run \
   --full-review <FULL_REVIEW> \
   --base-ref-override "<BASE_REF_OVERRIDE>" \
   --scope-args "<SCOPE_ARGS>" \
+  --depth <DEPTH> \
```
shafty023, line 133: `<DEPTH>` is now passed to `prepare-run`, but Stage 0 never initializes or consumes `--depth`. A direct `/start --depth shallow` leaves `--depth shallow` in `SCOPE_ARGS` and can pass an unresolved `<DEPTH>` into the helper, so the advertised tier may not actually run. Please default `DEPTH` to `standard`, parse and validate `--depth shallow|standard|deep`, and remove the flag/value before computing `SCOPE_ARGS` with a regression or smoke case for direct `/start --depth shallow`.

plugins/code-review/tools/python/config/stages.json
```diff
@@ -114,7 +118,8 @@
         "stage_04_finalize_cache"
       ],
       "on_failure": "continue",
-      "enabled": true
+      "enabled": true,
+      "min_depth": "shallow"
```
shafty023, line 124: `stage_07_auto_incremental` now runs in every depth tier, but its args still do not include the requested depth. `cmd_auto_incremental` reads `review_state.json` directly, so a cached shallow review can seed `last_sha` for a later standard/deep run and skip older files that never received premise/critic coverage. Thread `--depth {depth}` through this stage and the CLI, then gate cached entries with `_entry_satisfies_depth` before reusing `sha`.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -682,6 +682,108 @@ def _check_sensitive_files(
...
+        reasons.append(f"public API surface{'s' if len(api_hits) > 1 else ''} touched: {sample}{more}")
+
+    if not reasons:
+        return []
+
+    return [{
+        "reviewer": "hygiene",
+        "source": "hygiene",
+        "finding_scope": "system",
+        "system_marker": "tier_mismatch_nudge",
```
shafty023, line 795: `tier_mismatch_nudge` is emitted as a system finding, but this marker is not in the canonical system-marker enum, and the finding uses `severity: "LOW"` while the canonical schema only accepts BLOCKING/HIGH/MEDIUM and validation discards low severities. The nudge can be produced by hygiene and then disappear before final output. Register a schema-valid marker/severity, or add a supported non-blocking channel, and cover it through validate/finalize instead of only testing `_check_tier_mismatch_nudge`.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -682,6 +682,108 @@ def _check_sensitive_files(
...
+    reasons: list[str] = []
+
+    if total_loc > _TIER_MISMATCH_NUDGE_LOC_THRESHOLD:
+        reasons.append(
+            f"diff size ({total_loc} LOC) exceeds {_TIER_MISMATCH_NUDGE_LOC_THRESHOLD} LOC"
+        )
+
+    schema_hits = [
+        f for f in files
+        if any(p in f for p in _TIER_MISMATCH_NUDGE_SCHEMA_PATTERNS)
```
shafty023: The schema/migration heuristic only checks substrings with a leading slash, so root-level paths like `migrations/0001.sql`, `schemas/user.py`, or `models/user.rb` will not fire the shallow mismatch nudge. Use path-aware matching that covers both root and nested `migrations`, `schemas`, and `models` directories.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9701,41 +10108,116 @@ def _plan_target_str() -> str:
...
-    if _is_docs_only(diff_data):
-        bha_partitions = 0
-    else:
-        leftover = max(0, cap - len(required) - len(best_effort_final))
-        bha_partitions = max(bha_floor, min(leftover, max_bha))
+    # Annotate cap-deferred entries so operators can tell which bucket
+    # capacity rejected them. Budget-deferred entries carry no
+    # ``defer_reason`` — that's the historical default and the absence
+    # is the signal.
+    deferred_cap_marked = _annotate_defer_reason(
```
shafty023, line 10308: Only `def_be_critics` are annotated into `deferred_for_budget`; required critics rejected by `DOMAIN_CRITIC_CAP` are omitted from the machine-readable deferred list. The exposed contract says required cap drops should both emit coverage gaps and appear in `deferred_for_budget` with `defer_reason: "domain_critic_cap"`, so operators and telemetry underreport the cap decision. Include `def_req_critics` in the annotated cap-deferred list while preserving the coverage-gap findings.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -10317,6 +10801,119 @@ def cmd_derive_spawn_spec(args: argparse.Namespace) -> int:
...
+    agents, skipped = _derive_spawn_agents_from_plan(
+        static_plan, partitions, models,
+        bha_partitions_cap=DEFAULT_MAX_BHA_AGENTS,
+    )
+
+    bha_count = sum(1 for a in agents if a["reviewer"] == "bug_hunter_a")
+    spec = {
+        "fast_path": False,
+        "gated_by_verify": False,
+        "arbitrate_status": "static",
```
shafty023, line 10993: This emits a new `arbitrate_status: "static"`, but `code_review_schema.SPAWN_SPEC_ARBITRATE_STATUSES` still only allows `ok`, `blocked_by_verify`, and `fallback`. Consumers or tests using the shared schema can reject or mishandle shallow static specs. Add `static` to the canonical enum and pin a schema/compatibility test for a generated shallow spec.

plugins/code-review/tools/python/config/stages.json
```diff
@@ -547,7 +572,30 @@
...
+        "--cr-dir",
+        "{cr_dir}",
+        "--partitions",
+        "{cr_dir}/partitions.json"
+      ],
+      "stdout": null,
+      "expected_outputs": [
+        "{cr_dir}/spawn.json"
+      ],
+      "depends_on": [
```
shafty023, line 594: `stage_19c_derive_static_spec` has a hard dependency on `stage_17_partition`, but Gate B explicitly skips `stage_17_partition` when `FAST_PATH == true`; the helper even has a fast-path branch that does not need `partitions.json`. In a dependency-aware walk, shallow fast-path reviews can skip the static spec before it emits the single fast reviewer. Make this stage fast-path-safe by removing the partition hard dependency or splitting the fast-path/static-spec path.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9701,41 +10108,116 @@ def _plan_target_str() -> str:
...
+    # Required critics get priority over best-effort, sorted within
+    # bucket by (priority asc, reviewer name asc). Excess required
+    # critics emit coverage-gap findings AND land in
+    # ``deferred_for_budget`` with ``defer_reason: "domain_critic_cap"``.
+    sel_req_critics, sel_be_critics, def_req_critics, def_be_critics = (
+        _select_domain_critics(
+            required_critics_in, best_effort_critics_in, DOMAIN_CRITIC_CAP,
+        )
+    )
+    for idx, entry in enumerate(def_req_critics):
```
thadeusb: A required critic that gets cap-dropped here lands in coverage_gaps with required:True (via _make_coverage_gap_finding), and Rule 1 in _compute_canonical_verdict turns any required coverage gap into CHANGES_REQUESTED. I traced it: cmd_finalize_result glob-reads coverage_gaps.json straight into the verdict. So a repo whose critic-gates.json resolves 6+ required domain critics on a PR now gets auto-CHANGES_REQUESTED on every review, code quality aside. That's not 'narrowing the fleet' like the summary says, it's a hard block at a hardcoded threshold of 5. And it's the inverse of your BLOCKING branch, which deliberately pushes the same def_req_critics into deferred_for_budget with no gap. Was the block intended? If not, route def_req_critics to deferred_for_budget here like the BLOCKING path does. Separately, on this path the cap-dropped required critics show up in neither deferred_for_budget nor dropped_required, only as gaps, so the plan itself can't tell you what got cut.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -9701,41 +10108,116 @@ def _plan_target_str() -> str:
...
+        coverage_gaps.append(
+            _make_coverage_gap_finding(
+                entry,
+                reason=_DEFER_REASON_DOMAIN_CRITIC_CAP,
```
thadeusb: _(Anchored on this cap-reason call site; the recommendation text itself lives in the unchanged `_make_coverage_gap_finding`, outside this diff — but this is the new call that routes the domain_critic_cap reason into it.)_

This recommendation is wrong for the domain_critic_cap case. DOMAIN_CRITIC_CAP is hardcoded at 5 and independent of --cap, so 'Raise --cap' does nothing when the entry was dropped by the critic cap. Operator follows it, bumps --cap, critic's still gone. Fine for budget_exceeded, just not this reason.

thadeusb (commented):
BHA-first and the tiering read right, but I want to settle the required-critic cap before this lands. On the normal path, capping a required critic emits a required:True coverage gap, which forces CHANGES_REQUESTED, so a repo whose critic-gates resolves >5 required critics auto-blocks every review. Your BLOCKING branch does the exact opposite. Flagged that plus a dead-end recommendation.

shafty023:
Non-inline finding: `plugins/code-review/README.md` still advertises `.claude-plugin/plugin.json` as `version 2.6.0` in the architecture tree while this PR bumps the manifest to `2.29.0`. Since this PR updates the README for the same release, the generated docs still publish stale runtime metadata. Remove the hardcoded manifest version or regenerate/update it from `plugin.json` so docs do not drift on every release.

plugins/code-review/commands/start.md
wongk: Fixed in c7b80a6. `start.md` Stage 0a now initializes `DEPTH = "standard"`, parses `--depth shallow|standard|deep`, validates the value (rejects others), and removes both tokens from `$ARGUMENTS` before computing `SCOPE_ARGS`. `test_stage_12_hygiene_passes_depth` + `test_stage_07_auto_incremental_passes_depth` cover the end-to-end thread.

plugins/code-review/tools/python/config/stages.json
wongk: Fixed in c7b80a6. `cmd_auto_incremental` now accepts `--depth`, validates it via `_validate_invocation_depth`, and gates both cached-entry-read paths (since-last-review error path + auto-incremental scope path) with `_entry_satisfies_depth`. A weaker cached tier short-circuits to full diff. `stage_07_auto_incremental` threads `--depth {depth}` via the template context. Three tests pin the behavior: `test_auto_incremental_skips_when_cached_tier_weaker`, `test_since_last_review_rejects_weaker_cached_tier`, `test_auto_incremental_without_depth_preserves_legacy_behavior`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c7b80a6. Severity bumped LOW → MEDIUM (confirmed `SEVERITY_NORMALIZE["low"] = "DISCARD"`, which would have silently filtered the nudge). MEDIUM survives validate without escalating any verdict rule because category is `Coverage` (not `Premise`), so Rule 4's cumulative-Premise-MEDIUM gate doesn't fire. The system_marker stays `tier_mismatch_nudge` — these are descriptive strings rather than an enum (matches the existing `coverage:*`, `budget-exceeded`, etc. patterns), so registration is a non-issue.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c7b80a6. New `_matches_schema_or_migration_path` helper splits on `/` and looks for any directory segment in `{migrations, schemas, models}`, plus a special-filename check for `schema.prisma` / `schema.sql` at any depth. Root-level and nested paths both fire; substring false-positives (`migration_history.txt`, `model_loader.py`) don't. Five new tests cover root, nested, false-positive, and filename cases.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c7b80a6. After re-reading `_compute_canonical_verdict` Rule 1, emitting coverage-gap findings for cap-deferred required critics would auto-CHANGES_REQUESTED any repo whose critic-gates.json resolved >5 required critics — which thadeusb also caught on the line below. Routed `def_req_critics` into `deferred_for_budget` with `defer_reason: "domain_critic_cap"` and dropped the coverage-gap emission entirely. The PASS path now matches the BLOCKING-branch behavior. Pinned by `test_cap_deferred_required_routed_to_deferred_for_budget` (asserts `gaps == []` AND verdict computed from the gaps is `APPROVED`).

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c7b80a6. `"static"` added to `SPAWN_SPEC_ARBITRATE_STATUSES` in `code_review_schema.py` with an inline rationale describing the user-intent vs upstream-failure distinction from `"fallback"`. `test_static_in_canonical_arbitrate_statuses` pins the enum membership; existing Phase 3 tests already exercise the static-spec generation path end-to-end against this enum.

plugins/code-review/tools/python/config/stages.json
wongk: Fixed in c7b80a6. Changed `stage_19c_derive_static_spec.depends_on` from `["stage_17_partition"]` to `["stage_19_cache_check"]` — that's the latest always-running stage before Gate B (per the array-order rule in start.md, stage_19_cache_check runs before stage_17 even though the id is higher). `cmd_derive_static_spec`'s existing fast-path branch already handles missing/empty partitions.json gracefully (emits the single fast agent). `test_stage_19c_does_not_hard_depend_on_partition` pins the dependency change.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good catch — confirmed by tracing through `_compute_canonical_verdict`. Rule 1 turns any required coverage gap into CHANGES_REQUESTED, so the original PASS-path code was a hardcoded block-at-5 disguised as a soft cap. Fixed in c7b80a6 by routing `def_req_critics` to `deferred_for_budget` with `defer_reason: "domain_critic_cap"` and DROPPING the coverage-gap emission entirely. The PASS path now matches your BLOCKING-branch contrast exactly. Both buckets' cap-deferred entries appear in `deferred_for_budget` (annotated, so a single read of the plan tells the operator everything the cap dropped), and `dropped_required` is empty on cap-only drops. Pinned by `test_cap_deferred_required_routed_to_deferred_for_budget` which asserts `_compute_canonical_verdict(verified=[], coverage_gaps=gaps) == "APPROVED"`. CHANGELOG v2.29.1 has the full explanation.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Confirmed and resolved as a side-effect of the cap-routing fix on the line above. With `def_req_critics` now landing in `deferred_for_budget` instead of going through `_make_coverage_gap_finding`, the "Raise --cap" recommendation never fires for the cap reason — the recommendation text was only ever wrong for cap drops, and cap drops no longer emit findings at all. The `budget_exceeded` path (where the recommendation IS valid) is untouched. See c7b80a6.

wongk:
Non-inline comment fixed in c7b80a6. `plugins/code-review/README.md` no longer hardcodes a manifest version in the architecture tree; the line now points at the `version` field in `plugin.json` directly so the README doesn't drift on future bumps. v2.29.1 published the CHANGELOG entry.

PR #147, opened 2026-06-05
feat(code-review): optional codebase-memory-mcp graph integration (v2.31.0)

wongk:
## Summary

Enables the code-review plugin to opportunistically use the **`codebase-memory-mcp`** knowledge graph when it's connected and the repo is already indexed. The two cross-file reviewers — the **Impact Analyzer** (deep tier) and **Bug Hunter B** (every standard review) — prefer graph queries over grep for cross-file usage discovery, catching re-exports, aliased imports, and dynamic dispatch that grep misses, and resolving service/API implementations by qualified name instead of Glob-guessing the file.

**Activation: auto-detect + silent fallback.** Each relevant reviewer probes `index_status` once; uses the graph only when the repo is indexed; otherwise falls back to Grep/Glob silently. Behavior is byte-identical when the MCP server isn't installed. Reviews never trigger `index_repository` (no review-time indexing stalls).

## Scope decisions

- **Impact Analyzer + Bug Hunter B** — the two reviewers whose work is definitionally cross-file. BHB runs on every standard review, so even modest per-lookup gains compound.
- **Finding Verifier left frozen** — its primary audit (read callsite + `callsite_snippet_hash`) is already substrate-agnostic, and the grep-replay sub-check keeps working because the Impact Analyzer still emits a real `grep_query_used` for every finding regardless of how it discovered the callsite. Graph-based completeness auditing is deferred as a possible follow-up.

## Changes

| File | Change |
|---|---|
| `agents/code-review-worker.md` | Add six read-only `mcp__codebase-memory-mcp__*` tools to the allowlist (the gating change). Write/index tools excluded. |
| `tools/prompts/shared_prompt.txt` | Single detection/fallback protocol (read by all roles) + untrusted-content policy extended to graph outputs. |
| `tools/prompts/impact_analyzer_prompt.txt` | Graph-preferred caller enumeration with grep fallback; `grep_query_used` + snippet hash restated as the verifier proof-of-record; tool list + injection policy extended. |
| `commands/start.md` | BHB suffix + fast-path PASS 2 + Impact dispatch graph guidance; corrected deep-tier "greps the codebase" wording. |
| `.claude-plugin/plugin.json` | `2.30.5` → `2.31.0` (MINOR, optional capability, no breaking change). |

## Verifier contract (the one must-not-miss constraint)

The Impact Analyzer **always** records a real `grep_query_used` and reads each `callsite_snippet` (+ hash) even when it discovered callsites via the graph. The graph speeds discovery; grep remains the reproducible proof the verifier replays. The verifier prompt and golden fixtures are untouched.

## Validation

- `uv run pyright` — 0 errors
- `pytest plugins/code-review/tools/python/` — **1188 passed, 5 skipped** (no fixture drift)
- `pytest plugins/judges/tools/python/test_validate_agent_registry.py` — 38 passed; the new `mcp__` tools validate


plugins/code-review/tools/prompts/impact_analyzer_prompt.txt
```diff
@@ -97,6 +124,13 @@ underscored, inside `__init__` blocks not in `__all__`, etc.). Ignore
 
 For each candidate symbol:
 
+  - **If the knowledge graph is available** (per the Inputs section),
+    enumerate callers first with `search_graph(qn_pattern=<symbol>)` +
+    `trace_path(<symbol>, mode=calls)`, then still construct the grep
+    query below so `grep_query_used` is populated for the verifier — the
```
thadeusb: The verifier contract says `grep_query_used` must find "those same external usages" and the cross-check instruction says to "trust the union" of graph and grep results, putting graph-only callsites into `external_usages_found`. But the verifier replays `grep_query_used` and compares against `external_usages_found` (line 319 of `verifier_prompt.txt`), so graph-only callsites that grep literally can't surface (aliases, dynamic dispatch) lower the overlap. If >50% of `external_usages_found` entries are graph-only, the verifier rejects the finding even though every callsite was individually verified by reading the file.

This is the exact scenario the graph is supposed to help with, so the feature partially defeats itself on the hardest cases. Not blocking since the common case (graph refining what grep also finds) is fine, but worth a tracking note. A follow-up could either split `external_usages_found` into graph-discovered and grep-discovered buckets, or teach the verifier to accept graph-verified callsites without requiring grep reproducibility.

thadeusb (approved):
Auto-detect + silent fallback is the right shape, injection policy correctly extended, verifier contract preserved. One design tension in the impact analyzer worth tracking.

shafty023:
@wongk make sure as part of this, any github workflow for our claude reviews has these new MCP servers added in `--allowedTools`

image

plugins/code-review/tools/prompts/impact_analyzer_prompt.txt
```diff
@@ -97,6 +124,13 @@ underscored, inside `__init__` blocks not in `__all__`, etc.). Ignore
...
+    graph tells you which callers exist; the grep query is the
+    reproducible proof. Cross-check the two: if grep surfaces callsites
```
shafty023: The new union rule conflicts with the verifier contract. The verifier still replays `finding.grep_query_used` and compares the hits with `reasoning_certificate.fields.external_usages_found`; if the graph finds alias/re-export/dynamic callsites that grep misses, adding them to the union makes the finding fail or downgrade even though those are the cases this feature is meant to catch. Please either keep `external_usages_found`/`external_impact` grep-replayable under the unchanged verifier, or add graph provenance fields plus verifier/schema/tests that validate graph-discovered callsites.

plugins/code-review/agents/code-review-worker.md
```diff
@@ -1,7 +1,7 @@
 ---
 name: code-review-worker
 description: Worker agent for partitioned code review tasks. Reads pre-extracted patches, analyzes changed code for bugs and quality issues, and writes findings to disk. Use when spawning background review agents that need file access.
-tools: Read, Write, Grep, Glob
+tools: Read, Write, Grep, Glob, mcp__codebase-memory-mcp__index_status, mcp__codebase-memory-mcp__search_graph, mcp__codebase-memory-mcp__trace_path, mcp__codebase-memory-mcp__get_code_snippet, mcp__codebase-memory-mcp__search_code, mcp__codebase-memory-mcp__query_graph
```
shafty023: This expands the base `code-review-worker` allowlist, so the graph tools are available to every task that uses this agent, not just BHB/Impact. `start.md` also spawns verifier and PLN-725 singleton agents with `subagent_type: "code-review:code-review-worker"`, and those prompts still assume only Read/Grep/Glob and do not load the graph protocol. That widens the trust boundary for verifier input/source snippets and singleton prompts. Please split a graph-enabled worker for the graph-aware roles, or otherwise keep these MCP tools out of the generic worker used by verifier/singleton dispatches.

plugins/code-review/tools/prompts/shared_prompt.txt
```diff
@@ -21,6 +23,28 @@ TOOL USAGE:
...
+role prompt to prefer the codebase knowledge graph for cross-file lookups when
+it is available. If — and ONLY if — your role prompt directs you to use it:
+- DETECT FIRST: call `mcp__codebase-memory-mcp__index_status` once. Treat the
+  graph as available only when the tool answers AND reports this repo as
+  indexed. If the tool is unavailable (the MCP server is not connected), errors,
+  or reports the repo as not indexed, silently fall back to Grep/Glob — do NOT
+  mention the absence in your findings, and do NOT change what you report.
+- NEVER index: do not call `index_repository` or any write/index tool (they are
+  not in your allowlist). Reviews must never block on indexing a repo.
+- WHEN AVAILABLE, prefer these read-only tools for cross-file work:
```
shafty023: The protocol checks whether "this repo" is indexed but never tells graph calls how to stay in that repo. codebase-memory-mcp can hold multiple indexed projects and its troubleshooting docs say wrong-project results require a `project` filter, but `search_graph`/`trace_path`/`get_code_snippet`/`search_code`/`query_graph` here are allowed without a project/root constraint or returned-path validation. That can pull another indexed repo's source into review findings or GitHub comments. Please resolve the current project, pass it to graph calls, and reject any returned file outside the current checkout; if `query_graph` cannot be constrained, leave it out of the allowlist.

wongk:
> @wongk make sure as part of this, any github workflow for our claude reviews has these new MCP servers added in `--allowedTools`
> 
> image

this is optional and won't really make much difference for GH reviews.

wongk:
Thanks @thadeusb and @shafty023 — all three issues were valid (and #3 was a real gap, not just a nicety, since `project` is a *required* arg on every graph tool). Addressed in 7370086:

**1. Verifier-contract tension (the union vs. grep-replay) — @thadeusb's note + @shafty023's :132**
You were right that "trust the union" made graph-only callsites (aliases, re-exports, dynamic dispatch) tank the replay overlap and self-reject — the exact cases the graph exists to catch. Rather than constrain to grep-reproducible (which would defeat the point), I went provenance-aware:
- Each `external_impact[]` entry now carries `discovery: "grep" | "graph"`.
- The grep-replay completeness check is scoped to the grep-discovered set only (`external_usages_found`). Graph-only usages go in a new `graph_discovered_usages` certificate field and are **not** counted as missing replay hits; an all-graph finding (empty `external_usages_found`) skips the replay gate entirely.
- Graph entries are verified by the per-entry file-read + `callsite_snippet_hash` audit, which was already substrate-agnostic. A graph entry that fails read/hash is still trimmed on DOWNGRADE just like a grep entry — provenance doesn't exempt it from verification.

**2. Allowlist reaches verifier/singletons — @shafty023 worker.md:4**
Split a dedicated `code-review-worker-graph` agent. Only Bug Hunter B, the Impact Analyzer, and the fast-path reviewer spawn as it; the generic `code-review-worker` (verifier fleet, PLN-725 singletons, and every other reviewer) reverts to `Read, Write, Grep, Glob` only. No graph access for the adversarial/verification roles.

**3. Multi-project / cross-repo leak — @shafty023 shared_prompt.txt:37**
The orchestrator now resolves `GRAPH_PROJECT` once via `list_projects`, matching it to the current checkout (empty/ambiguous → grep-only, never guess), and threads it into the graph-aware prompts. Agents pass `project=<GRAPH_PROJECT>` on every call and validate that returned paths are inside the checkout before citing them. Dropped `query_graph` from the allowlist (raw Cypher — can't be path-constrained), leaving the four scopeable read tools.

Schema: `ExternalImpact.discovery` + `EXTERNAL_IMPACT_DISCOVERY` vocabulary, validated in `validate_finding`, documented in `SCHEMA.md`. New tests cover the vocabulary and the substrate-agnostic DOWNGRADE-trim (verified graph entries survive even when a grep entry is rejected; unverified graph entries are dropped). 1235 tests pass; ruff + pyright clean.

plugins/code-review/tools/prompts/impact_analyzer_prompt.txt
wongk: Fixed in 7370086 — went provenance-aware rather than constraining to grep-reproducible. Entries now carry `discovery: grep|graph`; grep-replay completeness is scoped to the grep set (`external_usages_found`), graph-only callsites go in a new `graph_discovered_usages` certificate field and are verified by the per-entry read+hash audit instead of the replay. So alias/dynamic callsites no longer lower the overlap or self-reject — exactly the split you suggested.

plugins/code-review/tools/prompts/impact_analyzer_prompt.txt
wongk: Fixed in 7370086. Added graph-provenance fields plus verifier + schema + tests: `external_impact[].discovery` (grep|graph), a `graph_discovered_usages` certificate field separate from the grep-replayable `external_usages_found`, verifier scopes its grep-replay completeness check to the grep set only (graph entries verified by read+hash), `EXTERNAL_IMPACT_DISCOVERY` vocab validated in `validate_finding`, and tests for the substrate-agnostic DOWNGRADE-trim. The union no longer breaks the contract.

plugins/code-review/agents/code-review-worker.md
wongk: Fixed in 7370086 — split a dedicated `code-review-worker-graph` agent. Only Bug Hunter B, the Impact Analyzer, and the fast-path reviewer spawn as it; the generic `code-review-worker` (verifier fleet, PLN-725 singletons, and all other reviewers) reverts to `Read, Write, Grep, Glob` only. No graph tools in the trust boundary for the adversarial/verification roles.

plugins/code-review/tools/prompts/shared_prompt.txt
wongk: Fixed in 7370086. The orchestrator resolves `GRAPH_PROJECT` once via `list_projects`, matching it to the current checkout (empty/ambiguous → grep-only, never guess), and threads it into the graph-aware prompts. Agents pass `project=<GRAPH_PROJECT>` on every call and validate returned paths are inside the checkout before citing them. `query_graph` is dropped from the allowlist since raw Cypher can't be path-constrained, leaving the four scopeable read tools.

PR #148, opened 2026-06-08
fix(code-review): suppress out-of-hunk findings from inline posting (v2.31.1)

wongk:
## Problem

GitHub code reviews stopped leaving inline comments on PRs that **move or rename files**. The summary comment still posted, but no inline (line-anchored) comments appeared.

Root cause: GitHub's `POST /repos/{owner}/{repo}/pulls/{pr}/comments` rejects any inline comment whose line is **not part of the PR diff** with a `422`. On moved/renamed files, reviewers flag pre-existing lines that fall outside every diff hunk. The validator deliberately keeps high-confidence out-of-hunk findings (tagged `out_of_hunk_kept`), but `cmd_post_comments` then tried to post all of them inline — every one `422`'d, yielding `Posted 0, failed N` and zero inline comments. The findings survived only in the summary, which is exactly the "summary posts, no inline" symptom.

Observed in the wild:
```
Skipped packages/app/shared/lib/s3-upload.ts:1 — line not in diff (422)
Skipped packages/app/loops/lib/compute-target-conflict.ts:20 — line not in diff (422)
...
Posted 0, skipped 0 (dedup=0, non-inline=0), failed 4
```

## Fix

`cmd_post_comments` now:
- **Pre-skips** findings tagged `out_of_hunk_kept` (no API call) — those lines are provably outside the diff and cannot be anchored.
- **Reclassifies the runtime 422** ("line not in diff") from `failed` to a non-inline skip, covering moved/renamed files where the `-U0` diff parser and GitHub disagree on what is in-diff even for findings the validator counted as in-range.

These findings still surface in the **summary comment** (it renders all validated findings) — the chosen behavior is summary-only suppression, not a separate inline fallback. The posting tally gains an `out-of-hunk=` counter so the case is visible instead of silently reported as a failure:

```
Posted N, skipped M (dedup=…, non-inline=…, out-of-hunk=K), failed N
```

## Notes / scope

This stops the silent failure and keeps these findings in the summary, but does **not** make moved-file findings appear inline — GitHub fundamentally cannot anchor a comment to a line outside its diff. A better follow-up (separate PR) is upstream: not feeding reviewers the full body of pure-rename files so they stop flagging unchanged lines in the first place.

## Changes

- `tools/python/code_review_helpers.py` — `cmd_post_comments` pre-skip + 422 reclassification + counter
- `prompts/github-review.md` — step 6b documents poster-owned out-of-hunk suppression (findings file must still include all verified findings)
- tests — `test_422_counted_as_skip_not_failure`, `test_out_of_hunk_kept_skips_inline_post`; updated `test_422_continues`
- `plugin.json` — `2.31.0` → `2.31.1` (PATCH)

## Verification

- `uv run pytest plugins/code-review/tools/python/` → `1207 passed, 5 skipped`
- `uv run ruff check` + `uv run pyright` clean on changed files

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code-review v2.31.1
+
+PATCH bump — fix GitHub-mode inline comments silently dropping when a PR moves or renames files. GitHub's `pulls/{pr}/comments` API rejects any inline comment whose line is not part of the PR diff with a 422; reviewers flagging lines outside the changed hunks (common on moved/renamed files) therefore failed to post and survived only in the summary.
+
+#### Fixed
+- `cmd_post_comments` no longer attempts an inline post for findings the validator tagged `out_of_hunk_kept` — those lines are outside the diff and cannot be anchored — and now treats a runtime 422 ("line not in diff") as a non-inline skip rather than a failure, covering moved/renamed files where the diff parser and GitHub disagree on what is in-diff. The affected findings still appear in the summary comment, and the posting tally gains an `out-of-hunk=` counter so the case is visible instead of reported as a failure.
+
+#### Changed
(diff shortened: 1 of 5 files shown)


thadeusb (approved):
Correct fix, both the pre-skip and the 422 reclassification are right. Tests pass.

PR #149, opened 2026-06-08
feat(code): constrain harness adapter/method names to closed enums

aponamarev:
## Summary

Tightens the harness orchestration contract so that the runner (adapter) name and the dispatched method name are no longer arbitrary strings. Both are now closed `StrEnum` categories — `AdapterName` (`claude`, `codex`) and `MethodName` (the five adapter methods) — validated at the CLI boundary. The CLI dispatcher uses an exhaustive `match`/`case` with `assert_never`, so pyright proves every method is handled and flags any future enum member that lacks a case.

The net effect: every production adapter registers under a known runner name rather than a free-form string, and an unsupported `<adapter>`/`<method>` CLI argument fails fast with a descriptive error naming the allowed set.

## Scope decisions

- **Validation lives at the edge** — `adapter_name_from_str` / `method_name_from_str` convert raw CLI args once; everything downstream operates on validated enum members ("parse, don't validate").
- **One shared converter** — both helpers delegate to a single `_str_enum_from_value` so the "validate a stringly-typed arg against a closed category" behavior has a single source of truth.
- **Registry re-typed by `AdapterName`** — `_REGISTRY`, `register`, and `get_adapter` are keyed by the enum, making the `HarnessAdapter.name` ClassVar a closed type.

## Changes

| File | Change |
|---|---|
| `harness/types.py` | Add `AdapterName` / `MethodName` StrEnums, shared `_str_enum_from_value`, and `adapter_name_from_str` / `method_name_from_str` helpers. |
| `harness/cli.py` | Validate both CLI args at the boundary; dispatch via exhaustive `match`/`case` + `assert_never`. |
| `harness/adapter.py` | Type `HarnessAdapter.name` as `ClassVar[AdapterName]`. |
| `harness/registry.py` | Key `_REGISTRY` / `register` / `get_adapter` by `AdapterName`. |
| `harness/__init__.py` | Re-export the new enums and conversion helpers. |
| `conftest.py`, `test_harness_*.py` | Update fakes to use `AdapterName` members; add enum-value and string-conversion (valid + invalid) coverage. |
| `.claude-plugin/plugin.json` | `1.12.4` → `1.13.0` (MINOR — additive public surface, no breaking change). |

## Validation

- `uv run pyright` — 0 errors
- `pytest plugins/code/tools/python/test_harness_types.py test_harness_cli.py test_harness_registry.py` — passing

### DIFF
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.12.4",
+  "version": "1.13.0",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/code/.claude-plugin/plugin.json
+++ b/plugins/code/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.13.0",
+  "version": "1.12.5",
   "author": {
     "name": "ClosedLoop",
     "email": "support@closedloop.ai"
--- a/plugins/code/tools/python/harness/registry.py
+++ b/plugins/code/tools/python/harness/registry.py
@@ -6,14 +6,16 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
+from harness.types import AdapterName
+
 if TYPE_CHECKING:
     from harness.adapter import HarnessAdapter
 
 # Module-level registry: maps adapter name -> adapter class.
-_REGISTRY: dict[str, type["HarnessAdapter"]] = {}
+_REGISTRY: dict[AdapterName, type[HarnessAdapter]] = {}
 
 
-def register(adapter_cls: type["HarnessAdapter"]) -> type["HarnessAdapter"]:
+def register(adapter_cls: type[HarnessAdapter]) -> type[HarnessAdapter]:
     """Register an adapter class under its ``name`` attribute.
 
     Can be used as a class decorator::
@@ -37,13 +39,13 @@ def register(adapter_cls: type["HarnessAdapter"]) -> type["HarnessAdapter"]:
     return adapter_cls
 
 
-def get_adapter(name: str) -> type["HarnessAdapter"]:
+def get_adapter(name: AdapterName) -> type[HarnessAdapter]:
     """Return the registered adapter class for ``name``.
 
     Parameters
     ----------
     name:
-        The adapter's ``name`` attribute value.
+        The adapter's ``name`` attribute value, a member of ``AdapterName``.
 
     Raises
     ------
@@ -55,7 +57,6 @@ def get_adapter(name: str) -> type["HarnessAdapter"]:
     if name not in _REGISTRY:
         available = ", ".join(sorted(_REGISTRY)) or "(none)"
         raise KeyError(
-            f"No adapter registered with name {name!r}. "
-            f"Available adapters: {available}"
+            f"No adapter registered with name {name!r}. Available adapters: {available}"
         )
     return _REGISTRY[name]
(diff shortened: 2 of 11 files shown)


plugins/code/tools/python/harness/registry.py
```diff
@@ -6,14 +6,16 @@
...
 from typing import TYPE_CHECKING
 
+from harness.types import AdapterName
+
 if TYPE_CHECKING:
     from harness.adapter import HarnessAdapter
 
 # Module-level registry: maps adapter name -> adapter class.
-_REGISTRY: dict[str, type["HarnessAdapter"]] = {}
+_REGISTRY: dict[AdapterName, type[HarnessAdapter]] = {}
```
shafty023, line 15: This types the registry as `AdapterName`, but `register()` still accepts and stores any runtime `adapter_cls.name`. Since the CLI now validates argv against `AdapterName`, an adapter can register under `"fake"` successfully and then be undispatchable through `python -m harness.cli fake ...`. We should validate/canonicalize through `AdapterName` in `register()`, reject invalid names here, and add a negative registry test.

plugins/code/tools/python/test_harness_cli.py
```diff
@@ -33,13 +39,24 @@
...
 from harness.types import Command, InvocationRequest, TurnResult, TerminalFailure
 
+# Mock the closed AdapterName enum with a superset that adds a test harness so
+# the fake adapter passes the CLI's adapter-name validation boundary.
+class _MockAdapterName(StrEnum):
+    CLAUDE = "claude"
+    CODEX = "codex"
+    FAKE = "fake"
+
+_htypes.AdapterName = _MockAdapterName
```
shafty023, line 65: This patches the production enum to add `FAKE`, so the green subprocess tests do not prove the real closed boundary accepts `claude`/`codex`. We should register the fake adapter under a real `AdapterName` member and call `_run_cli("claude", ...)`; keep invalid adapter coverage separate from valid-but-unregistered coverage.

plugins/code/tools/python/harness/registry.py
```diff
@@ -55,7 +57,6 @@ def get_adapter(name: str) -> type["HarnessAdapter"]:
     if name not in _REGISTRY:
         available = ", ".join(sorted(_REGISTRY)) or "(none)"
         raise KeyError(
-            f"No adapter registered with name {name!r}. "
-            f"Available adapters: {available}"
+            f"No adapter registered with name {name!r}. Available adapters: {available}"
```
shafty023, line 60: For a valid enum adapter with no registered class, this formats `name!r`, so CLI stderr becomes `<AdapterName.CODEX: 'codex'>` instead of `codex`. Since this path is now user-facing for valid-but-unregistered adapters, can we format with `name.value`/`str(name)` and assert the stderr?

plugins/code/.claude-plugin/plugin.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "code",
   "description": "Code and planning framework plugin",
-  "version": "1.12.4",
+  "version": "1.12.5",
```
shafty023, line 4: This is a patch bump, but the PR describes an additive public harness surface and says the version is `1.13.0`. `CONTRIBUTING.md` says backward-compatible plugin changes should be MINOR, so we should either make this `1.13.0` or update the PR/version rationale if this is intentionally only a patch.

plugins/code/tools/python/test_harness_cli.py
thadeusb, line 65: This monkeypatch means the test at ~line 200 (`test_nonexistent_adapter_...`) now exercises a different error path than it thinks it does. `"nonexistent-adapter"` fails here at `adapter_name_from_str` (ValueError) because it's not in this 3-member mock enum — it never reaches `get_adapter` for the KeyError.

The docstring still says "when the adapter name is not registered" but it's actually testing "when the string isn't a valid enum member."

The real gap: passing `"claude"` (valid mock enum member, no adapter registered) would hit the KeyError path end-to-end and that's now untested through the subprocess.

thadeusb (approved):
Enum tightening and the match/case dispatch are the right shape. One test is now testing a different error path than it thinks it is, see inline.

PR #152, opened 2026-06-09
Plan and Execute implemented as stand alone commands (v1.13.0)

aponamarev:
## Summary

Simplified execution of 'Plan' and 'Execute' commands by moving away form run-loops.sh. This was achieved by splitting the monolithic `prompt.md` into two separate-session commands.

## What changed

**Added**
- `plan.json` top-level `simple_mode` and `plan_was_imported` booleans in `schemas/plan-schema.json` (optional; non-breaking — no `additionalProperties: false`).
- Parametrized tests in `test_validate_plan.py` covering both flags present/absent/non-bool extraction.

**Changed**
- `plan-writer` (FINALIZE MODE) now persists both flags into `plan.json`; `plan-prompt.md` Phase 2.7 instruction updated to require it.
- `validate_plan.py` `extract_data` surfaces `simple_mode` / `plan_was_imported` in its VALID output (default `False`, keeping legacy plans safe).
- `execute-prompt.md` Phase 3.0 recovers both flags from the `plan-validate` output instead of nonexistent planning-session memory; Phase 5.5 skips when either is true.

**Fixed**
- `/code:execute-implementation` no longer hard-stops simple-mode or imported plans at Phase 5.5. Previously both flags defaulted to `false` in the fresh execute session, so plans created without a decision table hit an empty-`decision_table_path` hard stop with a misleading "re-run /code:create-plan" message.
- `commands/execute-implementation.md` frontmatter now lists `SlashCommand` in `allowed-tools`; without it the Phase 6.5 in-session review round could not invoke `/code-review:start` and silently never ran.
- Corrected the planning phase range from `0.9–2.7` to `0.9–2.8` across `create-plan.md`, `plan-prompt.md`, and `README.md` (completion actually occurs at Phase 2.8); added Phase 2.8 to the README workflow diagram.

## Validation
- pytest (`plan-validate`): **34 passed** · ruff clean · pyright 0 errors · `plan-schema.json` valid JSON
- CHANGELOG `### code v1.13.0` heading matches `plugin.json`; no residual `0.9–2.7` references


plugins/code/skills/plan-validate/scripts/validate_plan.py
```diff
@@ -501,6 +501,14 @@ def extract_data(data: dict) -> dict:
     dt_path = decision_table.get("path", "") if isinstance(decision_table, dict) else ""
     dt_status = decision_table.get("status", "") if isinstance(decision_table, dict) else ""
 
+    # Mode flags persisted by plan-writer at finalization. The execute-implementation
+    # session runs in a fresh context with no planning-session working memory, so it
+    # recovers these from plan.json via this output to decide whether to skip Phase 5.5
+    # (behavioral verification). Default False keeps legacy plans (written before these
+    # fields existed) safe: they fall through to running Phase 5.5 rather than skipping.
+    simple_mode = bool(data.get("simple_mode", False))
```
thadeusb: `bool()` is too loose here. If an LLM writes `"simple_mode": "false"` (a JSON string instead of a boolean), `bool("false")` evaluates to `True` and Phase 5.5 gets silently skipped when it shouldn't be. Same for `plan_was_imported` on the next line.

Use `data.get("simple_mode") is True` or `data.get("simple_mode", False) is True` so only the actual boolean `True` passes.

Your test case for int coercion (`1 -> True`, `0 -> False`) would need to change too, but that's the right tradeoff: ints round-tripping through JSON as booleans isn't a real scenario, strings are.

plugins/code/commands/execute-implementation.md
```diff
@@ -0,0 +1,19 @@
...
+
+Follow the orchestrator instructions in the prompt file specified by `CLOSEDLOOP_PROMPT_FILE` in the config output above. There is NO external loop here — this is a single-shot, in-session implementation run. Your previous work (if any) is visible in files and git history.
+
+IMPORTANT: You are an ORCHESTRATOR running IMPLEMENTATION ONLY (phases 3–7 plus in-session review rounds). After reading the prompt file, your FIRST action must be TodoWrite. Do NOT read project files (PRD, plan.json, code, etc.). Delegate all project file reading to subagents.
+
+This command requires an existing `plan.json`. If none exists, the orchestrator HARD STOPS and points you to `/code:create-plan` — it never drafts a plan itself.
+
+After Phase 7 completes, write `state.json` with `status: COMPLETED`, output `<promise>IMPLEMENTATION_COMPLETE</promise>`, and stop. If tasks remain pending or validation fails, end naturally WITHOUT the promise so a re-invocation resumes the remaining work.
+
+CRITICAL RULE: Only output `<promise>IMPLEMENTATION_COMPLETE</promise>` when implementation is genuinely finished (all tasks complete, build validation passes). Do not output false promises to end early.
```
thadeusb, line 19: This CRITICAL RULE says "only output the promise when implementation is genuinely finished", but the `AWAITING_USER_SEQUENCE` in `execute-prompt.md` explicitly emits `<promise>IMPLEMENTATION_COMPLETE</promise>` on hard stops like missing `plan.json` or missing decision table.

Same contradiction exists between `create-plan.md` and `plan-prompt.md`.

The LLM gets two authoritative instructions that directly conflict. If it follows the command file, it won't emit the promise on `AWAITING_USER` and might keep running instead of stopping. Either soften the CRITICAL RULE to acknowledge `AWAITING_USER` as a valid exit, or change the `AWAITING_USER_SEQUENCE` to stop without the promise.

thadeusb (approved):
The split is clean and the mode-flag plumbing through plan-writer, validate_plan, and the two new prompts all lines up. Two things inline, one of which is a real coercion bug.

plugins/code/prompts/execute-prompt.md
```diff
@@ -0,0 +1,283 @@
...
+2. If `FORMAT_ISSUES`: launch @code:plan-writer to fix format issues, then re-activate `code:plan-validate`
+3. If `VALID`: launch @code:plan-validator with prompt: "WORKDIR=$CLOSEDLOOP_WORKDIR. SEMANTIC ONLY: Check semantic consistency of $CLOSEDLOOP_WORKDIR/plan.json — verify storage/query alignment and task/architecture decision consistency. Skip structural validation (already passed)."
+4. If semantic check finds issues: launch @code:plan-writer to fix, then re-activate `code:plan-validate`
+
+### AWAITING_USER_SEQUENCE
+
+Use this sequence at any hard-stop that requires user action before continuing:
+1. **FIRST** — Write state.json with AWAITING_USER status:
+   `echo '{"phase": "<current phase>", "status": "AWAITING_USER", "reason": "<why>", "userAction": {"description": "<what user should do>", "file": "<path or null>", "command": "<resume command>"}, "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json`
+2. **ONLY AFTER state.json is written** — Output `<promise>IMPLEMENTATION_COMPLETE</promise>`
```
shafty023: This hard-stop helper writes `AWAITING_USER` and then emits the terminal success marker. The no-plan, malformed-plan, validation-exhausted, and decision-table failure paths all route through here, so `/code:execute-implementation` can look complete even though implementation never finished. We should reserve `<promise>IMPLEMENTATION_COMPLETE</promise>` for the Phase 7 success path and end hard-stops without the completion promise, or use a distinct non-success marker.

plugins/code/prompts/plan-prompt.md
```diff
@@ -0,0 +1,232 @@
...
+2. If `FORMAT_ISSUES`: launch @code:plan-writer to fix format issues, then re-activate `code:plan-validate`
+3. If `VALID`: launch @code:plan-validator with prompt: "WORKDIR=$CLOSEDLOOP_WORKDIR. SEMANTIC ONLY: Check semantic consistency of $CLOSEDLOOP_WORKDIR/plan.json — verify storage/query alignment and task/architecture decision consistency. Skip structural validation (already passed)."
+4. If semantic check finds issues: launch @code:plan-writer to fix, then re-activate `code:plan-validate`
+
+### AWAITING_USER_SEQUENCE
+
+Use this sequence at any hard-stop that requires user action before continuing:
+1. **FIRST** — Write state.json with AWAITING_USER status:
+   `echo '{"phase": "<current phase>", "status": "AWAITING_USER", "reason": "<why>", "userAction": {"description": "<what user should do>", "file": "<path or null>", "command": "<resume command>"}, "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json`
+2. **ONLY AFTER state.json is written** — Output `<promise>PLAN_COMPLETE</promise>`
```
shafty023: Same false-success issue on the plan side: `AWAITING_USER_SEQUENCE` emits `<promise>PLAN_COMPLETE</promise>`, and the Phase 1.1 “Stop” path also marks the plan complete before Phase 2.8 finalization. That can make a draft or blocked finalization look like a validated plan. `PLAN_COMPLETE` should only come from the actual Phase 2.8 completion path; hard-stops/review stops need a non-completion state.

plugins/code/prompts/execute-prompt.md
```diff
@@ -0,0 +1,283 @@
...
+```
+If `CHANGED` is `0`, log "No code changes — skipping review rounds" and proceed to Phase 7.
+
+**Review loop (while `cycle <= max_cycles`):**
+1. Update state.json with `"reviewCycle": cycle`.
+2. **Run review:** Run the `/code-review:start --base $startSha` command via the SlashCommand tool (`code-review:start` is a slash command, not a `Skill`). (This writes a fresh session directory under `$CLOSEDLOOP_WORKDIR/.closedloop-ai/code-review/cr-*`.)
+3. **Read the verdict** via Bash (no file reads into orchestrator context beyond these scalars):
+   ```bash
+   CR_DIR=$(ls -td "$CLOSEDLOOP_WORKDIR"/.closedloop-ai/code-review/cr-* 2>/dev/null | head -1)
+   VERDICT=$(jq -r '.verdict // "approve"' "$CR_DIR/verdict.json" 2>/dev/null || echo "approve")
```
shafty023: This fallback turns a missing `CR_DIR`, missing `verdict.json`, `jq` failure, or `/code-review:start` failure into `approve`, and the next line treats empty `CR_DIR` as “Code review passed.” That silently drops the new in-session review gate. We should require a fresh review directory and readable verdict from this invocation; if either is missing, retry or hard-stop instead of approving.

plugins/code/skills/plan-validate/scripts/validate_plan.py
shafty023: `bool(...)` makes string values like `"false"` truthy, and `validate_schema_fields()` does not reject non-boolean values for these new fields. Since execute now uses these values as the source of truth for skipping Phase 5.5, a malformed/stringy plan can bypass behavioral verification. We should validate both fields as actual booleans when present and extract with the stored boolean value, defaulting to false only when absent; the new test case that blesses non-bool coercion should flip to negative coverage.

plugins/code/prompts/plan-prompt.md
```diff
@@ -0,0 +1,232 @@
...
+
+**PHASE 2.8: PLAN COMPLETION**
+
+- Confirm the plan is valid: activate `code:plan-validate` skill once more; if it does not return `VALID`, fix via @code:plan-writer and re-validate before completing.
+- Append a short summary of the plan to `$CLOSEDLOOP_WORKDIR/log.md` via a haiku subagent (description: `"plan-editor"`) — do NOT read or write files yourself.
+- Write state.json **first**, then the promise:
+  `echo '{"phase": "Phase 2.8: Plan completion", "status": "COMPLETED", "planStatus": "PLAN_COMPLETE", "startSha": "'$START_SHA'", "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json`
+- Run the telemetry line: `bash "$CLAUDE_PLUGIN_ROOT/scripts/record_phase.sh" 2>/dev/null || true`
+- **ONLY AFTER state.json is written** — output `<promise>PLAN_COMPLETE</promise>`
+- Tell the user: the plan is ready at `$CLOSEDLOOP_WORKDIR/plan.md`. Run `/code:code` (or resume via the loop) to implement it.
```
shafty023: This completion handoff points users back to the legacy combined `/code:code` command, even though this PR introduces `/code:execute-implementation` as the standalone consumer for the finalized plan. That can route users back into the old full-loop path instead of cleanly executing the plan they just created. The primary next command here should be `/code:execute-implementation` with the compatible args.

plugins/code/scripts/setup-closedloop.sh
```diff
@@ -150,6 +150,18 @@ while [[ $ARG_INDEX -lt $ARG_COUNT ]]; do
             MAX_ITERATIONS="${ARGS[$ARG_INDEX]}"
             ARG_INDEX=$((ARG_INDEX + 1))
             ;;
+        --review-cycles)
+            ARG_INDEX=$((ARG_INDEX + 1))
+            if [[ $ARG_INDEX -ge $ARG_COUNT ]]; then
```
shafty023, line 177: This only checks that another arg exists. With the real command shape, `/code:execute-implementation --review-cycles` becomes `--review-cycles --prompt execute-prompt`; the parser consumes `--prompt` as the value and then treats `execute-prompt` as the workdir, producing a misleading path error instead of “requires a value.” Value-consuming options should reject the next token when it starts with `-` (and ideally validate/cap the cycle count) with a targeted parser test.

plugins/code/skills/plan-validate/scripts/validate_plan.py
aponamarev: Changed extraction to strict identity comparison at lines 513–514: `data.get("simple_mode") is True` and `data.get("plan_was_imported") is True`. Added boolean type enforcement in `validate_schema_fields()` at lines 327–329 to reject non-boolean values. Commit 24ea6eb.

plugins/code/commands/execute-implementation.md
aponamarev: The CRITICAL RULE in execute-implementation.md is satisfied. The execute-prompt.md at HEAD delegates AWAITING_USER_SEQUENCE to the `code:orchestrator-sequences` skill (line 42), which documents that completion promise emission on hard-stops is by design. The contradiction visible in the diff hunk is not present in current code. Resolving.

plugins/code/prompts/execute-prompt.md
aponamarev: This thread is marked is_outdated:true on the PR. The inline AWAITING_USER_SEQUENCE the reviewer cited has been replaced with a delegation to `code:orchestrator-sequences` skill (line 42 in current execute-prompt.md). The skill documents hard-stop promise emission as intentional design. Resolving.

plugins/code/prompts/plan-prompt.md
aponamarev: This thread is marked is_outdated:true on the PR. The plan-prompt.md at HEAD also delegates to `code:orchestrator-sequences` skill instead of inlining AWAITING_USER_SEQUENCE. The design decision to emit completion promise on hard-stops is documented in SKILL.md:28 as intentional. Resolving.

plugins/code/prompts/execute-prompt.md
aponamarev: Replaced the silent `approve` fallback with an explicit `__review_failed__` sentinel at lines 219–223. When `/code-review:start` produces no output or `verdict.json` is missing, the code now routes to `consecutive_failures` escalation instead of masquerading as a pass. Step 4 (line 225) branches on the sentinel. Commit 24ea6eb.

plugins/code/skills/plan-validate/scripts/validate_plan.py
aponamarev: Applied both fixes for this thread: strict identity extraction at lines 513–514 and schema-level type enforcement at lines 327–329. A plan with non-boolean values (e.g., `"simple_mode": "false"`) is now rejected during validation before reaching extract time. Commit 24ea6eb.

plugins/code/prompts/plan-prompt.md
aponamarev: At HEAD, plan-prompt.md:667 directs users to `/code:execute-implementation` as the primary path for single-shot implementation. The legacy `/code:code` is mentioned only as the external-loop alternative. The earlier draft visible in the diff hunk has been updated. Resolving.

plugins/code/scripts/setup-closedloop.sh
aponamarev: Added a guard at line 181: `[[ "${ARGS[$ARG_INDEX]}" == -* ]]` to reject flag-like values before consuming the `--review-cycles` value. Misparses like `--review-cycles --prompt` now exit with an error. Commit 24ea6eb.

PR #153, opened 2026-06-10
feat(code): native-command perf telemetry via UserPromptExpansion hook

aponamarev:
## Summary

Give the native in-session `/code:create-plan` and `/code:execute-implementation` commands the same perf/telemetry trail the external `run-loop.sh` already produces. Previously only `run-loop.sh` wrote `run` and `iteration` events to `perf.jsonl`; the single-shot commands left no iteration record, so native runs were invisible to downstream telemetry.

## What changed

**Added**
- `hooks/user-prompt-expansion-hook.sh` + a new `UserPromptExpansion` hook in `hooks.json` matching `create-plan|execute-implementation`. The hook seeds native PLAN/EXECUTE perf metadata: it resolves the run's `config.env` from candidate sources — `CLOSEDLOOP_WORKDIR`, input-JSON workdir fields (`workdir`/`working_directory`/`target_workdir`/`command.*`), the command's first positional argument, then `cwd` — writes `CLOSEDLOOP_RUN_ID`, `CLOSEDLOOP_ITERATION=0`, and `CLOSEDLOOP_COMMAND` (`PLAN`/`EXECUTE`) into both `config.env` and the environment, and records the run-start event via `record_run.sh`. Fails open on every error so prompt expansion never blocks the command.
- `scripts/record_iteration.sh` — appends one synthetic native-command `iteration` event to `perf.jsonl`. Derives `status`/`claude_exit_code` (`ok`/0 vs `error`/1) from whether `state.json` reached `COMPLETED`, and computes `duration_s` from the matching `run` event's `started_at`.
- `scripts/closedloop_env.sh` — shared `load_closedloop_env` helper that hydrates `CLOSEDLOOP_RUN_ID`/`CLOSEDLOOP_ITERATION`/`CLOSEDLOOP_COMMAND` from `config.env` while letting non-empty environment values win (env is source of truth, `config.env` is fallback). Single source of truth for both `record_iteration.sh` and `record_phase.sh`; no-op when the config file is absent.

**Changed**
- `scripts/record_phase.sh` now sources `closedloop_env.sh` and hydrates env vars from `config.env` before recording, so native phase telemetry picks up the metadata the expansion hook seeded.
- `prompts/plan-prompt.md` (Phase 2.8 completion) runs `record_iteration.sh` after `record_phase.sh`, so a completed plan-only session emits an iteration event.
- `prompts/execute-prompt.md` (completion) runs `record_iteration.sh` before emitting `<promise>IMPLEMENTATION_COMPLETE</promise>`, so a completed implementation session emits an iteration event.
- `plugin.json` — `1.13.1` → `1.13.2` (PATCH; additive hook/scripts, no breaking change to existing contracts).

## Validation
- pytest (`test_user_prompt_expansion_hook.py`, `test_record_iteration.py`, `test_record_phase.py`): **18 passed**
- `uv run ruff check` clean · `uv run pyright` 0 errors
- `plugin.json` + `hooks.json` valid JSON
- CHANGELOG `### code v1.13.2` heading matches `plugin.json`


plugins/code/scripts/record_iteration.sh
```diff
@@ -0,0 +1,75 @@
...
+CONFIG_FILE="$WORKDIR/.closedloop-ai/config.env"
+load_closedloop_env "$CONFIG_FILE"
+
+PERF_FILE="$WORKDIR/perf.jsonl"
+STATE_FILE="$WORKDIR/state.json"
+
+RUN_ID="${CLOSEDLOOP_RUN_ID:-unknown}"
+ITERATION="${CLOSEDLOOP_ITERATION:-0}"
+COMMAND="${CLOSEDLOOP_COMMAND:-interactive}"
+STATUS="ok"
```
thadeusb, line 29: Default is backwards. `STATUS` starts at `"ok"`, and only flips to `"error"` inside the `if [[ -f "$STATE_FILE" ]]` block. So if `state.json` is missing entirely (command crashed before writing it), you record a successful iteration. Flip the default to `STATUS="error"` / `CLAUDE_EXIT_CODE=1` and set `"ok"`/`0` only when `STATE_STATUS == COMPLETED`.

plugins/code/tools/python/test_record_iteration.py
```diff
@@ -0,0 +1,95 @@
...
+        for line in (tmp_path / "perf.jsonl").read_text().splitlines()
+        if line.strip()
+    ]
+    assert len(records) == 1
+    event = records[0]
+    assert event["event"] == "iteration"
+    assert event["run_id"] == "run-from-config"
+    assert event["iteration"] == 3
+    assert event["command"] == "EXECUTE"
+    assert event["status"] in {"ok", "success", "failure", "completed", "COMPLETED"}
```
thadeusb: This accepts 5 values but the code only ever produces `"ok"` or `"error"`. And this test doesn't create a `state.json`, so it's exercising the missing-file path from the bug above. Once the default is fixed, this should assert `status == "error"` (no `state.json` = not completed).

thadeusb (approved):
Telemetry plumbing is solid, closedloop_env.sh extraction is the right call. One inverted default in record_iteration.sh to fix.

shafty023 (commented):
Please address the inline findings before merge.

plugins/code/hooks/user-prompt-expansion-hook.sh
```diff
@@ -0,0 +1,95 @@
...
+# Fails open on all errors so prompt expansion never blocks the user command.
+
+trap 'exit 0' ERR
+
+CLOSEDLOOP_STATE_DIR=".closedloop-ai"
+
+INPUT=$(cat)
+COMMAND_NAME=$(echo "$INPUT" | jq -r '.command_name // empty')
+CWD=$(echo "$INPUT" | jq -r '.cwd // empty')
+COMMAND_ARGS=$(echo "$INPUT" | jq -r '
```
shafty023, line 16: This reads `command_arguments`, `arguments`, and `args`, but the documented `UserPromptExpansion` payload uses `command_args`. That means a normal `/code:create-plan /some/workdir` or `/code:execute-implementation /some/workdir` launched from a different cwd can miss the target workdir/config and skip the native run envelope entirely. Can we parse `command_args` first and add a regression test with the real hook payload shape and `cwd != target workdir`?

plugins/code/hooks/user-prompt-expansion-hook.sh
```diff
@@ -0,0 +1,95 @@
...
+  try_config_candidate "${COMMAND_ARGS%% --*}"
+fi
+try_config_candidate "$CWD"
+
+if [[ -z "$CONFIG_FILE" ]]; then
+  exit 0
+fi
+
+# shellcheck disable=SC1090
+source "$CONFIG_FILE"
```
shafty023: This executes a workspace-discovered `config.env` as shell before the command bootstrap rewrites it. Since the hook resolves that file from command args/cwd, a repo can put shell in `.closedloop-ai/config.env` and have it run when the user invokes the native commands. We should treat this file as data here and in `closedloop_env.sh`: parse only the allowlisted `CLOSEDLOOP_*` keys with a strict assignment parser, and add an abuse test for values like `CLOSEDLOOP_RUN_ID=$(...)`.

plugins/code/hooks/user-prompt-expansion-hook.sh
```diff
@@ -0,0 +1,95 @@
...
+grep -v -E '^(CLOSEDLOOP_RUN_ID|CLOSEDLOOP_ITERATION|CLOSEDLOOP_COMMAND)=' "$CONFIG_FILE" > "$CONFIG_TMP" 2>/dev/null || true
+{
+  cat "$CONFIG_TMP"
+  printf 'CLOSEDLOOP_RUN_ID=%q\n' "$RUN_ID"
+  printf 'CLOSEDLOOP_ITERATION=0\n'
+  printf 'CLOSEDLOOP_COMMAND=%q\n' "$CLOSEDLOOP_COMMAND_VALUE"
+} > "$CONFIG_FILE"
+rm -f "$CONFIG_TMP" 2>/dev/null || true
+
+export CLOSEDLOOP_WORKDIR="$WORKDIR"
```
shafty023, line 93: These exports only affect this hook subprocess. Later `PreToolUse`/`PostToolUse` hook processes do not inherit them, and `pre-tool-use-hook.sh` still writes tool/spawn sentinels from ambient env only. So native PLAN/EXECUTE can get `phase`/`iteration` rows from `config.env` while `tool`, `skill`, and `spawn` rows stay blank or `unknown`, which misses the same-run attribution requirement. Can we hydrate the persisted run metadata in the tool telemetry producer path and add a boundary test for expansion hook -> PreToolUse/PostToolUse with no ambient `CLOSEDLOOP_*` env?

plugins/code/prompts/plan-prompt.md
```diff
@@ -206,6 +206,7 @@ This is a single-shot interactive command, so review happens in-session via `Ask
 - Write state.json **first**, then the promise:
   `echo '{"phase": "Phase 2.8: Plan completion", "status": "COMPLETED", "planStatus": "PLAN_COMPLETE", "startSha": "'$START_SHA'", "timestamp": "'$(date -u +%Y-%m-%dT%H:%M:%SZ)'"}' > $CLOSEDLOOP_WORKDIR/state.json`
 - Run the telemetry line: `bash "$CLAUDE_PLUGIN_ROOT/scripts/record_phase.sh" 2>/dev/null || true`
+- Record the native-command iteration: `bash "$CLAUDE_PLUGIN_ROOT/scripts/record_iteration.sh" 2>/dev/null || true`
```
shafty023: This records an iteration only on the Phase 2.8 happy path. Other native terminal paths still skip it: the plan review `Stop` option writes `status: COMPLETED` and emits `PLAN_COMPLETE`, and execute can end on final validation failure, pending tasks, or other hard stops without an iteration row. The requirement is one synthetic iteration per native invocation, so we need a once-per-terminal-path owner, for example a stop hook with idempotency or explicit calls on every terminal branch, plus tests for at least one non-happy path.

plugins/code/scripts/record_iteration.sh
```diff
@@ -0,0 +1,75 @@
...
+ENDED_AT=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+END_EPOCH=$(date +%s)
+STARTED_AT="$ENDED_AT"
+START_EPOCH="$END_EPOCH"
+
+if [[ -f "$PERF_FILE" ]]; then
+  RUN_STARTED_AT=$(jq -r --arg run_id "$RUN_ID" 'select(.event == "run" and .run_id == $run_id) | .started_at // empty' "$PERF_FILE" 2>/dev/null | head -n1 || true)
+  if [[ -n "$RUN_STARTED_AT" ]]; then
+    STARTED_AT="$RUN_STARTED_AT"
+    if date -j -f "%Y-%m-%dT%H:%M:%SZ" "$RUN_STARTED_AT" +%s >/dev/null 2>&1; then
```
shafty023: `record_run.sh` writes `started_at` as UTC (`...Z`), but this macOS parse path omits `-u`, so BSD `date` treats the timestamp fields as local time. In US time zones short runs get a future `START_EPOCH`, then `duration_s` is clamped to `0`; longer runs are off by the local UTC offset. Can we parse with `date -j -u -f ...` here, matching the existing pattern in `post-tool-use-hook.sh`?

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
+### code v1.13.2
```
shafty023, line 23: Repo guardrails say not to manually edit `CHANGELOG.md`/README for plugin changes and to run `/update-documentation` after the code changes are final. This PR changes both docs files, but the validation section does not include that workflow. Can we run `/update-documentation` or add the exact evidence if it already ran?

plugins/code/scripts/record_iteration.sh
aponamarev: Thanks. This path is already addressed in the current branch: a missing `state.json` now records the native iteration as `status: "error"` with `claude_exit_code: 1`, so a crash-before-state-write is no longer reported as success.

plugins/code/tools/python/test_record_iteration.py
aponamarev: Thanks. This is already covered in the current branch: the schema-compatible success test now asserts `status == "ok"` and `claude_exit_code == 0`, and there is a dedicated missing-`state.json` regression asserting `error` / `1`.

plugins/code/hooks/user-prompt-expansion-hook.sh
aponamarev: Fixed in `41f54b4`: `user-prompt-expansion-hook.sh` now prefers the documented `command_args` field before the legacy argument fields, and the regression test covers `command_args` with `cwd != target workdir` and no ambient `CLOSEDLOOP_*` env.

plugins/code/hooks/user-prompt-expansion-hook.sh
aponamarev: Fixed in `41f54b4`: the expansion hook no longer sources workspace `config.env`; it loads the shared safe parser from `closedloop_env.sh` and treats allowlisted `CLOSEDLOOP_*` values as data. Added a regression test with command-substitution-looking config content to prove it is not executed.

plugins/code/hooks/user-prompt-expansion-hook.sh
aponamarev: Fixed in `41f54b4`: `pre-tool-use-hook.sh` now hydrates persisted native run metadata from `.closedloop-ai/config.env` after resolving the workdir and before writing sentinels or Agent spawn events. New telemetry-hook tests cover both sentinel and spawn attribution with no ambient run env.

plugins/code/prompts/plan-prompt.md
aponamarev: Fixed in `41f54b4`: added `record_native_iteration_once.sh` and updated `/code:code`, `/code:create-plan`, `/code:execute-implementation`, the plan Stop path, and shared hard-stop handoffs to record native terminal iterations after terminal `state.json` writes without duplicate rows.

plugins/code/scripts/record_iteration.sh
aponamarev: Fixed in `41f54b4`: the macOS parse branch now uses `date -j -u -f "%Y-%m-%dT%H:%M:%SZ" ...`, matching the UTC timestamp semantics and the existing post-tool hook pattern.

CHANGELOG.md
aponamarev: Fixed in `41f54b4`: followed the repo's local `/update-documentation` workflow instructions after code changes, bumped the `code` plugin to `1.13.3`, added a fresh root `CHANGELOG.md` entry, and minimally updated `plugins/code/README.md` for the new telemetry behavior.

PR #154, opened 2026-06-11
feat(code): Claude Design handoff pipeline (design-inventory)

shafty023:
## Summary

Implements FEA-1739 / PLN-859: a staged pipeline from a Claude Design export zip to DRAFT feature tickets in ClosedLoop, with a hard human review gate in the middle. Designs frequently contain vibe-coded changes the designer never intended to ship; nothing in a design is implemented by default.

- **Stage A (inventory):** deterministic export decomposition into typed design units (screens, chrome regions like nav bars, standalone components like a chat dialog), interaction-signal extraction (scroll sync, hover, drag and drop, pointer-drag scrubbing), repo-side route map and Storybook component index, per-unit token-resolved visual specs, then parallel `design-unit-analyst` agents emitting schema-validated `findings.json`. Rendered outputs: a markdown report and a self-contained HTML review page.
- **Stage B (review gate):** themes decided first with per-finding overrides, likely-intentional findings pre-accepted under a veto model; the page exports `decisions.json`. No tickets exist before it does.
- **Stage C (tickets):** for accepted units only, design packs (sliced design source, reference screenshots, decision-applied findings, visual spec) plus ticket bodies with acceptance criteria, an explicit declined-changes do-not-implement list, a component reuse table with exact imports, and dependency callouts for net-new components and backend gaps.

All scripts are strict TypeScript (team policy: no new Python) bundled to committed `dist/*.mjs`, so consumers need only Node 18+. CI gains a job running tsc, vitest, and a stale-dist guard. CLAUDE.md records the policy.

## Validation

- 134 vitest tests (ported test-for-test from the original suites), tsc clean, repo pytest/ruff/pyright untouched and green.
- Exercised end to end against a real 23 MB export: 34 typed units detected, 52 routes and 5 chrome layouts mapped, 173 components indexed (85 enriched with props/variants), analyst findings validated, review page generated, sample decisions exercised the gate, and a full design pack with an 11.4 KB ticket body produced.
- Cross-implementation parity during the TS port: manifests, signals, visual specs, and ticket bodies byte-identical to the original implementation's output.

## Notes for reviewers

- ClosedLoop refs: FEA-1739 (feature), PLN-859 (plan, executed).
- Stage C ticket creation via MCP is specified in the skill but was deliberately not exercised against the org; the first real run should follow an actual design review, not sample decisions.
- First TS toolchain in this repo: `plugins/code/skills/design-inventory/scripts` (npm ci / npm test / npm run typecheck / npm run build).

## Repo policy changes (explicit callout)

- **Version bumps: once per branch, not once per commit.** CLAUDE.md previously read as requiring a bump in every commit touching a plugin; this PR rewrites it to the actual policy (bump exactly once relative to main; CI already enforced base-vs-HEAD). This branch follows it: single bump 1.13.3 -> 1.14.0.
- **Plugin payload rule (new).** Plugins ship only runtime payload; dev toolchains (package.json, lockfiles, node_modules, TS sources, tests) must live outside the plugin tree because the installer npm-installs and copies everything it finds. The design-inventory toolchain moved to tools/design-inventory accordingly; installed size dropped from ~78MB to ~250KB.
- **TypeScript-only for new tool scripts** (replaces the Python tools convention for new work; existing Python remains until ported).


.github/workflows/ci.yml
```diff
@@ -100,3 +100,27 @@ jobs:
...
+          node-version: "22"
+          cache: npm
+          cache-dependency-path: plugins/code/skills/design-inventory/scripts/package-lock.json
+      - run: npm ci
+      - run: npm run typecheck
+      - run: npm test
+      - name: Verify committed dist bundles are up to date
+        run: |
+          npm run build
+          if ! git diff --exit-code -- dist; then
```
thadeusb: The bundles this guard is supposed to protect don't exist: there are zero dist/*.mjs files in this PR, because root .gitignore line 2 has an unanchored dist/ that swallows scripts/dist. And this guard can't catch that, git diff only reports tracked files, so after npm run build the fresh output is untracked plus ignored and the diff is empty. I verified in a worktree: ran npm run build, this step exits 0, git status shows nothing. So CI stays green while anyone installing the plugin gets ERR_MODULE_NOT_FOUND on every command in SKILL.md, and the CHANGELOG claim that dist is committed is false. Three fixes needed: a !plugins/code/skills/design-inventory/scripts/dist/ exception in root .gitignore, the bundles actually committed, and this step hardened to fail when dist is missing or untracked, e.g. git ls-files --error-unmatch dist/*.mjs before the diff check.

plugins/code/skills/design-inventory/scripts/src/render-review-html.ts
```diff
@@ -0,0 +1,797 @@
...
+        remainingBudget -= uris.length;
+      }
+    }
+
+    const themes = Array.isArray(doc["themes"]) ? (doc["themes"] as unknown[]) : [];
+    for (const theme of themes) {
+      if (typeof theme === "object" && theme !== null && !Array.isArray(theme)) {
+        const themeObj = theme as JsonObject;
+        allThemes.push([themeObj, unitName]);
+        themeToFindings.set(String(themeObj["id"] ?? ""), []);
```
thadeusb: This set() wipes member findings when two units emit the same theme id. The map is keyed by bare theme id across all docs, so doc2's set() replaces the array doc1's findings were already pushed into. I repro'd with two units both emitting thm-artifact-table, which is the exact example theme in the design-unit-analyst prompt: unit 1's finding appears nowhere on the page, not under its theme and not standalone, and both theme cards render unit 2's members. The radios collide too, 8 inputs share name=thm-artifact-table so the two theme cards behave as one linked radio group and export a single decision key covering both units. Findings silently vanishing from the human review gate is the one failure mode this stage exists to prevent. Key the map and radio names by unitId plus themeId, or hard-fail on cross-doc duplicate theme ids.

tools/design-inventory/src/design-export-extract.ts
```diff
@@ -0,0 +1,719 @@
...
+export function safeExtract(zipPath: string, extractDir: string): string[] {
+  const zipData = readFileSync(zipPath);
+  const entries = unzipSync(zipData);
+
+  // Compute total uncompressed size
+  let total = 0;
+  for (const [, data] of Object.entries(entries)) {
+    total += data.byteLength;
+  }
+  if (total > MAX_TOTAL_UNCOMPRESSED) {
```
thadeusb: This guard runs after unzipSync already inflated every entry into memory, so a small zip that expands to 30GB OOMs the process before the check fires. fflate's unzip filter callback gets originalSize per entry, sum there and bail before decompressing.

tools/design-inventory/src/capture-design-shots.ts
```diff
@@ -0,0 +1,400 @@
...
+    "Playwright not found. Install it in the target repo (npm i -D playwright && " +
+      "npx playwright install chromium) or run from a repo that already has @playwright/test.",
+  );
+}
+
+function serveDir(root: string): Promise<{ server: Server; port: number }> {
+  return new Promise((resolvePromise) => {
+    const server = createServer((req, res) => {
+      const urlPath = decodeURIComponent((req.url ?? "/").split("?")[0] ?? "/");
+      const file = join(root, urlPath === "/" ? "index.html" : urlPath.slice(1));
```
thadeusb: join(root, urlPath.slice(1)) doesn't contain the path, a raw GET /../../../../etc/passwd from any local client reads outside the extract dir. You were careful about zip-slip in the extractor, the same containment check belongs here.

tools/design-inventory/src/build-route-map.ts
```diff
@@ -0,0 +1,439 @@
...
+    return null;
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Filesystem walk (used for pages-router glob)
+// ---------------------------------------------------------------------------
+
+/** Recursively collect all files under dir. */
+function walkFiles(dir: string): string[] {
```
thadeusb: This walker recurses node_modules, .git, .next, everything, and only filters per file afterwards via isExcluded. build-component-index.ts has a walkFiles that prunes at walk time, and its header comment explains exactly why that matters in monorepos. Against a repo like symphony-alpha this version stats every file in every node_modules tree. There are now three walkers in this directory (this one, component-index's, rglob in extract-visual-spec), they should be one shared pruned implementation.

tools/design-inventory/src/build-route-map.ts
```diff
@@ -0,0 +1,439 @@
...
+
+// ---------------------------------------------------------------------------
+// Core builder
+// ---------------------------------------------------------------------------
+
+/** Read a source file as UTF-8, replacing errors. */
+function readText(filePath: string): string {
+  const buf = readFileSync(filePath);
+  // Node doesn't have a direct "replace errors" option in readFileSync,
+  // but latin1 -> utf8 conversion trick gives best-effort decoding.
```
thadeusb: this comment describes a latin1 decode trick the code doesn't do, it's a plain utf8 toString. Do the fallback or drop the comment.

tools/design-inventory/src/extract-visual-spec.ts
```diff
@@ -0,0 +1,547 @@
...
+        }
+      }
+    }
+  }
+  return colors;
+}
+
+function rglob(dir: string, pattern: RegExp, excludeParts: Set<string>): string[] {
+  const results: string[] = [];
+  const entries = readdirSync(dir);
```
thadeusb: no try/catch in rglob, one dangling symlink or unreadable dir anywhere in the repo throws and takes down the whole spec extraction. It also recurses .git since EXCLUDED_PARTS doesn't include it.

tools/design-inventory/src/build-design-pack.ts
```diff
@@ -0,0 +1,412 @@
...
+  const shotsDir = join(packDir, "screenshots");
+  mkdirSync(sourceDir, { recursive: true });
+
+  const designSources = Array.isArray(unit["design_sources"])
+    ? (unit["design_sources"] as string[])
+    : [];
+  for (const rel of designSources) {
+    const src = join(extractDirPath, rel);
+    if (existsSync(src) && statSync(src).isFile()) {
+      copyFileSync(src, join(sourceDir, basename(rel)));
```
thadeusb: basename collision: design_sources like ui_kits/app/index.jsx and ui_kits/admin/index.jsx both land as index.jsx and the second copy clobbers the first. Same in the screenshots loop below.

tools/design-inventory/src/build-design-pack.ts
```diff
@@ -0,0 +1,412 @@
...
+    return 1;
+  }
+
+  const decisions = (decisionsDoc as JsonObject)["decisions"] as Record<string, JsonObject>;
+  const findingsArr = Array.isArray((doc as JsonObject)["findings"])
+    ? ((doc as JsonObject)["findings"] as JsonObject[])
+    : [];
+
+  const accepted = findingsArr.filter((f) => ACCEPTED_STATES.has(effectiveDecision(f, decisions)));
+  const declined = findingsArr.filter((f) => effectiveDecision(f, decisions) === "declined");
```
thadeusb, line 442: what about findings still pending after review? They land in neither accepted nor declined, so they fall out of the ticket entirely, not in the acceptance criteria and not in the do-not-implement list, with no warning. At least print a count so an undecided finding doesn't silently drop out of scope.

CLAUDE.md
```diff
@@ -72,7 +83,7 @@ All commits MUST follow the conventions in `CONTRIBUTING.md`. Specifically:
 
 ### Version Bumps (Required)
 
-**Any change to a plugin's files MUST include a version bump in that plugin's `plugin.json`.** This is not optional -- the version must be updated in the same commit as the code change. Semver rules: PATCH for bug fixes/prompt wording, MINOR for new agents/skills/commands, MAJOR for breaking changes to orchestration/hook API/skill interfaces.
+**Any branch that changes a plugin's files MUST bump that plugin's `plugin.json` version exactly ONCE relative to `main`.** Bump once per branch/PR, not once per commit: pick the right semver level for the branch's overall change and keep it for the life of the branch (CI compares base vs HEAD). Semver rules: PATCH for bug fixes/prompt wording, MINOR for new agents/skills/commands, MAJOR for breaking changes to orchestration/hook API/skill interfaces.
```
thadeusb, line 88: This rewrites the version-bump convention from once-per-commit to once-per-branch for the whole repo. It does match what the CI job actually enforces, but it's a team-wide policy edit buried in a 9k-line feature PR, call it out in the description so it doesn't land unseen.

thadeusb (approved):
The pipeline itself is well built, but the dist/*.mjs bundles everything in SKILL.md runs are not actually in this PR, root .gitignore eats them and the CI guard can't see untracked files, so it ships broken with green CI. Also found a review-page bug that silently drops findings when two units reuse a theme id. Both fixes are mechanical, details inline.

.github/workflows/ci.yml
shafty023: Confirmed independently and fixed before this review landed, with your hardening added on top. Root cause was exactly the unanchored dist/ on root .gitignore line 2; rather than a root exception, scripts/.gitignore now re-includes the directory (!dist/) and all bundles are committed (c576848, currently 16 tracked .mjs). The guard was extended in the same commit to fail on untracked dist files, and c7bb46f adds your missing-bundles floor: it now fails outright when fewer than 10 tracked bundles exist, before the build/diff check, so the never-tracked case can't pass vacuously again. CHANGELOG claim is accurate as of those commits.

plugins/code/skills/design-inventory/scripts/src/render-review-html.ts
shafty023: Great catch, and it outlived the file: render-review-html was deleted in the review-doc rework, but the same collision class carried into the replacement (derive-decisions-from-doc's surviving-id set, render-review-doc's anchors, and plan-ticket-graph all treated theme ids as global while uniqueness was only validated per-doc, so unit A's surviving theme would mark unit B's same-id theme accepted). Fixed in c7bb46f with your hard-fail option: theme ids are now contractually unit-scoped (thm-<unit-slug>-<topic>, analyst prompt + example updated), and a shared theme-id-guard makes render/derive/plan exit 1 listing each duplicate id with the units involved. Tests cover the collision and happy paths in all three tools.

tools/design-inventory/src/design-export-extract.ts
shafty023: Fixed in c7bb46f: the limit check moved into fflate's filter callback, summing originalSize per entry before decompression and rejecting unsafe paths there too, so an over-limit archive aborts before inflation. One honest caveat: declared sizes in local headers are forgeable, so the post-extract checks remain as defense in depth. Test injects a small maxTotal and asserts the throw happens pre-extraction.

tools/design-inventory/src/capture-design-shots.ts
shafty023: Fixed in c7bb46f: serveDir now routes through a containedPath helper using the same resolve-plus-relative-escape containment as the extractor, 404 on anything outside the extract root. Tests cover plain and percent-encoded traversal.

tools/design-inventory/src/build-route-map.ts
shafty023: Consolidated in c7bb46f: new shared fs-walk.ts (descent-time pruning of node_modules and all dot-directories, try/catch per readdir, symlinks never followed), carrying over the monorepo rationale comment. All three private walkers (route-map, component-index, extract-visual-spec's rglob) are deleted in favor of it; grep confirms one implementation remains.

tools/design-inventory/src/build-route-map.ts
shafty023: Fixed in c7bb46f: comment corrected to match the plain utf8 read; no fallback added.

tools/design-inventory/src/extract-visual-spec.ts
shafty023: Fixed in c7bb46f via the shared fs-walk adoption: readdir wrapped in try/catch, dot-directories (including .git) pruned at descent, symlinks not followed, so a dangling symlink can no longer kill spec extraction.

tools/design-inventory/src/build-design-pack.ts
shafty023: Fixed in c7bb46f: both copy loops now preserve the source-relative path under design-source/ and screenshots/ (recursive mkdir per file), so same-basename files land at their own nested paths. Collision test added with two index-style files in different directories.

tools/design-inventory/src/build-design-pack.ts
shafty023: Fixed in c7bb46f: pending findings now render as an 'Undecided findings (excluded from this ticket's scope)' section in the UI ticket body, the CLI summary JSON carries pending count and ids, and the exit-3 message distinguishes all-declined from pending-remaining, so nothing leaves scope silently.

CLAUDE.md
shafty023: Fair, it deserved daylight: the PR description now has a 'Repo policy changes' section calling out the once-per-branch rewrite (plus the two other policy additions this PR makes: plugin payload rule and TS-only tool scripts). For the record the rewrite came from the repo owner correcting my once-per-commit reading mid-branch; CI behavior is unchanged.

PR #156, opened 2026-06-15
fix(code-review): isolate local PR review at the PR head

wongk:
## Problem

Running `/code-review <PR>` **locally** while not on the PR's head branch dismissed **every** finding. The diff was computed from the fetched remote refs (`origin/<base>...origin/<head>`), but reviewer and verifier agents read source via Read/Grep against the **working tree** (a different branch). The verifier's existence check found none of the PR's code, so it rejected every finding with `evidence_not_found`. GitHub CI mode was unaffected because the runner checks out the PR head.

## Fix — PR-head worktree isolation

`resolve-scope` now isolates source reads for local PR review:

- Reads the working tree directly **only** when `HEAD` already is the PR head **and** the tree is clean.
- Otherwise materializes a detached git worktree at the PR head SHA under the session dir, surfaced as `review_root`/`worktree_path`/`head_sha` in `scope.json`.
- Reviewer prompts and per-finding verifier inputs (including the dismissed-finding second-opinion fleet) read source under `review_root`.
- **Fails closed**: if isolation is required but can't be established (head unresolvable, or `git worktree add` fails), the run aborts with an actionable message rather than reviewing the wrong branch.
- Skipped for hygiene-only / staged / file / branch scope and GitHub mode.

## Hardening (from review feedback)

- `review_root`/`worktree_path` validated against the canonical `<cr_dir>/pr_head_worktree` before use or destructive teardown; basename guard backstops `_remove_pr_head_worktree`.
- Startup GC reclaims worktrees orphaned by runs that aborted before the footer teardown; cleanup commands never raise.
- Graph-aware reviewers (BHB / Impact / fast-path) run grep-only (`GRAPH_PROJECT=""`) under isolation, since the knowledge graph indexes the operator checkout, not the PR head.
- Override content-hash validation resolves cited files under `review_root`.

## Tests / validation

- New: fail-closed, dirty-tree-isolates, forged-`review_root` rejection, footer-teardown path validation, never-raises teardown, dismissed-input passthrough, github-mode-with-PR-scope guard isolation, and an end-to-end resolve → verify-prepare → footer flow test (pins the PR head SHA reaches `git worktree add`).
- `pytest plugins/code-review/tools/python/` → **1228 passed, 5 skipped**; `ruff` + `pyright` clean.
- `code-review` plugin version bumped 2.33.0 → **2.34.0** (MINOR); CHANGELOG generated via `/update-documentation`.


plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -274,6 +274,186 @@ def _resolve_pr_scope(
...
+        abspath = os.path.abspath(raw_line[len("worktree "):].strip())
+        if os.path.basename(abspath) != "pr_head_worktree":
+            continue
+        if abspath == current_expected:
+            continue
+        # Only reclaim worktrees that live directly under THIS repo's
+        # code-review CR root (``<cr_root>/cr-*/pr_head_worktree``).
+        if os.path.dirname(os.path.dirname(abspath)) != cr_root:
+            continue
+        _remove_pr_head_worktree(abspath)
```
thadeusb, line 444: `_gc_stale_pr_head_worktrees` will destroy worktrees that belong to a concurrent in-flight review, not just orphans. It skips `current_expected` (the current run's path) but force-removes every other `cr-*/pr_head_worktree` under the same CR root with no check for whether the owning run is still active.

**Scenario:** session A starts a local PR review, creates `cr-1/pr_head_worktree`, agents start reading from it. Session B starts a second local PR review on the same repo, creates `cr-2`, runs `_gc_stale_pr_head_worktrees(cr-2)`, sees `cr-1/pr_head_worktree` is not `current_expected`, removes it. Session A's agents now get file-not-found on every read, which is exactly the bug this PR is fixing.

Narrow in practice since concurrent local reviews are unusual, but the failure mode is silent and destructive. Checking for a `.git/worktrees/<name>/locked` file, or just skipping force removal when the worktree directory actually exists (let `git worktree prune` handle truly stale entries), would close it.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -8974,6 +9274,18 @@ def cmd_re_assert(args: argparse.Namespace) -> int:
...
+    # Anchor the override hash to the same source the next run's
+    # ``_override_is_valid`` will hash: the PR-head worktree when one is
+    # active. The worktree is ephemeral (torn down by the footer), so only
+    # use it while it still exists on disk; otherwise hash the working tree.
+    re_assert_scope = _read_optional_json(cr_dir / "scope.json", {})
+    review_root = _validated_review_root(
+        cr_dir,
+        re_assert_scope.get("review_root") if isinstance(re_assert_scope, dict) else None,
+    )
+    if review_root and not os.path.isdir(review_root):
```
thadeusb: When the worktree is gone (footer already tore it down), this falls back to hashing against the working tree. But the next `/start` creates a fresh worktree and hashes against the PR head. If the operator's checked-out branch has different content at the finding's `file:line`, the hash won't match and the re-assert override is silently dropped on the next run.

`head_sha` is still in `scope.json` at this point, so `git show <head_sha>:<file>` would give a stable hash without needing the worktree on disk. Worth it, or too narrow to care about?

thadeusb (approved):
Worktree isolation logic is solid, validation and fail-closed design are well done. Two edge cases on the GC and re-assert paths, see inline.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Good catch — fixed in d7f2eb7. The GC now applies an age guard (`_PR_WORKTREE_STALE_SECONDS` = 6h, far above any real review wall-time): a present worktree directory is reclaimed only when its mtime is older than the threshold. A concurrent in-flight review's worktree was created moments ago and is only ever read from (mtime stays at creation), so a live sibling is never deleted; `git worktree prune` still reclaims vanished registrations. Regression test: `TestGcStalePrHeadWorktrees::test_skips_recent_sibling_removes_old`.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Worth it — fixed in d7f2eb7. `_file_content_hash` now anchors the override hash to the PR head: it reads the worktree when present, else falls back to `git show <head_sha>:<file>` (`head_sha` persists in scope.json and is validated as a hex digest via `_validated_head_sha` before interpolation, so it can't be a ref expression or option-like value). The worktree is a detached checkout of `head_sha`, so both paths yield identical windowed content — a re-assert after the footer tore the worktree down now matches the next run's verifier hash instead of silently dropping. Threaded through `cmd_re_assert`, `cmd_verify_prepare`, and `cmd_review_dismissed_consolidate`; tests in `TestFileContentHashHeadSha`.

wongk:
Both inline edge cases addressed in d7f2eb7:

- **GC concurrency** — startup GC now reclaims a present worktree only when it's older than a 6h age guard, so a concurrent in-flight review's freshly-created worktree is never deleted (vanished registrations are still handled by `git worktree prune`).
- **Re-assert hashing** — override hashes now anchor to the PR head via `git show <head_sha>:<file>` when the worktree is gone, so an override re-asserted after teardown survives the next run instead of being silently dropped.

New tests: `TestGcStalePrHeadWorktrees`, `TestFileContentHashHeadSha`. Full suite 1232 passed; ruff + pyright clean.

PR #159, opened 2026-06-16
feat(code): add data provenance layer to design-inventory backend gaps

shafty023:
## Problem

The design-inventory skill decomposes a design into `ui` and `api` tickets. A `backend-gap` finding means the UI needs data the backend does not provide, and the planner emitted exactly one `api` ticket for it. But that stops at the SERVING layer and silently assumes the data already exists upstream. The real pipeline is origin/capture to ingestion/sync to storage/model to API/serving to UI. In a monorepo where data is produced by a desktop client, an agent harness, or webhooks and synced in, the api ticket alone is wrong: the data-origin ticket was missed, so nobody tracked where the data actually comes from.

## Fix

Force a provenance trace on every backend-gap finding, and when the data is not captured or ingested yet, emit a SEPARATE data-source ticket that BLOCKS the api ticket.

### `data_flow` provenance block (schema)

Backend-gap findings now carry a REQUIRED `data_flow` block, validated against `GAP_LAYERS = ["capture","ingestion","model","serving","unknown"]`:

```
data_flow: {
  gap_layer: "capture" | "ingestion" | "model" | "serving" | "unknown";
  origin: string;            // source-of-truth component, discovered by grep
  captured_today: boolean;   // is the raw data produced at the source today?
  ingested_today: boolean;   // does a sync path land it in the platform DB today?
  refs?: string[];           // capture/ingestion/model pipeline paths
}
```

Non-backend-gap findings only shape-check `data_flow` when it is present. `SCHEMA_VERSION` stays 1 (additive; every consumer lives in this repo). The analyst traces the data to its source of truth (model first, then ingestion, then capture) by grepping the repo before classifying the deepest missing layer.

### New `data` ticket kind and the data to api to ui blocks chain

`plan-ticket-graph` adds a third ticket kind, `data`. For each unit with accepted backend-gap findings it still emits the `api` ticket, and ADDITIONALLY emits a `data:<unit>` ticket when any accepted backend-gap is at the capture or ingestion layer. The blocks edges now form the chain `data -> api -> ui`: the data ticket blocks the api ticket (data must be captured and synced before an endpoint can serve it), the api ticket blocks the ui ticket. A serving- or model-layer gap gets no data ticket.

### build-design-pack bodies

- The API body gains a `## Data Provenance` section: one bullet per backend-gap (`layer / origin / captured today / ingested today`, refs sub-bullet), plus a note when a capture/ingestion gap means a separate data-source ticket blocks it.
- A new `ticket-body-data.md` is written for capture/ingestion gaps: it lists the data to capture and ingest (origin, what is missing - instrument the source and/or add the sync mapping, pipeline refs) and notes it BLOCKS the unit's API/serving ticket.

## Tests

- Schema: `data_flow` required and validated on backend-gap; rejected when missing or malformed; ignored but shape-checked on other categories.
- plan-ticket-graph: data ticket emitted for a capture/ingestion gap with the `data -> api` and `api -> ui` edges; not emitted for a serving- or model-only gap; emitted once when capture and serving gaps coexist.
- build-design-pack: api body contains the Data Provenance section; `ticket-body-data.md` written for capture/ingestion gaps and absent for serving/model-only; data body names the origin and the BLOCKS note; replay removes a stale data body.

Existing fixtures and tests in build-design-pack, render-review-doc, and derive-decisions-from-doc that build backend-gap findings were updated to carry a `data_flow` block. Full suite: 393 tests pass; typecheck clean. Regenerated the affected dist bundles; plugin bumped to 1.14.5.

---

## Update: pipeline edges are RELATES_TO

Follow-up commit (`60af2db`): the pipeline adjacencies in the ticket graph (data -> api, api -> ui) are now `RELATES_TO` edges, not `BLOCKS`. Those layers are parallelizable; the api and ui build against the contract and render empty / no-data states until the upstream data lands (graceful degradation), and the data model often lives inside the serving ticket, so a hard block over-constrained them. `BLOCKS` is now reserved for a genuine build-time prerequisite: a net-new shared component built in its primary unit's UI ticket BLOCKS the consumer UI tickets that import it. The plan now emits both `blocks` and `relates` edge arrays; C4 creates a BLOCKS link per blocks edge and a RELATES_TO link per relates edge.


tools/design-inventory/src/design-findings-schema.ts
```diff
@@ -40,6 +40,17 @@ export const FINDING_CATEGORIES = [
...
 ] as const;
+/**
+ * Pipeline layers a backend-gap can sit at, deepest-missing first.
+ *   capture   - the raw data is not produced/captured at its source today.
+ *   ingestion - the data is produced but no sync/ingestion path lands it in the platform DB.
+ *   model     - the data lands somewhere but the storage/domain model does not hold it.
+ *   serving   - the data exists in the platform DB but no API/endpoint serves it to the UI.
+ *   unknown   - the provenance could not be traced.
+ * A capture- or ingestion-layer gap means nobody tickets where the data comes
+ * from; it becomes a separate data-source ticket that blocks the serving/API ticket.
```
thadeusb, line 51: This docstring says "blocks the serving/API ticket" but the entire point of this PR is that the data-source ticket RELATES_TO the API ticket (parallel, not sequential). The code, the agent prompt, the SKILL.md, and the plan-ticket-graph all say RELATES_TO. This docstring is the one place that still says BLOCKS. CLAUDE.md [mistake]: audit adjacent comments when modifying behavior.

tools/design-inventory/src/plan-ticket-graph.test.ts
```diff
@@ -272,14 +293,95 @@ describe("buildTicketGraph", () => {
     expect(plan.blocks).toHaveLength(0);
   });
 
-  it("API ticket BLOCKS UI ticket when both exist", () => {
+  it("API ticket RELATES_TO UI ticket when both exist", () => {
     const visual = makeFinding("CHG-scr-b-01", "visual");
-    const backend = makeFinding("CHG-scr-b-02", "backend-gap");
+    const backend = makeBackendGap("CHG-scr-b-02", "serving");
```
thadeusb, line 298: The `makeFinding(..., "backend-gap")` calls at lines 263 and 286 produce findings without `data_flow`, which is now schema-invalid. These pass because `buildTicketGraph` skips validation, but they're lying about what real data looks like. Swap to `makeBackendGap(id, "serving")` so the fixtures actually match the schema.

tools/design-inventory/src/build-design-pack.test.ts
```diff
@@ -302,6 +314,183 @@ describe("build-design-pack", () => {
     expect(existsSync(join(pack, "ticket-body-api.md"))).toBe(false);
   });
 
+  // -------------------------------------------------------------------------
+  // Data provenance: data_flow on backend-gap findings drives a data-source ticket
+  // -------------------------------------------------------------------------
+
+  /** A backend-gap finding with a data_flow at the given gap layer. */
+  function backendGap(
```
thadeusb, line 322: This `backendGap()` factory is a near-copy of `backendGapFinding()` in test-fixtures.ts. Same story with `makeBackendGap()` in plan-ticket-graph.test.ts. Three files, three factories, same shape. CLAUDE.md [mistake]: extract test helpers used by 2+ files into shared modules. Use the one in test-fixtures.ts.

thadeusb (approved):
Good feature, clean implementation. Three things to fix, all straightforward, see inline.

PR #175, opened 2026-07-02
feat(code-review): wire start.md to the run-prefix loop (PLN-1229 Phase 3)

wongk:
## Summary

**PLN-1229 Phase 3 — orchestrator wiring.** The payoff phase: rewires the `/code-review` orchestrator (`start.md`) to run the deterministic prefix via the `run-prefix` loop (built + parity-proven in #173/#174) instead of walking ~19 helper stages one at a time. This collapses the prefix's serial model turns into a handful of orchestrator turns — the single biggest turn-count reduction in a review — while keeping the per-stage walk as a live, documented fallback (no behavior lost).

This is a **prose-only** change (orchestrator command + one skill + a SCHEMA doc fix). The runner it drives is already byte-parity-proven; this phase's safety net is careful authoring + an expert review of the rewire.

## What changed

**`commands/start.md`**
- New **"Deterministic Prefix — `run-prefix` loop"** section: invoke `run-prefix`, then dispatch on `next_action` —
  - `needs_singleton` → invoke the `singleton-dispatch` skill, then re-invoke `--resume-from <resume_stage>`;
  - `hygiene_exit` → Gate A present + EXIT;
  - `ready_for_reviewers` → print the cache/fast-path notices, read `CACHE_DIR` from `cache_config.json`, hand off to the Walker Contract at `stage_20_spawn_reviewers`;
  - `error` → fall back to the per-stage walk from `failed_stage`.
- Reframed the **Walker Contract** as (a) the always-walked reviewer/verify/present tail and (b) the labeled per-stage prefix fallback. Gate A/B now execute inside `run-prefix` (their prose is the fallback recipe); Gate C/D still fire in the tail. Updated the "Batch deterministic helper stages" note to point at `run-prefix`, and kept the operator todo list accurate across the now-silent prefix.

**`skills/singleton-dispatch/SKILL.md`** — step 4 now branches on caller: return to `start.md` for `--resume-from` in the `run-prefix` flow; advance the walker only in the fallback.

**`SCHEMA.md`** — corrected the route-failure `failed_stage` to `stage_19_cache_check` (matching the merged runner) and noted the shallow-tier `derive-static-spec` stage.

## Review

An expert review traced all four `next_action` outcomes against the runner code and confirmed the tail resolves everything it needs from on-disk artifacts (**no HIGH findings**). All flagged items fixed: the stale skill step 4, fast-path notice model-name parity, the shallow-tier stage note, and the SCHEMA `failed_stage` value.

## Testing

- Full code-review suite: **1306 passed, 3 skipped** (prose-only change — confirms no code impact)
- `ruff` / `pyright` unaffected (no code changed)

Version: code-review **3.3.0 → 3.4.0** (MINOR — additive with a live fallback; touches the orchestration contract).

Completes the landable sequence Phase 0→3. Phase 4 (measure the turn/cost reduction with the cost tool) is the remaining follow-up.

### DIFF
--- a/plugins/code-review/.closedloop-ai/subagent-start-hook-debug.log
+++ b/plugins/code-review/.closedloop-ai/subagent-start-hook-debug.log
@@ -1,1 +1,0 @@
-Thu Jul  2 13:07:59 CDT 2026: SubagentStart hook fired — agent_type=general-purpose agent_id=a0547cb37060e68fd session_id=a9c0487b-72c4-41de-9063-fa1f697c8743
--- a/plugins/code-review/.closedloop-ai/subagent-start-hook-debug.log
+++ b/plugins/code-review/.closedloop-ai/subagent-start-hook-debug.log
@@ -1,0 +1,1 @@
+Thu Jul  2 13:07:59 CDT 2026: SubagentStart hook fired — agent_type=general-purpose agent_id=a0547cb37060e68fd session_id=a9c0487b-72c4-41de-9063-fa1f697c8743
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,9 @@ __pycache__/
 .closedloop-ai/*
 !.closedloop-ai/settings/
 !.closedloop-ai/bootstrap-metadata.json
+# Plugin-local runtime working dirs must never be committed (the patterns above
+# are root-anchored; hook/review artifacts can otherwise land under plugins/**).
+plugins/**/.closedloop-ai/
 
 # Generated route knowledge files
 .claude/docs/routes/*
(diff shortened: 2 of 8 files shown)


plugins/code-review/.closedloop-ai/subagent-start-hook-debug.log
wongk, line 1: remove this superfluous file

wongk: Removed in 6957a10. Also added a `plugins/**/.closedloop-ai/` guard to `.gitignore` — the existing `.closedloop-ai/` patterns are root-anchored, so plugin-local runtime dirs (like this SubagentStart hook artifact) were never covered and could slip in again.

PR #179, opened 2026-07-13
fix(code-review): revert #178 parallel GitHub fleet → restore synchronous headless dispatch (FEA-3085)

peterulsteen:
## Summary

Reverts #178 (`code-review` v3.5.0) and bumps **3.4.1 → 3.6.0**, restoring **synchronous** reviewer-fleet dispatch for GitHub/headless mode at `stage_20_spawn_reviewers`. Local mode keeps parallel background dispatch + blocking `TaskOutput`.

#178 unified both modes on `run_in_background: true` + blocking `TaskOutput`, betting that headless `claude -p` (Claude Code ≥ v2.1.182) reliably awaits background subagents *as long as the orchestrator issues the blocking collection*. Empirically it does **not**: in ~18% of `symphony-alpha` `claude-code-review.yml` `pull_request` runs on 2026-07-13 (9/50, up from 1/10 the day #178 shipped) the orchestrator launched the reviewer fleet (Bug Hunter A/B, Unified Auditor) as background agents and then **ended its turn** ("I'll wait for the harness to notify me") with no pending synchronous tool call. `claude -p` then exited (`terminal_reason: "completed"`) before `stage_21_collect_findings … stage_30_footer` ran — no findings collected, no `.closedloop-ai/code-review-*` artifacts written, and the workflow's **"Validate migrated review artifacts"** guard failed the job. Runs that took the synchronous fast-path were unaffected, hence the intermittency.

Synchronous one-at-a-time dispatch (`run_in_background: false`, wait for each `DONE findings=N file=...` before the next) keeps the headless turn alive **by construction**, rather than depending on the model choosing to issue a blocking `TaskOutput`. This restores the exact pre-#178 state that ran reliably (the verifier fleet at `stage_23` was already background+collect in both modes before #178 and is unchanged).

**Trade-off:** GitHub review wall-clock returns to sum-of-reviewers instead of slowest-reviewer. Correctness over speed for a merge-adjacent gate.

Root-cause detail and the 9 failing run IDs are tracked in **FEA-3085**.

## Feature Flags

N/A — plugin behavior change, no runtime flag.

## Breaking changes

None. Restores prior GitHub-mode dispatch semantics; output contract and artifacts unchanged.

## Test plan

- `uv run --frozen --group dev pytest` → **2079 passed, 3 skipped**
- `ruff check .` → clean
- `pyright` → 0 errors, 0 warnings
- Restored FEA-2162 synchronous-GitHub reviewer test class passes.
- Post-merge: watch `symphony-alpha` `claude-code-review.yml` `pull_request` runs for the artifact-validation failure signature (expect it to clear once CI pulls v3.6.0).


plugins/code-review/skills/verify-findings/SKILL.md
```diff
@@ -30,10 +30,6 @@ This stage runs when the walker reaches `stage_23`. It implements PLN-722's find
...
-
-> **Legacy fallback (Claude Code < v2.1.182 only) — an operator concern, NOT a runtime branch the agent picks.** Same as the reviewer fleet (see `spawn-reviewers/SKILL.md`): the orchestrating agent cannot introspect its own CLI version, so it always takes the parallel path above, and `anthropics/claude-code-action@v1` ships ≥ v2.1.182. An operator forced onto an older CLI — where headless does not await background subagents, so a backgrounded verifier dies when the turn ends — must edit this skill to dispatch verifiers synchronously one at a time so no verifier is left outstanding. This mirrors the reviewer fleet so both stages degrade the same way on older CLIs.
-
 For each entry in `verify_manifest.json.to_verify[]`:
 
 1. Spawn one background `Task` with `subagent_type: "code-review:code-review-worker"`. The agent's tool allowlist (`Read`, `Write`, `Grep`, `Glob`) is identical to the Reviewer Fleet's — no permission changes needed.
```
mikeangstadt, line 35: This is the part that bugs me about the revert. The whole justification is that headless claude -p does NOT reliably await backgrounded subagents: the orchestrator backgrounds the fleet, ends its turn ("I'll wait for the harness to notify me"), and the process exits before the tail stages run. But stage_23 verifiers still dispatch background + blocking TaskOutput in MODE=github here, the exact pattern you just declared unsafe for stage_20.

The difference is only how loud the failure is. A dead reviewer fleet trips the workflow's artifact-validation guard and fails the job, which is how you caught #178. But stage_23 is on_failure: continue and missing outputs degrade to pending_verification[]. So if the same turn-ending happens at stage_23, BLOCKING/HIGH findings ship unverified on a merge-adjacent gate and nothing fails, nobody sees it. That's arguably worse than the reviewer symptom you're reverting.

So either the background-await is actually reliable and reviewers didn't need reverting, or it isn't and verifiers need the same synchronous treatment. Which is it? Fine to ship this revert as-is since it strictly restores known-good reviewer behavior, but I want the verifier answer tracked on FEA-3085, not left implicit.

CHANGELOG.md
```diff
@@ -4,10 +4,10 @@ All notable changes to the claude-plugins project will be documented in this fil
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/). Entries are listed newest-first; each plugin section is treated as released when merged to `main`.
 
-### code-review v3.5.0
+### code-review v3.6.0
```
mikeangstadt, line 7: This replaces the v3.5.0 entry rather than adding v3.6.0 above it. Line 5 right here says entries are newest-first and treated as released when merged to main, and v3.5.0 did merge and ship, so deleting its entry erases that it ever existed. Keep the v3.5.0 Changed block and put v3.6.0 on top of it. Also CLAUDE.md says regenerate the changelog via /update-documentation, not hand-edit it.

mikeangstadt (approved):
Sound revert. Restores the sync GitHub dispatch that we know keeps the headless turn alive, and the version bump + tests line up. One real question about whether the verifier fleet has the same headless death you're fixing here, plus a changelog nit. Not blocking either.

CHANGELOG.md
peterulsteen: Fixed in 080cae3. Restored the v3.5.0 `Changed` block and put v3.6.0 above it — newest-first preserved, and v3.5.0's shipped history stays.

On the `/update-documentation` convention: acknowledged. This ran from a headless symphony-alpha session that can't invoke the plugin's slash command, so I hand-reconciled to match what the generator produces — version heading == `plugin.json` (3.6.0), entries newest-first, single root `CHANGELOG.md` (no `plugins/<name>/CHANGELOG.md`), and every claim maps to pre-PR behavior. Say the word if you'd rather I regenerate it via `/update-documentation` before merge.

plugins/code-review/skills/verify-findings/SKILL.md
peterulsteen: Tracked: **FEA-3154**. You're right, and it's the worse half — `stage_23` is `on_failure: continue`, so the same turn-ending degrades **silently** to `pending_verification[]` instead of tripping the artifact guard.

The dichotomy has a third resolution: a blocking `TaskOutput`, *when actually issued*, is a pending synchronous call that **does** keep the turn alive — the `stage_20` bug was the orchestrator ending its turn **without** issuing it ("I'll wait for the harness…"). Synchronous dispatch removes that skippable step. So reviewers genuinely needed the revert (observed ~18% failures); verifiers carry the same latent risk, just unobserved (silent degradation ≠ safe).

Not folding the verifier fix into this PR on purpose: github-synchronous verifiers are **net-new** behavior — pre-#178 verifiers were background in *both* modes, so there's nothing to revert to and it needs its own tests, not a smuggle into a strict revert. FEA-3154 covers making `stage_23` github-synchronous and/or making the degradation loud (surface a coverage-gap artifact on missing verifier output). v3.6.0's changelog now states `stage_23` is unchanged and points at it.

PR #183, opened 2026-08-04
feat(code-review): review non-application files + three TestQuality shapes

mikeangstadt:
## Summary

Extends the shared reviewer prompt to cover the file classes the fleet currently does not review, plus three `TestQuality` shapes that stay green while proving nothing.

**Prompt-only.** All behavior lives in `tools/prompts/shared_prompt.txt`, which every reviewer role loads. No dispatch, roster, category, or schema change; no new agent; no new stage.

## Why: the fleet's coverage is categorically absent outside application source

Measured against five days of one consuming repo — **2,617 original review comments** (replies excluded) across **326 PRs**, 2026-07-30 → 2026-08-04, attributed per reviewer agent:

| file class | total findings | from this fleet | fleet share |
|---|---:|---:|---:|
| CI guard/deploy scripts | 101 | **0** | 0% |
| workflow definitions (`.github/`) | 69 | **0** | 0% |
| docs / MDX | 94 | **0** | 0% |
| migrations / schema | 28 | **0** | 0% |
| test files | 219 | **1** | 0.5% |
| app code `.tsx` | 684 | 465 | **68%** |
| prototypes | 193 | 139 | **72%** |
| app code `.ts` | 1,099 | 301 | 27% |

The fleet is strong where it is pointed and produces **zero** findings on four whole file classes. Those classes are **22% of the corpus** and are carried almost entirely by other reviewer agents — so the capability gap is real, not a question of whether those files contain reviewable defects.

Note the limit of this measurement: it shows output, not cause. It does not by itself prove the reviewers *would* have found these defects if prompted — only that they are not reporting them today. If the zeros turn out to come from file-scope filtering in partitioning rather than from prompt silence, this PR is necessary but not sufficient, and the follow-up is in `derive-spawn-spec`/partitioning rather than here.

## What changed

**`<non_application_files>`** — declares those files in scope and enumerates the generic defect classes per class, all repo-agnostic:

- *CI/pipeline*: privileged jobs running repository lifecycle scripts before minting a write token; workflow-scope write permissions and persisted checkout credentials; the implicit "all dependencies succeeded" condition that is **false for a skipped dependency**; missing job timeouts; cancel-on-new-run over irreversible side effects; trigger filters that do not match what they gate; a shell step whose unguarded command pre-empts the branch owning its exit code.
- *Guard/meta-test scripts*: guards that assert a predicate **appears** rather than executing the decision; guards anchored to the first matching block instead of the production node; allowlist entries broad enough to suppress future real violations.
- *Migrations*: destructive statements selecting survivors by a proxy for canonicity; idempotency guards that no-op over an invalid object left by a failed run; an enum value added without updating every classifier that switches on it.
- *Docs/API specs*: claims stronger than the code enforces (unconditional where the code is conditional, product-wide where the behavior is one lane); a behavior change whose describing doc is not updated; generated-client specs admitting combinations the runtime always rejects.

**Pinned-file pair detection** — repos routinely hold meta-tests asserting another file's literal content (an action-version pin, a config snapshot, a required-context manifest, an allowlist). Reviewers now grep for a test referencing the changed path or a distinctive changed literal and report when that test exists but is not in the diff. Emitted as **`Correctness`**, not `CompanionChange`, deliberately: it is a provable CI failure with a concrete assertion to cite, and it carries no rule-driven `reasoning_certificate` — which `templates/companion_change.md` interpolates and would render empty for an LLM-emitted finding.

**`<test_quality>`** — three shapes, all generic:

1. *Wiring unproven* — helper tested directly, production caller never asserted; deleting the call site leaves the suite green.
2. *Assertion of existence* — source grepped/AST-inspected for a predicate instead of executing that decision against synthetic inputs, so it cannot distinguish a correct implementation from a widened one.
3. *Input-invariant fixture* — a mock or route handler returning identical results regardless of the params the test claims to cover.

New `TestQuality` subcategories route to the existing `templates/testquality_specialized.md` via the `TestQuality/*` fallback row, so `/code-review fix` needs no new template.

## Test plan

- `python3 -m pytest test_code_review_helpers.py test_code_review_schema.py` — **1,286 passed**, 0 failed.
- Checked that no test pins `shared_prompt.txt` content before editing it (the same rule this PR teaches): every test writes a synthetic prompt file to `tmp_path`; the real file is only ever copied verbatim into the CR dir by `code_review_helpers.py:13719`. No pinned-file pair broken.
- Verified only `.claude-plugin/plugin.json` is bumped per release (`.codex-plugin/plugin.json` tracks separately and was last touched by an unrelated commit), matching the v3.7.0 release commit's file set. Bumped 3.7.0 → 3.8.0 with a CHANGELOG entry.
- `git diff --check` clean; `plugin.json` parses.

**Cost.** `shared_prompt.txt` grows 317 → 409 lines (~17.5 KB → ~24 KB), about **+1.6k input tokens per reviewer agent**. It is identical across agents and stable across runs so it caches, but on a large fleet the uncached first-touch cost is real — `/code-review cost` will show the fleet-side delta.

**What could still break.** Prompt additions can shift reviewer attention: a fleet told to review workflow YAML may spend budget there on PRs where the application diff mattered more, and the new heuristics could produce false positives on repos whose CI conventions differ (a deliberate `cancel-in-progress`, an intentionally broad allowlist). The existing FLAG criteria and the verifier pass both still apply, and every new heuristic requires a concrete file:line, so these should surface as verifier-refuted rather than shipped — but the first few runs on a CI-heavy PR are worth watching.


wongk (changes requested):
this adds explicit things to review to the shared reviewer prompt, rather than assigning these things to specific reviewers. this means every review agent is going to duplicate these reviews, and it will dilute the agent-specific review areas.

shafty023:

plugins/code-review/tools/prompts/shared_prompt.txt
```diff
@@ -102,6 +102,98 @@ Do NOT flag:
 - Hypothetical edge cases dependent on specific inputs/state
 </constraints>
 
+<non_application_files>
+Config, CI, infrastructure, migration, docs, and test files in <files_assigned>
```
shafty023, line 143: This puts the full domain checklist in `shared_prompt.txt`, so every BHA partition, BHB, auditor, Design Critic, and Impact Analyzer pays the ~1.6k-token cost and is told to repeat work outside its assigned role. Deduplication can hide duplicate output, but it cannot recover the review attention already spent. Can we keep only the scope declaration shared and route each checklist by file class/owning reviewer instead?

plugins/code-review/tools/prompts/shared_prompt.txt
```diff
@@ -102,6 +102,98 @@ Do NOT flag:
...
+- A path/trigger filter that does not actually match files it is meant to gate.
+- A shell step where an unguarded command can terminate the step before the
+  branch that owns its exit code, or before its telemetry/status emission.
+
+GUARD AND META-TEST SCRIPTS (lint rules, source tests, policy checks):
+- PINNED-FILE PAIR BROKEN. Repos commonly hold meta-tests that assert another
+  file's literal content — a workflow's pinned action version, a config
+  snapshot, a required-context manifest, an allowlist. When the diff edits such
+  a file, grep for a test referencing that path or a distinctive literal from
+  the changed lines; if that test exists and is NOT in the diff, the change
```
shafty023: A path reference is not proof that a pinned assertion is stale. This PR itself is a counterexample: `test_code_review_helpers.py:21401-21412` references the changed `plugin.json` path as fixture input but does not assert its version or content. Also, partitioned BHA workers only receive their own patch, so they cannot establish full-PR diff membership. Please require a cited assertion that actually contradicts the pinned-head file instead of inferring a CI failure from a reference plus absence from the worker patch.

plugins/code-review/tools/prompts/shared_prompt.txt
```diff
@@ -102,6 +102,98 @@ Do NOT flag:
...
+  breaks it. This is a provable CI failure, so report it as "Correctness" with
```
shafty023: Routing this as generic `Correctness` breaks the existing `/fix` contract. Diff-scope validation requires the finding anchor to be the changed file, while `/fix` auto-edits only that anchor; for a stale meta-test, `--apply` can therefore revert the intended config/workflow change instead of updating the unchanged assertion. Please give this case a dedicated subcategory routed to manual or multi-file handling, with structured metadata identifying the companion test.

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ All notable changes to the claude-plugins project will be documented in this fil
...
... a write token; workflow-scope write permissions and persisted checkout credentials; the implicit "all dependencies succeeded" condition that is false for a *skipped* dependency; missing job timeouts; cancel-on-new-run over irreversible side effects; trigger filters that do not match what they gate; a shell step whose unguarded command can pre-empt the branch owning its exit code; guards that assert a predicate *appears* rather than executing the decision; allowlist entries broad enough to suppress future real violations; destructive migrations selecting survivors by a proxy for canonicity; idempotency guards that no-op over an invalid object left by a failed run; an enum value added without updating every classifier; docs claims stronger than the code enforces; generated-client specs admitting combinations the runtime rejects), and a `<test_quality>` block naming the three shapes that stay green while proving nothing — wiring-unproven (helper tested, production caller never asserted), assertion-of-existence (source scanned for a predicate instead of executing it), and input-invariant fixture (a handler returning identical results regardless of the params the test claims to cover).
```
shafty023: This release entry includes corpus counts, reviewer-attributed percentages, and a five-day measurement window that are not derivable from the allowed changelog sources. `.claude/commands/update-documentation.md` limits entries to the diff, `plugin.json`, and commit subjects so generated release notes stay reproducible. Please regenerate/rewrite this through `/update-documentation` and keep only claims supported by those sources.

plugins/code-review/tools/prompts/shared_prompt.txt
mikeangstadt: Fixed in 01e357f. `shared_prompt.txt` now carries only the scope declaration you allowed — a `<file_scope_classes>` block saying those file classes are in scope on the same evidence standard, plus a pointer that the checklists live in the role suffixes and an explicit instruction not to reconstruct another role's list from it. 92 shared lines became 17.

Every defect class was moved, none dropped. The split is by whether the defect is decidable from the changed file:

- `tools/prompts/bha_suffix.txt` (Bug Hunter A, the diff-local correctness role): CI/pipeline definitions, migrations/schema, the three guard-script shapes visible in the guard itself, and the two test shapes readable in the test file (assertion-of-existence, input-invariant fixture).
- The Bug Hunter B suffix in `skills/spawn-reviewers/SKILL.md` (cross-file, unpartitioned, one agent, reads `patches_all.txt`): the pinned-file pair, wiring-unproven, and the docs/API-spec classes. Each of those needs a file outside the diff by construction, which is why they are B's. Added to both the standard and fast-path copies of that suffix.

So the auditor, design critic, impact analyzer, domain critics and verifier now pay 17 lines instead of 92 and are told nothing about reviewing outside their role; BHA partitions carry only diff-local correctness, which is their role; and the cross-file classes are paid once by the one unpartitioned agent that can actually act on them.

`uv run pytest plugins/` 2175 passed, 3 skipped; `ruff check .` and `pyright` clean.

plugins/code-review/tools/prompts/shared_prompt.txt
mikeangstadt: Both corrections made in 01e357f. Your counterexample reproduced exactly — `test_code_review_helpers.py:21401-21412` (`test_plugin_json_fires`) names `plugins/code-review/.claude-plugin/plugin.json` in a synthetic `files=[...]` fixture list and asserts only that a nudge fires and that the string `plugin.json` appears in the explanation. Nothing about version or content. Under the old text this PR was a false positive against itself.

The class now lives in the Bug Hunter B suffix with an explicit evidence bar:

> EVIDENCE BAR — a path reference is NOT evidence. You must quote the specific assertion, at its own file:line, whose expected value the pinned-head file no longer satisfies: a literal comparison, an exact-match membership check, a snapshot equality. A test that merely names the changed path — as a fixture input, a synthetic `files=[...]` list, a glob it iterates, a docstring — pins nothing and is not a finding. If you cannot name the contradicted expected value and the new value beside it, you do not have this finding.

And the diff-membership inference is removed outright rather than qualified:

> DO NOT infer the defect from the companion test's absence from the diff. You are not given full-PR diff membership (a partitioned Bug Hunter A worker sees only its own patch, and your `<files_assigned>` is a scope, not a manifest of every file the PR touches), so "the test is not in the diff" is unknowable and is never part of the argument. The contradiction between the assertion and the file as it now stands is the whole case; read the companion test at HEAD and check whether it was already updated before you report.

Reading the companion at HEAD is what replaces the inference: it is decidable from the checkout, where diff membership is not.

plugins/code-review/tools/prompts/shared_prompt.txt
mikeangstadt: Confirmed and fixed in 01e357f. Your `--apply` claim reproduces:

- `skills/fix/SKILL.md:99` (before this change) routed `Correctness` with no subcategory to **auto-fix**, "Direct code edit at anchor line".
- `SKILL.md:119` anchors the drift check and the edit on `finding.file` at `finding.line`.
- Diff-scope validation requires that anchor to be the changed file, so for a stale meta-test the edit lands on the config/workflow change, not on the unchanged assertion — reverting the intended change, as you said.
- There is no fail-safe default for `Correctness` either. Hygiene has one at line 107; a `Correctness` finding carrying an unrecognized subcategory still falls through to the auto-fix row, so a subcategory alone would not have been enough.

Changes:

- The class emits `Correctness` with subcategory `pinned-file-pair`, and the BHB suffix states the subcategory is required "because it is what routes the finding to manual handling instead of an auto-edit at the changed file".
- New dispatch row in `skills/fix/SKILL.md` ahead of the generic `Correctness` row, routing `pinned-file-pair` to **manual-surface**.
- New `templates/correctness_pinned_file_pair.md` rendering both sides and leaving to the operator which one is wrong — the change may be intended and the assertion stale, or the assertion may be the contract. Option 3 is an explicit dismissal path for the fixture-input false positive from the thread above.

Structured companion-test metadata rides the existing `other_locations[]` field (`{file, line, issue}`, `code_review_schema.py:845`), with the contradicted expected value in `issue`. No category enum or schema change — `subcategory` is already free-form `str | None` and `other_locations[]` already flows through the pipeline.

CHANGELOG.md
mikeangstadt: Rewritten in 01e357f against `.claude/commands/update-documentation.md` constraint 7 (derivable from git diff + `plugin.json` + commit subjects).

Removed: the 2,617-comment / 326-PR corpus, the five-day window, "one consuming repo", every per-class ratio (0 of 101, 0 of 69, 0 of 94, 0 of 28, 1 of 219), the 68% / 72% / 22% attributions, and the whole `#### Notes` section — both because the token estimate in it was not derivable and because constraint 3 allows only Added/Changed/Fixed/Removed.

What remains is the code change: which file classes are declared in scope, which defect classes went to which role suffix and on what split, the three-part pinned-file-pair contract, and the `/fix` dispatch row with the auto-fix behavior it prevents. All readable off the diff.

Also in the same commit, since this thread is on CHANGELOG.md: the entry moved from v3.8.0 to **v3.11.0**. `main` had already shipped a v3.8.0 entry, so the branch's heading and `plugin.json` would have merged clean onto a version collision.

mikeangstadt:
@wongk — your CHANGES_REQUESTED is addressed in 01e357f. You and @shafty023 (thread on `shared_prompt.txt:106`) raised the same objection: the checklists were in the shared reviewer prompt, so every agent duplicated them and the role-specific review areas were diluted.

The checklists are out of `shared_prompt.txt` and assigned to specific reviewers. What stays shared is a scope declaration only — those file classes are in scope and carry the same evidence standard — plus an explicit instruction not to reconstruct another role's checklist from it. 92 shared lines became 17.

The assignment is split by whether the defect is decidable from the changed file:

- **Bug Hunter A** (`tools/prompts/bha_suffix.txt`, diff-local correctness): CI/pipeline definitions, migrations/schema, guard-script shape, and the two test shapes readable in the test file.
- **Bug Hunter B** (suffix in `skills/spawn-reviewers/SKILL.md`, cross-file, unpartitioned, standard + fast-path copies): the pinned-file pair, wiring-unproven, and docs/API-spec. Each needs a file outside the diff by construction.

No defect class was dropped. The auditor, design critic, impact analyzer, domain critics and verifier now carry the 17-line declaration and nothing outside their role.

Also in that commit, from @shafty023's three other threads: the pinned-file-pair class now requires citing the contradicting assertion rather than a path reference (his `test_code_review_helpers.py:21401-21412` counterexample reproduced — it names `plugin.json` as fixture input and asserts nothing about it); the "not in the diff" inference is removed, since a partitioned BHA worker cannot establish full-PR diff membership; the class emits `Correctness`/`pinned-file-pair` routed to manual-surface in `/fix`, because generic `Correctness` auto-fixes at the anchor and `--apply` would have reverted the intended config change; and the CHANGELOG entry is rewritten to `/update-documentation`-derivable sources only.

Merged `main` in — the branch was conflicting, and `main` had already shipped a v3.8.0, so the version moved to 3.11.0. `uv run pytest plugins/` 2175 passed, 3 skipped; `ruff check .` and `pyright` clean.

I have not dismissed your review — only you can. Re-review when you have a moment.

shafty023 (approved):

PR #184, opened 2026-08-04
code-review: make DOMAIN_CRITIC_CAP operator-tunable

cchenault4:
## Summary

Makes `DOMAIN_CRITIC_CAP` operator-tunable via `.closedloop-ai/settings/code-review.json`, the file that already hosts `bha_unified_threshold_loc` and `out_of_hunk_confidence_floor`.

**The problem.** The cap was a hardcoded `3` with no settings path, and it drops by `(priority asc, reviewer asc)`. A repo whose `critic-gates.json` uses the legacy `moduleCritics[]` schema gets every entry migrated by `_migrate_module_critics` as `required: False` with **no** `priority` — so all of them sit at the default `2` and the tiebreak degenerates to **alphabetical by reviewer name**.

That is not a theoretical edge. On two `symphony-alpha` cost/metric PRs (#4330, #4349), the coverage critic proposed `logical-metric-reconciliation-auditor` *specifically because of the diff's tally/cost semantics*, consolidate accepted it, and the cap then dropped it — while `api-architect`, `caching-strategist`, and `database-architect` survived on their first letter. The reviewer best matched to the change was the one systematically excluded, on every such PR.

**The change.**
- `_load_code_review_settings` gains `domain_critic_cap` (int, ≥ 0) with the same validation discipline as its siblings: bool rejected (it is an `int` subclass), negative and wrong-type fall back to the default rather than crashing the stage.
- `cmd_arbitrate_budget` reads it with precedence mirroring `bha_unified_threshold_loc`: `--domain-critic-cap` flag → `code-review.json` → built-in `DOMAIN_CRITIC_CAP`.
- `arbitrate-budget` gains `--settings` (test isolation, mirroring `validate`) and `--domain-critic-cap`.
- `0` is a valid, documented kill switch — no domain critics spawn; `source: "core"` reviewers (Design Critic, Impact Analyzer) stay exempt as before.

**Docs.** README and `commands/start.md` no longer state the cap as a fixed `3`. Both now explain *when* raising it is the right call — and note that raising the cap is the blunt remedy, while migrating rules to the canonical `coverage[]` schema (which supports explicit `priority` and `required`) is the precise one.

## Test plan

- **9 new tests.** Five cover loader validation (override honored, `0` kill switch, negative → default, wrong type → default, bool rejected). Four cover `arbitrate-budget` behavior: the settings file actually raising the cap, `--domain-critic-cap` beating the settings file (precedence pinned so a refactor cannot silently invert it), `0` spawning zero domain critics with all of them cap-deferred and **no** coverage-gap findings, and a malformed value leaving the fleet at the default 3.
- **Test-isolation fix:** the shared `_run_arbitrate_budget` harness now always points `--settings` at a test-local path, so a `code-review.json` in the CWD can never leak into a run.
- **Snapshot updated deliberately:** `cli_parser_resolved.json` regenerated. The diff is exactly the two new `arbitrate-budget` flags and nothing else — I diffed parser-by-parser before writing to confirm no collateral drift.
- **Full suite: 1339 passed, 3 skipped, 0 failed** across `test_code_review_helpers.py`, `test_code_review_schema.py`, `test_golden_fixtures.py`, `test_prefix_golden.py` — including the golden-fixture and prefix-parity harnesses.
- **Regression check against baseline:** stashed the change, captured the clean-`HEAD` failure set, restored, and diffed. Exactly one new failure (the CLI snapshot, expected and then updated) and zero others. (An earlier run showed 95 failures on both sides — that was my venv missing `pyyaml`, not the repo.)

What could still break: the cap remains tier-uniform, so `--depth deep` still does not widen it — deep's extra breadth continues to come from the cap-exempt core reviewers. Setting `domain_critic_cap` above the total reviewer `--cap` has no additional effect, since total-cap arbitration still runs first.


shafty023:

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3092,6 +3103,15 @@ def _load_code_review_settings(path: Path | None) -> dict[str, Any]:
         and 0.0 <= float(raw_floor) <= 1.0
     ):
         out["out_of_hunk_confidence_floor"] = float(raw_floor)
+    raw_critic_cap = data.get("domain_critic_cap")
+    # Reject bool (an int subclass) so a stray `true` doesn't become 1.
+    # 0 is valid and meaningful: it is the no-domain-critics kill switch.
+    if (
+        isinstance(raw_critic_cap, int)
+        and not isinstance(raw_critic_cap, bool)
+        and raw_critic_cap >= 0
```
shafty023, line 3112: This makes review policy controllable by the PR being reviewed. In GitHub mode this settings path is read from the checked-out PR tree, so a contributor can add `domain_critic_cap: 0` and suppress every rule/critic reviewer with no coverage-gap finding (or raise it to amplify review work). Can we load this budget from a trusted base/external operator source and add a test proving a head-only settings change cannot alter that PR’s fleet?

plugins/code-review/README.md
```diff
@@ -190,7 +190,7 @@ Three tiers select which reviewer fleet runs:
...
+**Standard-mode budget arithmetic.** With PLN-807 Phase 4, `arbitrate-budget` reserves BHA partitions FIRST (from `_max_bha_partitions_by_loc`) and then allocates the remaining budget to critics and best-effort. The total domain-critic count across both required and best-effort buckets is capped uniformly at `DOMAIN_CRITIC_CAP` (default 3, operator-tunable via `.closedloop-ai/settings/code-review.json:domain_critic_cap`; by priority asc, reviewer asc) for both standard and deep tiers. Required-bucket critics dropped by the cap emit coverage-gap findings; cap-deferred entries carry `defer_reason: "domain_critic_cap"` in `deferred_for_budget`. PRs with sparse critic-gates rosters see identical fleet to pre-PLN-807.
```
shafty023, line 193: This still says required critics dropped by this cap emit coverage-gap findings, but the new zero-cap test explicitly asserts `gaps["findings"] == []`, and arbitration records them only in `deferred_for_budget`. An operator lowering the cap could rely on a safety signal that never appears. Please make this line and `commands/start.md` match the implemented no-gap behavior, or change the implementation if gaps are intended.

plugins/code-review/tools/python/test_code_review_helpers.py
```diff
@@ -8162,6 +8164,14 @@ def _run_arbitrate_budget(
     if verify_doc is not None:
         _write_coverage_section(tmp_path, "verify", verify_doc)
 
+    # ISS-5122: always point --settings at a test-local path so a
```
shafty023, line 8167: These tests bypass the production config path by always passing a test-local `settings` path into a hand-built namespace. The real stage omits `--settings` and relies on parser/default-path wiring, so that wiring can regress while this suite stays green. Can we add a run-prefix or real-parser fixture that creates `.closedloop-ai/settings/code-review.json` and proves the final spawn plan honors a raised or zero cap?

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -3064,6 +3064,16 @@ def _load_code_review_settings(path: Path | None) -> dict[str, Any]:
...
+      - ``domain_critic_cap`` (int, ≥ 0): how many domain critics may
+        spawn across the required and best-effort buckets combined.
+        Default :data:`DOMAIN_CRITIC_CAP` (3). Setting the value to 0 is a
+        kill switch (no domain critic spawns; ``source: "core"``
+        reviewers stay exempt). Raise it when a repo's
+        ``critic-gates.json`` legitimately resolves more relevant critics
+        than the default allows: the cap drops by ``(priority asc,
+        reviewer asc)``, so on a roster where every entry sits at the
+        default ``priority: 2`` the tiebreak is alphabetical, which
+        favors early reviewer names over relevance to the diff.
```
shafty023, line 3077: This adds shipped plugin behavior, but `plugins/code-review/.claude-plugin/plugin.json` is unchanged at `3.7.0`. The required Plugin Version Bump check is already failing on this exact head. Please bump the code-review plugin version once relative to `main` (the repo guidance classifies a backward-compatible new setting/CLI surface as a minor bump).

plugins/code-review/README.md
```diff
@@ -305,11 +305,15 @@ Sensitive-path policy. See `start.md` for the full glob syntax and the three sup
 
 ### `code-review.json` (PLN-774)
 
-Operator-tunable reviewer behavior. Currently exposes the BHA conditional-partitioning threshold; future knobs will land here too.
+Operator-tunable reviewer behavior.
 
 | Key | Default | Behavior |
 |---|---|---|
```
shafty023, line 311: The repo marks this README as generated and requires `/update-documentation` for plugin changes, including the root changelog entry. `CHANGELOG.md` is unchanged on this head, so the release docs are incomplete even though the README was edited. Please run the existing documentation workflow and commit its generated README/changelog output after the version bump.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -11550,15 +11570,46 @@ def cmd_arbitrate_budget(args: argparse.Namespace) -> int:
...
+    if cap_override is not None:
+        critic_cap = int(cap_override)
+    else:
+        cap_settings_path = Path(
+            getattr(args, "settings", None) or _CODE_REVIEW_SETTINGS_DEFAULT_PATH,
+        )
+        critic_cap = int(
+            _load_code_review_settings(cap_settings_path).get(
+                "domain_critic_cap", DOMAIN_CRITIC_CAP,
+            ),
```
mikeangstadt, line 11605: Raising this past the total-cap headroom isn't a no-op, it flips the verdict. Step 3 below (required_overflow) drops the excess required critics via _make_coverage_gap_finding(reason="budget_exceeded"), those findings carry required: True, and _compute_canonical_verdict Rule 1 turns any required coverage gap into CHANGES_REQUESTED.

I ran it: default cap=20, domain_critic_cap=25, 25 required source:"rule" entries -> 6 dropped_required and 6 budget-exceeded gaps, all required: True. The legacy moduleCritics case you're actually fixing is safe (migrated entries are required: False so they land in best_effort and get pruned silently), but a repo on canonical coverage[] with required: true rules trips it once the cap goes past roughly cap - bha_target - len(required_non_critic).

The Step 3 comment still says "this branch normally doesn't fire" — that was true when the cap was a hardcoded 3 against a default cap of 20. It isn't anymore. Either clamp critic_cap to the available headroom here, or at minimum update that comment so the next person doesn't read it as unreachable.

plugins/code-review/README.md
```diff
@@ -305,11 +305,15 @@ Sensitive-path policy. See `start.md` for the full glob syntax and the three sup
...
+| `domain_critic_cap` | `3` | How many domain critics may spawn across the required and best-effort buckets combined. `source: "core"` reviewers (Design Critic, Impact Analyzer) are exempt. **Setting the value to `0` spawns no domain critics at all (kill switch).** Invalid entries (wrong type, negative) silently fall back to the default. Also overridable per-run with `arbitrate-budget --domain-critic-cap`. |
+
+**When to raise `domain_critic_cap`.** The cap drops by `(priority asc, reviewer asc)`. A repo whose `critic-gates.json` still uses the legacy `moduleCritics[]` schema gets every entry migrated as `required: False` with **no** `priority`, so all of them sit at the default `2` and the tiebreak degenerates to **alphabetical by reviewer name** — systematically favoring early names over relevance, and cutting the very critic the coverage critic proposed *for* the diff. Raising the cap is the blunt remedy; the precise one is migrating the relevant rules to the canonical `coverage[]` schema, which supports explicit `priority` and `required` (see `_migrate_module_critics` for what the legacy form forces).
```
mikeangstadt, line 316: This tells operators to raise the cap without giving them a ceiling. Past `cap - bha_partitions - required core` the overflow path starts dropping required critics as `budget_exceeded` coverage gaps, and those are `required: True` so the run auto-lands on CHANGES_REQUESTED. Worth a sentence saying the useful range tops out around the total `--cap` headroom, especially since `--cap` itself isn't exposed in this settings file.

plugins/code-review/tools/python/config/cli.json
```diff
@@ -408,7 +408,11 @@
           "flags": [
             "--depth"
           ],
-          "choices": ["shallow", "standard", "deep"],
+          "choices": [
+            "shallow",
```
mikeangstadt, line 412: Was this reformat intentional? The choices block here, the × → \u00d7 escape on derive-static-spec's help, and the EOF newline all look like collateral from regenerating the file rather than hand-editing it. The escape will churn straight back the next time someone edits by hand.

mikeangstadt (approved):
Right fix for a real problem, and the loader validation lines up with its siblings. One thing worth a look before this lands: raising the cap past the total --cap headroom doesn't no-op the way the PR body says, it drops required critics through the budget_exceeded path and auto-flips the verdict. Details inline. shafty's version-bump and settings-trust points still need answers too.

PR #193, opened 2026-09-06
fix(code-review): key coverage-critic cache on the diff too

mikeangstadt:
**Stacked on #192** (base is `fix/iss-8961-review-signal-cache-key`, not `main`). Retarget to `main` after #192 merges. ISS-9674.

## Summary

- **This is ISS-8961 in the `coverage_critic/` namespace, and it was safe only by accident.** `coverage_critic_cache_key` was `(coverage_plan_initial_hash, signals_hash, diff_tip, prompt_hash, available_reviewers_hash)`. `_build_coverage_critic_input` returns `(main_input, diff_summary)` where `diff_summary` is literally `_build_signal_input(diff_data, intent_summary=None)` — the bundle the critic reads — and it appeared in no key component. As in #192, `diff_tip` is a ref *name* and never a commit id, so nothing in the tuple varied with the diff, while the cache directory is the pooled worktree's basename.
- **The trap this closes.** The namespace did not misfire only because `signals_hash` hashes `extract_signals.json`, which carries a wall-clock `generated_at`. The key therefore changed every run and the cache never hit. Anyone looking at a namespace with a 0% hit rate would find that timestamp, remove it, measure a real speedup, and ship a critic that serves one review's coverage plan to a different review — **with no test failing**. That is the whole reason this is a separate, filed ticket rather than a footnote.
- The key now includes `diff_summary_hash` (`signal_input_hash` over that bundle), and `cmd_coverage_critic_prepare` builds the agent input *before* computing the key rather than after. The manifest reports `diff_summary_hash` beside its sibling component hashes.
- Corrects the key docstring's false claim that "All five are content-addressed" (`diff_tip` is not), and the `SCHEMA.md` coverage-critic row, which documented the path as `<diff_tip>.json` when the code writes `<key>.json` and listed three of the five components. These are the two doc errors #192 deliberately left on the adjacent table row to keep that PR to one subject.

## Breaking changes

None — additive. One-time invalidation of the `coverage_critic/` namespace, which in practice never hit. `coverage_critic_cache_key()` gains a positional argument; it has no callers outside this module and its tests.

## UI Feature Flag

`[ui-flag:N/A]` — tooling-only change in a Python helper; no UI surface in this repo.

## Test plan

- `test_pooled_lane_with_a_different_diff_misses_prior_cache` is the regression test: one shared cache directory, and plan_initial, signals, prompt, roster, and `diff_tip` all held **constant** so that only the diff varies — which is precisely what the old key could not see.
- **Counterfactual:** dropping `diff_summary_hash` back out of the key payload turns 3 tests red, including that one. Restored: green.
- Key composition pinned to a frozen literal digest, not re-derived through the builder under test.
- `test_cache_hit_serves_directly` no longer re-derives the key to seed its fixture; it prepares once and seeds under the key the manifest reports — the false-green shape caught in #192's review.
- `diff_summary_hash` is normalized in the golden harness like its sibling hashes: it is fixture-coupled rather than wall-clock, so pinning it in goldens would force a hand-updated digest on every fixture diff edit. Composition is pinned by the unit test instead.
- `ruff check` clean; `uv run pyright` 0 errors; `uv run pytest plugins/` **2119 passed, 0 failed**.

### DIFF
--- a/plugins/code-review/tools/python/code_review_helpers.py
+++ b/plugins/code-review/tools/python/code_review_helpers.py
@@ -7860,16 +7860,32 @@ def _stable_json_hash(payload: Any) -> str:
 def coverage_critic_cache_key(
     coverage_plan_initial_hash: str,
     signals_hash: str,
+    diff_summary_hash: str,
     diff_tip: str,
     prompt_hash: str,
     available_reviewers_hash: str,
 ) -> str:
     """Cache key for the ``coverage_critic`` namespace (PLN-725).
 
-    Tuple ``(coverage_plan_initial_hash, signals_hash, diff_tip,
-    prompt_hash, available_reviewers_hash)`` is the complete set of
-    inputs the critic is a pure function of. All five are
-    content-addressed.
+    Tuple ``(coverage_plan_initial_hash, signals_hash, diff_summary_hash,
+    diff_tip, prompt_hash, available_reviewers_hash)`` is the complete set
+    of inputs the critic is a pure function of. All are content-addressed
+    except ``diff_tip``, which is only a ref *name* (``"HEAD"`` for every
+    local branch review, ``origin/<branch>`` for a PR) and therefore
+    discriminates nothing on its own — it is a coarse extra component,
+    never the diff identity.
+
+    ``diff_summary_hash`` (ISS-9674) fingerprints the ``diff_summary``
+    half of ``_build_coverage_critic_input`` — which is literally
+    ``_build_signal_input(diff_data, intent_summary=None)``, the bundle
+    the critic reads. Before it was added, no key component varied with
+    the diff at all; the key changed run-to-run only because
+    ``signals_hash`` transitively covered ``extract_signals.json``'s
+    wall-clock ``generated_at``. That made this namespace correct by
+    accident and permanently cold, and it meant the obvious optimization
+    — strip ``generated_at`` so the cache finally hits — would have
+    served one review's coverage plan to another review from the same
+    pooled worktree, which is ISS-8961 in this namespace.
 
     ``available_reviewers_hash`` is the deterministic hash of the
     AVAILABLE roster the agent will actually see (sorted, dedup'd, and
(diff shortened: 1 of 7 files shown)


shafty023:

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -8351,9 +8368,16 @@ def cmd_coverage_critic_prepare(args: argparse.Namespace) -> int:
     # stale (could propose a now-removed reviewer), so it must key
     # the cache.
     available_reviewers_hash = _available_reviewers_hash(available_reviewers)
+    # ISS-9674: build the agent input before the key, so the key can cover
+    # the diff bundle the critic actually reads. Mirrors the ordering
```
shafty023, line 8372: P1: This hashes `diff_summary`, but the critic is still dispatched only `manifest.input_path`. That JSON contains the plan, signals, and roster; the separately written `coverage_critic_diff_summary.json` is never passed through singleton dispatch, even though the prompt says the critic receives it.

What happens is a new diff gets a new cache key, then the cache miss still runs a diff-blind critic. Please include the summary (or explicitly pass its path) in the dispatched input contract, hash that exact delivered content, and add a contract test that proves the critic receives it.

plugins/code-review/tools/python/test_code_review_helpers.py
```diff
@@ -14836,6 +14844,57 @@ def test_cache_hit_serves_directly(self, tmp_path: Path) -> None:
...
+            args = self._args(lane_root, inputs)
+            # One shared pooled cache directory across both lanes.
+            args.cache_dir = str(shared_cache)
+            assert cmd_coverage_critic_prepare(args) == 0
+            manifest = _read_coverage_section(Path(args.cr_dir), "critic")
+            assert manifest["status"] == "needs_agent"
+            return str(manifest["cache_key"])
+
+        shared_cache = tmp_path / "cr-cache-global-repo-pool"
+        (shared_cache / CACHE_NAMESPACE_COVERAGE_CRITIC).mkdir(parents=True)
```
shafty023, line 14892: P2: This only proves that the two computed keys differ. `_lane()` never writes a lane-1 entry under `lane1_key`, so both calls return `needs_agent` even if the cache read/reuse path is broken.

Seed a valid lane-1 cache payload in the shared namespace, then assert lane 2 is a miss and does not consume it. That is the regression this ticket needs: a wrong prior hit must be rejected, not just a different key calculated.

PR #194, opened 2026-09-10
fix(code-review): make override-cache tests independent of wall-clock date

wongk:
## Problem

Three tests in `test_code_review_helpers.py` were failing on `main` with no accompanying code change:

```
FAILED TestOverrideCache::test_verify_prepare_short_circuits_on_valid_override
FAILED TestPR114ReviewFixes::test_prepare_then_consolidate_routes_override_to_verified
FAILED TestPR114ReviewFixes::test_prepare_then_consolidate_writes_re_asserted_to_stats
```

They are **time bombs**. Each writes an override with a hardcoded `asserted_at` of `2026-05-29T22:00:00+00:00` and then asserts the override is *honored* — landing in `override_hits`, or routed to a `RE_ASSERTED` verdict.

The overrides cache has a 90-day TTL (`CACHE_TTL_DAYS["overrides"]`), enforced by `_override_is_expired`. The tests passed while that fixed date was recent. Once wall-clock passed 2026-08-27, `_override_is_valid` correctly began returning `False`, `override_hits` came back empty, and the assertions broke.

The production code is correct — expiring a 100-day-old override is exactly the intended behavior. Only the fixtures were wrong.

## Fix

Derive `asserted_at` relative to now, so the tests exercise the honored-override path whenever they run.

The file already had a helper for this — `_stale_cached_at(days_ago)`, returning an ISO timestamp N days in the past — and two sibling TTL tests already used the relative idiom inline. Rather than add a near-duplicate, this reuses the existing helper and renames it to `_iso_days_ago()` so the name matches its already-generic docstring (it is used for both within-TTL and past-TTL fixtures, so "stale" was misleading). All 6 existing call sites updated, plus the two override TTL boundary tests, which now delegate instead of inlining `datetime.now(timezone.utc) - timedelta(...)`.

## Scope check: is this the only time bomb?

Swept every hardcoded timestamp feeding TTL-checked fields across all plugins' tests. Two remain, both safe by construction:

| Location | Value | Why it can't rot |
|---|---|---|
| `test_code_review_helpers.py:3560,3586` | `cached_at: 2025-01-01` | Deliberately ancient; tests assert overwrite/retention, which an older date only reinforces |
| `test_code_review_helpers.py:10071` | `asserted_at: 2026-05-29` | Write/load round-trip equality; never passes through TTL logic |

The rot-prone pattern is specifically "hardcoded recent timestamp expected to be *within* TTL", and all three instances are fixed here.

## Also included

`/update-documentation` is mandatory before push in this repo, and its README-verification step flagged two pre-existing inaccuracies in `plugins/code-review/README.md`. Both are unrelated to the test fix but verified correct, so they are included rather than left wrong:

- Golden-fixture count said "3 full scenarios + 6 README-stubs"; on disk it is 4 full + 3 stubs.
- Four files present on disk were missing from the architecture tree: `config/cli.json`, `config/stages.json`, `signal_taxonomy.json`, `conftest.py`.

## Verification

- `pytest plugins/` — **2107 passed, 3 skipped** (was 3 failed, 2107 passed)
- `uv run pyright` — 0 errors
- Both TTL boundary controls still pass in the right directions: 120 days → invalid, 30 days → valid

No production code changed — test fixtures, one test-helper rename, README, CHANGELOG, and a PATCH version bump (3.7.0 → 3.7.1).

https://claude.ai/code/session_011QjUMmrqfGU4QNXDDRLPrN


shafty023:

shafty023 (approved):

PR #195, opened 2026-09-10
refactor(code-review): decouple review agents from codebase-memory-mcp

wongk:
## Why

The reviewer fleet was hard-bound to one MCP server. `agents/code-review-worker-graph.md` declared an allowlist of six `mcp__codebase-memory-mcp__*` tools, and an allowlist is exhaustive — no other server could ever reach a reviewer, no matter what the operator had connected. Unresolvable entries were silently dropped rather than erroring, which is why the coupling went unnoticed on machines without that server: reviews just quietly ran grep-only forever.

This makes the plugin substrate-agnostic. Reviewers now use whatever code-intelligence MCP server the operator's session provides — codebase-memory-mcp, LemonCrow, closedloop-graph, or something that does not exist yet — with no plugin edit required.

## Approach

**Access.** The graph worker drops `tools:` entirely and inherits the spawning session, with `disallowedTools: Bash, Edit, NotebookEdit` as harness-level hardening. `code-review-worker` keeps its explicit four-tool allowlist, so the verifier fleet, Bug Hunter A, the domain critics, and the singleton prompts still inherit nothing — the adversarial trust boundary is unchanged.

**Protocol.** The `codebase-memory-mcp` block in `shared_prompt.txt` becomes `OPTIONAL — CODE INTELLIGENCE`: four capabilities (C1 symbol lookup, C2 usage/caller enumeration, C3 snippet read, C4 structure) that the reviewer binds to whatever it holds, discovering its own roster and loading deferred MCP schemas with `ToolSearch`. Same rewrite across the impact-analyzer, design-critic, verifier prompts and the four suffixes in `spawn-reviewers`.

**Orchestrator.** `GRAPH_PROJECT` collapses to one computed boolean, `CODE_INTEL_ALLOWED`, false whenever `review_root` is set. That preserves the stale-index kill switch — substrate-independent, since any index covers the operator checkout rather than the PR head — and drops the `list_projects` handshake that substituted a server-returned project name into the agents' trusted instruction zone.

**Schema:** comments only. `discovery: "graph"` was already substrate-neutral, so no schema change and no fixture churn.

## Verified at runtime, not from docs

Each claim below was checked by running the real plugin through headless `claude -p --plugin-dir` and inspecting the spawned worker's actual tool roster.

| Check | Result |
|---|---|
| Baseline (before) | Graph worker got exactly `Read, Write, Grep, Glob` — **zero MCP**. The allowlist was genuinely gating. |
| Unresolvable entries | Agent still spawns; missing MCP tools silently dropped. Confirms the failure was invisible. |
| After | Worker inherits 139 MCP tools; Impact Analyzer bound C2 to `mcp__lc__relations` — a server this plugin never names — and cross-checked with `mcp__lc__code_query`. |
| Kill switch | `CODE_INTEL_ALLOWED=false` → zero MCP calls, silent degradation, finding still emitted. |
| `disallowedTools` | Bash/Edit/NotebookEdit removed; MCP inheritance survives intact. |
| `tools: mcp__*` | Silently dropped — wildcards are denylist-only, so allowlist and inheritance are mutually exclusive. Worth knowing before anyone tries to have both. |

## A fail-open this surfaced

Inheriting tools means inheriting the parent's *absence* of them. In a session without `Grep`/`Glob`, the reviewer still emitted `grep_query_used: "_parse_agent_name\\("` — **a query it never ran** — which the verifier replays as its fabrication check. That is a fail-open in the exact mechanism designed to catch hallucination.

Fixed in both prompts: text search is a capability, not a tool name, and any recorded search must describe a query actually executed. With no text-search tool the analyzer now emits `grep_query_used: null`, `external_usages_found: []`, and tags entries `discovery: "graph"`, routing them to the per-entry file-read and content-match audit — which `verifier_prompt.txt` already handles as the all-graph case, so no verifier change was needed.

Re-spiked after the fix and asserted the replay-gate invariant against the real vocabulary:

```
discovery value in vocab: True -> graph
grep_query_used: None
external_usages_found: []
grep-discovery entries: 0
REPLAY-GATE INVARIANT HOLDS: True
```

## Verification

- `pytest plugins/` — 2107 passed, 3 skipped
- `uv run pyright` — 0 errors
- Zero references to `codebase-memory-mcp` or `GRAPH_PROJECT` remain outside historical CHANGELOG entries
- README re-verified against disk: accurate, no changes needed beyond the agent line

## Notes for review

- **Expected one-time cache miss.** `shared_prompt.txt` and `verifier_prompt.txt` bytes feed `prompt_hash`, so the first run after merge re-pays the BHA and verification namespaces. Correct behavior — cached findings were produced under the old protocol.
- **Known gap:** `disallowedTools` cannot reach shell-equivalent MCP tools (e.g. `mcp__lc__bash`). Addressed with one clause in the agent prompt rather than new machinery, on the principle that the harness owns permissions.
- Version 3.7.1 → 3.8.0 (MINOR: new capability, no breaking interface change).

https://claude.ai/code/session_011QjUMmrqfGU4QNXDDRLPrN


shafty023:

plugins/code-review/agents/code-review-worker-graph.md
```diff
@@ -1,50 +1,54 @@
...
-tools: Read, Write, Grep, Glob, mcp__codebase-memory-mcp__search_graph, mcp__codebase-memory-mcp__trace_path, mcp__codebase-memory-mcp__get_code_snippet, mcp__codebase-memory-mcp__search_code, mcp__codebase-memory-mcp__get_architecture, mcp__codebase-memory-mcp__query_graph
+description: Code-intelligence-aware review worker for the cross-file and design reviewers (Impact Analyzer, Bug Hunter B, fast-path, Design Critic). Identical to code-review-worker but inherits the parent session's tools, so whatever code-intelligence MCP server the operator has connected is available for cross-file usage discovery and project-structure / dependency-graph analysis. Use only for reviewers whose role prompt loads the code-intelligence protocol.
+disallowedTools: Bash, Edit, NotebookEdit  # harness-level: reviewers never shell out or mutate source. MCP inheritance is deliberately untouched — see shared_prompt.txt "OPTIONAL — CODE INTELLIGENCE".
```
shafty023: P1: This now inherits every parent-session MCP tool but only denies `Bash`, `Edit`, and `NotebookEdit`. In a session with a write-capable MCP integration, a review worker can still mutate external state; the prompt instruction is not an enforcement boundary. Keep the new substrate discovery, but constrain the worker to read-only code-intelligence capabilities rather than inheriting arbitrary session tools.

plugins/code-review/tools/prompts/impact_analyzer_prompt.txt
```diff
@@ -65,23 +64,33 @@ changes.
...
-     skip this substrate entirely (grep-only). When available, pass
-     `project=<GRAPH_PROJECT>` on every call and use
-     `search_graph(qn_pattern=<symbol>, project=<GRAPH_PROJECT>)` plus
-     `trace_path(<symbol>, mode=calls, project=<GRAPH_PROJECT>)` to
-     enumerate callers, and `get_code_snippet` to read them. A caller the
-     graph surfaces but grep CANNOT (aliased import, re-export,
-     dynamic dispatch) is a `discovery: "graph"` usage — exactly the
-     blast radius grep misses. Validate every returned path is inside
-     this checkout before citing it (per the shared protocol).
+     replays, so it runs whenever you hold a text-search tool.
```
shafty023, line 76: P2: The new no-text-search branch says to leave `grep_query_used` null, but this same prompt still says every Impact finding must populate it and will be rejected if it is absent. The generated Impact suffix also unconditionally says to “Always run grep.” A worker that only has caller intelligence will either fabricate a replay query or drop the graph-only finding, so the intended all-graph fallback never works. Make the replay fields conditional everywhere they are described, and add coverage for a graph-only finding with no text-search tool.

wongk: Good catch — this was right, and in two places rather than one.

**The prompt file** (`impact_analyzer_prompt.txt`) is fixed in db8a791. The requirements block stated it categorically ("Findings without a populated `grep_query_used` ... will be rejected by the verifier as malformed") 15 lines before the new exception, and Step 2 repeated "Record the **exact** query string". Both are now qualified, and a missing query is malformed only when at least one entry is `discovery: "grep"`.

**The generated suffix** is fixed in fd21741 — and you are right that this was the one that actually mattered, since it is injected directly into the analyzer's task prompt. It still carried every unconditional form: `Always run grep too`, `populated external_impact[] and grep_query_used`, `If grep returns zero external usages`, and `Use Read, Grep, and Glob`. All four now describe text search as a capability the session may or may not provide, with an explicit "never write a query you did not execute".

**Coverage** added in fd21741: `test_external_impact_all_graph_with_null_grep_query_valid` and `..._with_omitted_grep_query_valid` in `test_code_review_schema.py`. Every pre-existing `discovery` test set `grep_query_used`, so the exact shape the fallback emits was unpinned — if the validator ever tightened, the fallback would have silently dropped findings.

On "the intended all-graph fallback never works": I ran the analyzer end to end against a live session with no `Grep`/`Glob` before and after. Before, it emitted `grep_query_used: "_parse_agent_name\("` — a query it never ran. After, it emits `grep_query_used: null` with `external_usages_found: []` and cites the prompt rule as its reason. `pytest plugins/code-review/tools/python/` is 1332 passed / 3 skipped.

plugins/code-review/agents/code-review-worker-graph.md
wongk: The claim is correct and I am not disputing it — but I want to put the constraint I hit on the record, because it rules out the cheap version of your fix.

**What I tested.** I tried to close this at the harness level with a name-agnostic denylist:

```yaml
disallowedTools: Bash, Edit, NotebookEdit, mcp__*__bash, mcp__*__edit, mcp__*__write
```

It does not work. I spawned the worker and read its live roster: `mcp__lc__bash` and `mcp__lc__edit` were both still present. Per the docs the only patterns `disallowedTools` honors on the MCP side are `mcp__<server>` (a whole named server) and `mcp__*` (every MCP tool) — and `mcp__*` would remove the code-intelligence tools this PR exists to reach. There is no wildcard that means "deny write-shaped tools from any server".

So the options are: (a) enumerate write-capable servers in the denylist, which re-couples the plugin to a known list — the exact coupling being removed here; (b) a `PreToolUse` hook denying exec/write-shaped tool names for these agents, which is genuine enforcement but a new hook surface for a plugin that has none today, plus an over-blocking risk on innocently-named read tools; or (c) an allowlist, which ends open-ended discovery.

**Decision for this PR: keeping the current shape,** on the standing call that tool permissions belong to the harness rather than to plugin prompt text. What is enforced structurally is narrower than the CHANGELOG implied, so I am not going to claim otherwise: `disallowedTools` covers the three native tools and nothing else, and inherited MCP write tools are covered by prompt instruction only.

To be explicit about the residual risk you are pointing at: it is real, and it is strictly larger than the old allowlist's. The mitigating facts are that the untrusted-content policy already classifies all MCP output as data, and `code-review-worker` — which runs the verifier fleet, Bug Hunter A and the domain critics — keeps its explicit four-tool allowlist and inherits nothing.

**Leaving this thread open for you.** If you want the `PreToolUse` hook, say so and I will build it here rather than defer it; I did not want to add a hook surface to this plugin unilaterally inside a decoupling PR.

PR #198, opened 2026-09-12
fix(code-review): refuse a --github review of a tree that is not the PR head (ISS-8769)

peterulsteen:
## Summary

- `/code-review --github <PR>` run outside GitHub Actions reviewed **whatever tree happened to be checked out**, silently and with exit 0. `cmd_resolve_scope` skipped every PR-head check when the mode was github, but `--github` selects file-based handoff output — it does not declare "I am inside GitHub Actions" — and running it from a developer machine is supported. Such a run left `review_root` empty and every reviewer read the operator's branch while holding a diff computed from `origin/<base>...origin/<head>`. Worst on the deep tier, whose cross-file reviewers (Impact Analyzer, Design Critic, Bug Hunter B) grep for callsites the wrong tree does not have; silent in both directions, because nothing in the output said which tree was read.
- **GitHub mode now verifies and refuses.** It checks that the checked-out tree holds the PR head and exits non-zero when it cannot establish that. Two checkout shapes are accepted, because **both occur in Actions**: HEAD is the PR head (`actions/checkout` with an explicit `ref: <head sha>`), or HEAD is the PR's merge ref `refs/pull/N/merge` — the **default** checkout for a `pull_request` event — whose second parent is the head. Either way tracked files must be unmodified.
- **The block adds a refusal and changes nothing else.** No worktree, no `review_root`, and `head_sha` stays out of the emitted scope — deliberately, because populating it would re-route `_file_content_hash` (which prefers `git show <head_sha>:<file>` over the working tree) and the inline-comment `commit_id` in `cmd_post_comments`. Cost on an accepted run is at most three `rev-parse` calls plus one `status`.
- **Isolation stays local-only on purpose.** Setting `review_root` in github mode was the first shape of this fix and was withdrawn: it redirects a whole agent fleet, forces `CODE_INTEL_ALLOWED=false`, and `github-review.md`'s own outdated-thread step (Step 6a) reads file state without resolving it — so the wrong tree would have persisted into `code-review-threads.json`, which the workflow resolves on the PR. Refusing is smaller and honest; the remedy it names is to check out the PR branch, or drop `--github` and use the local flow, which isolates for you.
- Also fixes a second silent path the mode gate was hiding: when `gh pr view` fails in CI, the guess fallback sets `head_ref` from the detached `current_branch` and `diff_tip` becomes `origin/HEAD`. That is now a refusal instead of a review against a diff nobody can compute.

## Breaking changes

None for any tree that is actually the PR's source; CI is unchanged on both `actions/checkout` shapes (verified below). A `--github` run whose tree is not the PR head now exits 1 where it previously exited 0 — that is the fix, and the diagnostic names both accepted shapes and the remedy. No payload, field, or file contract changes: `scope.json` keeps every key with the same meaning, and github mode still emits `review_root: ""` / `head_sha: ""` exactly as before.

## UI Feature Flag

[ui-flag:N/A] Developer-tooling CLI change in the code-review plugin. No shipped UI on any surface.

## Test plan

Empirical, with `resolve-scope --mode github --pr-number 7157` against a real PR:

```
MISMATCHED tree (on another branch): BEFORE exit 0, review_root ''  (reviewers read that tree)
                                     AFTER  exit 1, stdout empty, diagnostic names both shapes
CONTROL A  HEAD == PR head:          AFTER  exit 0, review_root '', head_sha ''  (no worktree)
CONTROL B  HEAD is a merge commit
           whose ^2 == PR head:      AFTER  exit 0, review_root ''   (default-CI shape accepted)
CONTROL C  diff_tip == origin/HEAD
           (gh pr view failed):      BEFORE exit 0        AFTER exit 1
```

- Mutation table, each row killed by exactly the case that names it: remove the verification → 4 refusal cases fail; drop the merge-ref arm → `accepts_the_prs_merge_ref` fails; accept *any* merge commit instead of this PR's head → `refuses_a_merge_ref_for_a_different_head` fails; drop the clean-tree requirement → `refuses_a_dirty_tree_at_the_head` fails.
- `uv run pytest plugins/` — **1338 passed, 3 skipped, 0 failed**. `uv run ruff check .` clean; `uv run pyright` on the changed module 0 errors.
- `test_no_worktree_github_mode_with_pr_scope` pinned the defect (no worktree, rc 0, on a tree whose HEAD is not the PR head) and is replaced, not weakened: it encoded "github mode means a runner already holding the PR head" while setting up the state where that premise fails. Its CI half is now carried by the two explicit runner-shape cases.
- Not run: an end-to-end `/code-review` inside GitHub Actions, and the `design-inventory-ts` job (untouched).


shafty023:

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -5062,6 +5094,38 @@ def cmd_resolve_scope(args: argparse.Namespace) -> int:
...
+    # supported — so the old code, which skipped every check whenever the mode
+    # was github, silently reviewed whatever branch happened to be checked out
+    # while holding a diff computed from the PR's remote refs. Nothing in the
+    # run said which tree was read, and the deep tier's cross-file reviewers
+    # (Impact Analyzer, Design Critic, Bug Hunter B) are exactly the ones a
+    # wrong tree makes confidently wrong.
+    #
+    # This block ADDS a refusal and changes nothing else: no worktree, no
+    # ``review_root``, and ``head_sha`` stays out of the emitted scope so no
+    # downstream consumer of it changes behavior in github mode. Either the
```
shafty023, line 5166: P1: This still treats the guessed `diff_tip` as the requested PR head. `cmd_resolve_scope` allows `_resolve_pr_scope` to fall back to `current_branch` when `gh pr view` fails; a clean checkout of that branch then passes this check and gets reviewed against the requested PR diff. GitHub mode needs to require successful PR metadata/head resolution before verifying the checkout, with a regression case for a resolvable fallback branch.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -462,6 +464,33 @@ def _working_tree_clean() -> bool:
...
+    established, and a caller must refuse rather than review it. Uncommitted
+    changes to tracked files disqualify either shape: the tree then holds
+    content that is in no commit this PR contains. Untracked files do not,
+    matching ``_working_tree_clean``.
+
+    Costs at most three ``rev-parse`` calls plus one ``status``, and short-
+    circuits on the head-checkout case.
+    """
+    if not pr_head:
+        return False
```
shafty023, line 537: P1: `HEAD^2 == pr_head` does not prove this is PR #N's merge checkout. A clean local merge with the PR head as its second parent but an unrelated first parent also passes, so reviewers can still read a combined tree that does not match the PR diff. Bind this case to the actual PR merge/base identity and add a negative test for the unrelated-first-parent merge.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c4a0487. `_resolve_pr_scope` now reports `head_ref_from_metadata` (false when `gh pr view` fails or returns no head branch), and in github mode `cmd_resolve_scope` refuses on a guessed head before the checkout check runs. Regression test `test_github_mode_refuses_a_guessed_head_even_when_the_tree_matches_it` covers both failure shapes with a pushed, checked-out, clean fallback branch that the checkout check alone would have accepted.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in c4a0487. The merge shape now requires exactly two parents, the second parent equal to the PR head, and the first parent an ancestor of `origin/<PR base>` (base taken from PR metadata, before any `--base` override). Ancestry rather than equality, so a base that advanced after GitHub built `refs/pull/N/merge` is still accepted; any git error refuses. New negative tests: `test_github_mode_refuses_a_merge_whose_first_parent_is_not_on_the_base` and `test_github_mode_refuses_an_octopus_merge_of_the_head` (the old `HEAD^2` check also accepted an octopus that folds in a third branch). Also probed against real history: orphan first parent, unmerged side-branch first parent, octopus, and swapped parents are refused; base-unchanged, base-advanced, and older-base-tip merges are accepted.

PR #200, opened 2026-09-15
feat(code-review): record the reviewed checkout and commit in review output (ISS-9137)

wongk:
## Summary

With #191 and #198 merged, the pipeline knows which checkout it reviewed, but nothing a reader sees says so. A review of the wrong tree still looks ordinary. `diff_tip` can't fill the gap, because for a branch review it's the literal `HEAD`. This PR records the checkout and resolved commit in three places:

- **`review_result.json`** gains `review_root`, `review_root_sha` and `review_root_tree` (staged reviews only).
  - Values are re-validated from `scope.json`, and anything absent or malformed is written as `null`.
  - The fields are additive, so there is no `SCHEMA_VERSION` bump (SCHEMA.md §10). The validator rejects any value that is neither a string nor `null`.
- **Footer:** `footer.json` gains `reviewed_line`, which `start.md` prints under the footer line.
  - Normal review: ``**Reviewed:** `<root>` @ `<sha12>` ``
  - PR head isolated into a worktree: `PR #N head @ <sha12> (isolated worktree, removed after review)`. The same stage deletes that worktree, so it isn't named by path.
  - Staged review: appends `+ staged index (tree <tree12>)`.
  - No valid SHA recorded: `checkout not recorded`.
- **GitHub summary:** a new `render-reviewed-commit` subcommand prints a `**Reviewed commit:**` line, which `github-review.md` embeds right after **Status**.
  - It names the reviewed commit, marked `(PR head)` when it matches, or followed by `(PR head is <sha12>)` when the runner checked out a merge ref.
  - It never prints the runner's path.

Every value is re-validated before it is printed. A root must be an absolute path free of control characters, `<`, `>` and backticks. A SHA must be hex.

The optional `--review-root` override from the issue was dropped, because `campaign` already runs `/code-review:deep` from inside the chunk worktree. ISS-9265 (cwd-relative refs and paths in subagent prompts) stays separate: this change makes that failure visible but does not fix it.

## Test plan

- [x] `pytest plugins/code-review`: 1396 passed, 3 skipped
- [x] `uv run ruff check .` and `uv run pyright`: clean. `git diff --check`: clean
- [x] Mutation checks, each caught by the new tests:
  - dropping the root character guard
  - dropping `review_root` from the envelope
  - disabling isolated-worktree detection
  - breaking the PR-head comparison
- [x] End-to-end CLI probe against a real git repo, running `resolve-scope`, then `finalize-result`, `footer` and `render-reviewed-commit`:
  - branch scope: the path and commit match `git rev-parse HEAD`
  - staged scope: the tree matches `git write-tree`
  - missing `scope.json`: prints `not recorded`
- [x] Regenerated fixtures: 4 golden envelopes (3 new `null` keys each) and the CLI parser snapshot (1 new parser)

Closes ISS-9137.


shafty023:

plugins/code-review/SCHEMA.md
```diff
@@ -143,6 +143,9 @@ The terminal artifact of every review run.
   "pr_number": <int|null>,
   "head_sha": "<sha|null>",
   "diff_tip": "<sha>",
+  "review_root": "<absolute path|null>",     // checkout the review read (ISS-9137)
```
mikeangstadt, line 146: Four lines up, line 139, this same envelope block still reads schema_version: 1, but SCHEMA_VERSION is 2 and every golden fixture emits 2. Pre-existing drift, not something you introduced, except this PR's CHANGELOG entry specifically claims schema_version stays at 2, so the doc it points a reader at now contradicts the claim. Cheap to true up while you already have the block open.

mikeangstadt (approved):
Checked the thing that would have made this decorative: resolve-scope really does write review_root, review_root_sha and review_root_tree, so the new envelope fields carry real values instead of always-null. Ran the suite locally, 1396 pass, ruff and pyright clean. One stale doc line inline, not blocking.

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6151,8 +6151,123 @@ def _format_elapsed(seconds: int) -> str:
...
+        and not _REVIEW_ROOT_FORBIDDEN.search(raw_root)
+        else None
+    )
+    raw_pr = str(scope.get("pr_number") or "")
+    return {
+        "review_root": root,
+        "review_root_sha": _validated_head_sha(scope.get("review_root_sha")) or None,
+        "review_root_tree": _validated_head_sha(scope.get("review_root_tree")) or None,
+        "pr_head_sha": (
+            _validated_head_sha(setup.get("head_sha"))
```
shafty023: P1: This never has a PR-head SHA in a real GitHub merge-ref run. `resolve-scope` intentionally leaves `scope.json.head_sha` empty there, and `setup` does not write `head_sha`; the new test seeds that value by hand. The summary will therefore show only the merge commit instead of the required `PR head is …` provenance.

Persist the metadata-resolved PR head separately from the legacy `head_sha` field, consume it here, and cover the actual merge-ref path end to end.

plugins/code-review/tools/python/code_review_schema.py
```diff
@@ -1080,6 +1080,10 @@ def _validate_envelope_scalars(envelope: dict[str, Any]) -> list[str]:
     verdict = envelope.get("verdict")
     if verdict not in VERDICTS:
         out.append(f"verdict {verdict!r} not in {sorted(VERDICTS)}")
+    for key in ("review_root", "review_root_sha", "review_root_tree"):
```
shafty023, line 1083: P2: This validates the new fields at runtime, but the exported `result_envelope_json_schema()` still does not declare them. That leaves schema-driven consumers with a descriptor that disagrees with `review_result.json` and `SCHEMA.md`.

Add the three nullable-string properties to the canonical JSON Schema and assert them in its schema test.

plugins/code-review/prompts/github-review.md
```diff
@@ -354,9 +354,19 @@ SUMMARY_EOF
...
 
 **Status:** [Approved | Changes Requested | Needs Attention]
+
+**Reviewed commit:** [render-reviewed-commit output]
+```
+
+**Reviewed commit line (ISS-9137).** Do NOT hand-author it. Run the renderer and embed its output verbatim, on its own line, immediately after the **Status** line:
+
+```bash
+python "${CLAUDE_PLUGIN_ROOT}/tools/python/code_review_helpers.py" render-reviewed-commit --cr-dir <CR_DIR>
```
shafty023: P2: This bypasses the fallback path the command already supports. `start.md` allows an empty `CLAUDE_PLUGIN_ROOT` and resolves `<HELPERS>` from dogfood/marketplace cache, but this expands to `/tools/python/code_review_helpers.py` in that mode. The GitHub summary then misses its required reviewed-commit line.

Use the already-resolved helper path for this renderer too (or preserve the resolved plugin root for presentation commands).

plugins/code-review/tools/python/code_review_helpers.py
```diff
@@ -6151,8 +6151,123 @@ def _format_elapsed(seconds: int) -> str:
...
+        raw_root
+        if isinstance(raw_root, str)
+        and os.path.isabs(raw_root)
+        and not _REVIEW_ROOT_FORBIDDEN.search(raw_root)
+        else None
+    )
+    raw_pr = str(scope.get("pr_number") or "")
+    return {
+        "review_root": root,
+        "review_root_sha": _validated_head_sha(scope.get("review_root_sha")) or None,
```
shafty023, line 6192: P2: These files are re-read after reviewer workers have write access to the shared run directory, but a well-formed SHA/root is not bound to the checkout that was proved at scope resolution. A compromised or instruction-following worker can replace this metadata with a plausible commit and make the footer, summary, and envelope report false provenance.

Keep the proven value in a run-owned integrity-protected artifact, or re-prove the root and SHA/tree against the reviewed snapshot before rendering/finalization; reject tampered-but-well-formed values in tests.

plugins/code-review/SCHEMA.md
wongk: Fixed in 35a435a. The envelope example in SCHEMA.md now shows `"schema_version": 2`. The finding block above it stays at `1`, which is still the version findings emit.

plugins/code-review/tools/python/code_review_helpers.py
wongk: Fixed in 35a435a. You're right: `setup` never writes `head_sha`, and github mode leaves `scope.json`'s `head_sha` empty, so a real run could never show the PR head.

- **What changed:** `resolve-scope` now records the PR head it had already resolved under a separate key, `pr_head_sha`, for PR scope only. Provenance reads that key, and `setup.json` is no longer consulted.
- **What didn't:** `head_sha` stays empty in github mode, so `_file_content_hash` and the inline-comment `commit_id` behave exactly as before.
- **New test:** `test_github_merge_ref_records_the_pr_head_for_provenance` runs `cmd_resolve_scope` on a merge-ref checkout (a two-parent merge with the PR head as second parent, first parent on the base) and renders the summary line from that scope. It expects ``**Reviewed commit:** `eeeeeeeeeeee` (PR head is `aaaaaaaaaaaa`)``.
- **Updated test:** the summary-line test now uses resolve-scope's real github-mode scope shape instead of a hand-seeded `setup.json`.

plugins/code-review/tools/python/code_review_schema.py
wongk: Fixed in 35a435a. `result_envelope_json_schema()` now declares `review_root`, `review_root_sha` and `review_root_tree` as `["string", "null"]`, and `test_result_envelope_json_schema_well_formed` asserts all three.

plugins/code-review/prompts/github-review.md
wongk: Fixed in 35a435a. The existing `render-fleet-summary` line here and the one in `skills/present-local/SKILL.md` had the same problem. All three render commands now run `python <HELPERS> …`, the helper path `start.md` resolves in the normal, in-repo dogfood, and marketplace-cache cases.

plugins/code-review/tools/python/code_review_helpers.py
wongk: I'm leaving this out of this PR. The exposure is real, but it isn't specific to these fields.

- **Already exposed:** reviewer and verifier agents can already write any file in the run directory, and `verify-prepare` already reads `scope.json` after the reviewers have run to decide where verifiers read. A forged `scope.json` redirects that today.
- **A re-check here wouldn't close it:** `_require_review_root` checks the root and commit against `diff_data.json`'s file list, and the same agent can rewrite that file. With an empty list, any commit the checkout contains passes the drift check.
- **What would close it:** integrity for the whole run directory that agents can't forge, for example helper-owned state outside the agents' writable tree, or a keyed digest of the scope and diff artifacts checked at finalize. That belongs in its own issue, not bolted onto the provenance fields.

What this PR does catch is the failure ISS-9137 was filed for: accidentally reviewing the wrong tree because of working-directory drift.

